Issue #4, opened 2020-08-28
Remove references to AdSupport framework?


sjmadsen:
With Apple further restricting apps that track users in iOS 14, I'd like to see a version of this framework that removed any references to AdSupport or the IDFA (identifier for advertisers). The "upstream" repo that this one is based on has done this for their current 4.x release (see [this pull request](https://github.com/segmentio/analytics-ios/pull/892)), allowing apps that want to collect the IDFA to provide it, rather than having the framework itself make those calls into AdSupport.

How do PostHog iOS users feel about such a change? I'm happy to make the changes and create a PR if there is interest.

jamesefhawkins:
Hey @sjmadsen , sorry for the slow response here.

I think this makes total sense as a PR if it's something you'd use and then wouldn't mind sharing. I imagine functionality-wise we should have:

* enabling apps that do not want ad tracking to pass App Store guidelines in certain categories while still allowing apps that do ad tracking to continue to function.
* to put a paragraph just to explain how to do the above [in here](https://posthog.com/docs/integrations/ios-integration)

sjmadsen:
I'll get a PR together in the next few days.

mariusandra:
This PR is now merged and version 1.1.0 of the library is out! Thanks again @sjmadsen !

PR #5, opened 2020-10-02
Shift responsibility of IDFA collection to clients

sjmadsen:
**What does this PR do?**
This is a backport from segmentio/analytics-ios (https://github.com/segmentio/analytics-ios/pull/892).

Any references to Apple's AdSupport framework might trip the App Store's static analysis, making it a bit unclear if SDK users can confidently say that their apps do not use the IDFA. By moving the responsibility to collect the IDFA out of the PostHog SDK and into the client app, the AdSupport references are gone and the question is easily answered.

It's a pretty small PR, but note that the PHGIDFA() function is gone and the caller of that function now checks for and calls a block provided as part of the SDK initialization.

With no changes to a client app, trigger an event, and verify that `$device_advertisingId` no longer has any value.

Then modify the app to provide a block for `adSupportBlock` in the PostHog configuration. Triggering an event should now include the return value from this block as the value for `$device_advertisingId`.

#4 - Remove references to AdSupport framework

n/a

**Questions:**
- Does the docs need an update? Yes; if apps want to collect the IDFA from the user's device, merely setting `enableAdvertisingCapturing` is no longer sufficient.
- Are there any security concerns? No.
- Do we need to update engineering / success? No?

### DIFF
--- a/PostHog/Classes/Internal/PHGPostHogIntegration.m
+++ b/PostHog/Classes/Internal/PHGPostHogIntegration.m
@@ -154,8 +154,8 @@ static CTTelephonyNetworkInfo *_telephonyNetworkInfo;
     if (NSClassFromString(PHGAdvertisingClassIdentifier)) {
         dict[@"$device_adCapturingEnabled"] = @(GetAdCapturingEnabled());
     }
-    if (self.configuration.enableAdvertisingCapturing) {
-        NSString *idfa = PHGIDFA();
+    if (self.configuration.enableAdvertisingCapturing && self.configuration.adSupportBlock != nil) {
+        NSString *idfa = self.configuration.adSupportBlock();
         if (idfa.length) dict[@"$device_advertisingId"] = idfa;
     }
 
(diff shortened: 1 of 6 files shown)


timgl:
@sjmadsen Awesome, thanks so much for contributing this.

@mariusandra would you mind reviewing this?

mariusandra:
Thank you @sjmadsen , this is awesome!

Could you just also provide an example that we could put in the docs for users who want to use this AdSupport framework? Preferably for both Objective-C and Swift...

sjmadsen:
I sure can. It's probably worth calling out separately from the "all configuration options" section now, since it's not enough to use the defaults and still get a value from clients for `$device_advertiserId`.

To capture the IDFA in Objective-C, SDK clients will want to do something like this:

``` objective-c
// At the top of your .m file
@import AdSupport;  // or #import <AdSupport/AdSupport.h>

// During PostHog initialization
PHGPostHogConfiguration` *configuration = [PHGPostHogConfiguration configurationWithApiKey:@"YOUR_API_KEY"];
configuration.enableAdvertisingCapturing = YES;
configuration.adSupportBlock = ^{
    return [[ASIdentifierManager sharedManager].advertisingIdentifier uuidString];
};
```

Or in Swift:

``` swift
// At the top of your .swift file
import AdSupport

// During PostHog initialization
let configuration = PHGPostHogConfiguration(apiKey: "YOUR_API_KEY")
configuration.enableAdvertisingCapturing = true
configuration.adSupportBlock = {
    return ASIdentifierManager.shared().advertisingIdentifier.uuidString
}
```

PR #9, opened 2021-02-10
Renamed some functions from utils

serhey-dev:
**What does this PR do?**
It renames some functions from utils.

It just renames few functions from utils, and replaces their name with new ones throughout the project.

It shouldn't cause build errors when this package and https://github.com/segmentio/analytics-ios are using together.

It's impossible to use PostHog library together with Segment library on iOS. It happens because PostHog iOS package is based on Segment package, and they have duplicating function names which is causing failing of app build.

https://github.com/PostHog/posthog-react-native/issues/2

**Questions:**


mariusandra:
Thanks @serhey-dev for this! Apologies for the wait, yet `1.2.3` is now out with the change. I'll update the RN repo next.

PR #19, opened 2021-09-17
Remove invalid Exclude: option from Package config

csykes:
Screen Shot 2021-09-17 at 9 50 54 AM

**What does this PR do?**
- Removes the exclude line from the swift package configuration which points to a directory/file that does not exist

- This may be hard to test inside the posthog-ios project but if you setup another Xcode project that includes this package as a dependency then you should no longer see the warning

**Questions:**

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -20,7 +20,6 @@ let package = Package(
             name: "PostHog",
             dependencies: [],
             path: "PostHog/",
-            exclude: ["SwiftSources"],
             sources: ["Classes",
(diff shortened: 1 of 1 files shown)


mariusandra (approved):

mariusandra:
Thanks! 1.4.1 released with this patch.

csykes:
Verified its cleared up one of the warnings... still have the second warning about the info.plist tho... will follow up with another PR for that shortly

PR #22, opened 2021-11-19
Make enabled property public (readonly).

pixlwave:
**What does this PR do?**
Currently you can call `enable` or `disable` on `PHGPostHog` but there is no way to inspect whether the client is actually enabled or not. Rather than requiring library users to keep track of this themselves, this makes the `enabled` property public (but read only) so it is easy to inspect.

**Edit:** I do wonder if it might be worth renaming the getter on this property to `isEnabled` is it's more obviously distinct from the `enable` method, happy to do so if you think this would be worthwhile.

There's only 1 line changed :)

Create a client and then check the value after calling `enable`/`disable` on it.

N/A

N/A

N/A

**Questions:**
- Does the docs need an update? - Probably not.
- Are there any security concerns? - No

### DIFF
--- a/PostHog/Classes/PHGPostHog.h
+++ b/PostHog/Classes/PHGPostHog.h
@@ -14,6 +14,11 @@ NS_ASSUME_NONNULL_BEGIN
 @interface PHGPostHog : NSObject
 
 /**
+ * Whether or not the posthog client is currently enabled.
+ */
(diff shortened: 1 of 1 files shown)


mariusandra:
Thanks! 1.4.4 is out with the change!

Panda777Igor (commented):
0x7639575853d3A47d4D86770493cCaBa331d2de3a

PR #26, opened 2022-07-06
fix: Test runner

benjackwhite:
**What does this PR do?**
* This should fix some deprecated swift code for the primary test suite
* Also adds a Github action for automatic test running

@EDsCODE - you could try pulling this and it _may_ sort out your reference build errors (but probably not)


.github/workflows/tests.yml
```diff
@@ -0,0 +1,36 @@
...
+  run_tests:
+    runs-on: macos-12
+    strategy:
+      matrix:
+        include:
+          # Latest xcode
+          - xcode: "13.4.1"
+            ios: "15.5"
+          # Older version
+          - xcode: "13.1"
```
benjackwhite, line 25: For now I set this to just an older version of Xcode, but ideally we should run the tests against the oldest version we want to support

PR #28, opened 2022-07-07
Basic feature flag + Groups support

EDsCODE:
**What does this PR do?**
- Adds support for group api
- Adds support for basic boolean feature flags
- Adjusts handling of distinct/anonymous ids to match posthog-js

1. Distinct id handling
2. Group handlers
3. Feature flag handlers

TBD. Bundle and test in a local project?

- #17 

**TODO:**
- [x] Trigger feature flag on load
- [x] QA locally
- [ ] Handle decide endpoint more broadly
- [ ] Update docs


EDsCODE:
The distinct id - anonymous id handling here needs work

PostHog/Classes/PHGPayloadManager.m
```diff
@@ -102,15 +116,11 @@ - (void)identify:(NSString *)distinctId properties:(NSDictionary *)properties op
...
-        [self saveAnonymousId:anonymousId];
-    } else {
-        anonymousId = self.cachedAnonymousId;
-    }
+    [self saveAnonymousId:self.integration.distinctId];
+    [self.integration saveDistinctId:distinctId];
 
     PHGIdentifyPayload *payload = [[PHGIdentifyPayload alloc] initWithDistinctId:distinctId
-                                                                     anonymousId:anonymousId
+                                                                     anonymousId:self.cachedAnonymousId
```
liyiy, line 123: What is the context behind this? 
`Adjusts handling of distinct/anonymous ids to match posthog-js` = ?

PostHog/Classes/PHGPayloadManager.m
```diff
@@ -157,6 +167,69 @@ - (void)alias:(NSString *)alias options:(NSDictionary *)options
...
+    NSArray* keys = [flags allKeys];
+    
+#if TARGET_OS_TV
+        [self.userDefaultsStorage setArray:keys forKey:PHGActiveFeatureFlags];
+#else
+        [self.fileStorage setArray:keys forKey:kPHGActiveFeatureFlags];
+#endif
+    
+#if TARGET_OS_TV
+        [self.userDefaultsStorage setString:anonymousId forKey:PHGEnabledFeatureFlags];
```
EDsCODE, line 195: `setString:anonymousId` typo should be setArray:flags

PostHog/Classes/PHGPayloadManager.m
EDsCODE: The library right now isn't saving "current" distinct_id and the "previous" distinct_id in a way that allows for implicit handling in functions such as isFeatureEnabled. The client version of these functions don't need to be explicitly passed a distinct_id because theyll use the cache'd one to determine the flag

PostHog/Classes/PHGPayloadManager.m
liyiy: i have no idea why your tests that are failing are passing for me locally 🤣

PostHog/Classes/PHGPayloadManager.m
```diff
@@ -192,7 +192,7 @@ - (void)receivedFeatureFlags:(NSDictionary *)flags
 #endif
     
 #if TARGET_OS_TV
-        [self.userDefaultsStorage setString:anonymousId forKey:PHGEnabledFeatureFlags];
+        [self.userDefaultsStorage setDictinoary:flags forKey:PHGEnabledFeatureFlags];
```
liyiy, line 195: `setDictionary`

EDsCODE:
Feature flag works. Groups don’t work for some reason. The call is being queued in the client but nothing is making it to the posthog instance

PostHog/Classes/PHGPayloadManager.m
```diff
@@ -13,6 +13,16 @@
 static NSString *const PHGAnonymousIdKey = @"PHGAnonymousId";
 static NSString *const kPHGAnonymousIdFilename = @"posthog.anonymousId";
 
+static NSString *const PHGActiveFeatureFlags = @"PHGActiveFeatureFlags";
+static NSString *const kPHGActiveFeatureFlags = @"posthog.activeFeatureFlags";
+
+static NSString *const PHGEnabledFeatureFlags = @"PHGEnabledFeatureFlags";
+static NSString *const kPHGEnabledFeatureFlags = @"posthog.enabledFeatureFlags";
```
benjackwhite, line 20: Inline with some feedback on posthog-android I think we can get away with just storing one item https://github.com/PostHog/posthog-android/pull/11

benjackwhite (changes requested):

EDsCODE:
There's inconsistencies with the test suite. Even on master it seems tests will still fail. I'm suggesting the removal of the workflow for the time being

PostHogTests/CapturingTests.swift
```diff
@@ -70,6 +70,19 @@ class CapturingTests: QuickSpec {
       expect(screen?.name) == "Home"
       expect(screen?.properties?["referrer"] as? String) == "Google"
     }
+    
+    it("handles group:") {
+      posthog.group("some-type", groupKey: "some-key", properties: [
```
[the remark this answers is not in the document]
EDsCODE: obj-c function syntax. The first argument is supposed to be related to the "function name"

liyiy:
Does it make sense to add example functionality usages like we've been doing in the server libraries with `example.py` for example to make this easier for testing?

EDsCODE:
Added calls to the example directory projects

PostHog/Classes/PHGPayloadManager.m
benjackwhite: Seems weird thought that we're adding something here that we know we are going to never need... As far as I can see the `activeFeatureFlags` is purely a legacy thing for the frontend js. For new libraries we only need the dictionary version as we can always derive the boolean outcomes from that?

PostHog/Classes/PHGPayloadManager.m
```diff
@@ -102,15 +116,11 @@ - (void)identify:(NSString *)distinctId properties:(NSDictionary *)properties op
...
     NSCAssert1(distinctId.length > 0, @"distinctId (%@) must not be empty.", distinctId);
 
-    NSString *anonymousId = [options objectForKey:@"$anon_distinct_id"];
-    if (anonymousId) {
-        [self saveAnonymousId:anonymousId];
-    } else {
-        anonymousId = self.cachedAnonymousId;
-    }
+    [self saveAnonymousId:self.integration.distinctId];
+    [self.integration saveDistinctId:distinctId];
```
benjackwhite, line 120: Asking here as a lot of this logic is new / confusing to me 😅 

Why do we save the anonymousId to be the previous distinctId and then the distinctId to be the new one? 

More specifically I'm wondering why we need two IDs? Why is the flow not like
* Load or generate distinctID
* use ID everywhere for `distinct_id` 
* on Identify payload set the current `distinct_id` as `$anon_distinct_id` and the new one as `distinct_id`
  * save the new `distinct_id` to the persisted ID.

I'm sure there is a reason we have two separate IDs but I'm just not sure what it is 😅

benjackwhite (commented):
I'm happy that everything works, especially if @liyiy has tested as well, but I still have some question marks around some of the persisted values. My gut tells me we are inheriting bad design from the posthog-js which we don't need as we are adding these things fresh but I could be wrong there!

Regarding tests, it seems a shame not to have them working as perhaps there is a genuine bug we are ignoring if they sometimes fail...

PostHog/Classes/PHGPayloadManager.m
EDsCODE: Sorry, I meant follow up PR as a stacked PR so the changes would be easier to see. I just brought in those changes though so they're all here! This is cleaned up now

PostHog/Classes/PHGPayloadManager.m
EDsCODE: The primary reason I adhered to shifting old distinct_id to anon_distinct_id when id'ing is to satisfy feature flag continuity if necessary. We recently adjusted the logic such that anon_distinct_id is also passed during feature flag calls to track when experience continuity needs to happen. Apart from that, I would like to try to keep the client libraries in alignment. This seemed like a simple approach to do that but definitely isn't following posthog-js perfectly

benjackwhite (approved):

PR #33, opened 2022-10-28
adds PHGPostHogFeatureFlagsDidLoadNotification

chrisladd:
Big fan, first time caller. I'm starting to use PostHog's experiments, and wanted an easy way to know which group users fell into. In my testing, all users were `control` at launch, and then were assigned to their groups some time later as data is fetched from the server.

**What does this PR do?**

Allows clients to delay checking for feature flags until they've loaded, as well as listen for relevant updates.

I tried to fit in with the current style of the project, but am glad to make any stylistic changes.

On launch, or after calling `reloadFeatureFlags`, clients should expect the `PHGPostHogFeatureFlagsDidLoadNotification`.

Questions: 

- do we want to do any kind of equality checking to only post this notification if there are changes to feature flags? If so, any suggestions?
- do we want to give the notification a prettier name for Swift clients
- if so, do we want to update or alias the `PHGPostHogIntegrationDidStart` notification to better match, for example providing a `Notification` suffix so the compiler automatically creates an NSNotification.Name for us.

**Questions:**

Perhaps—if you all agree with the change, it would be useful to let clients know to listen.

### DIFF
--- a/PostHog/Classes/PHGNotificationNames.h
+++ b/PostHog/Classes/PHGNotificationNames.h
@@ -1,0 +1,16 @@
+//
+//  PHGNotificationNames.h
+//  
+//
+
+#import <Foundation/Foundation.h>
+
+/**
+ * NSNotification name, that is posted after integrations are loaded.
+ */
+extern NSString *_Nonnull PHGPostHogIntegrationDidStart;
+
+/**
+ * NSNotification name posted after feature flags have been successfully loaded.
+ */
+extern NSString *_Nonnull PHGPostHogFeatureFlagsDidLoadNotification;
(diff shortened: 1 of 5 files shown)


benjackwhite:
Just noticed this (we had few eyes on this repo as Christmas approached and passed). I'm working on this SDK atm so will check this out ASAP and get it in as it looks good at a glance 👍

benjackwhite (approved):

marandaneto:
Closed because of the new major version that should have addressed this use case (via callback and/or notification)
https://github.com/PostHog/posthog-ios/pull/75

PR #34, opened 2023-01-24
Update alias doc strings

lharries:
**What does this PR do?**

Update the alias doc string to match how aliasing actually works. It was implying the wrong order

### DIFF
--- a/PostHog/Classes/PHGPostHog.h
+++ b/PostHog/Classes/PHGPostHog.h
@@ -116,10 +116,8 @@ NS_ASSUME_NONNULL_BEGIN
 
  @abstract
  Merge two user identities, effectively connecting two sets of user data as one.
- This may not be supported by all integrations.
 
- @param alias         The new ID you want to alias the existing ID to. The existing ID will be either the
- previousId if you have called identify, or the anonymous ID.
+ @param alias         The alias ID that you want to attach to the user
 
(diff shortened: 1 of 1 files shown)


PostHog/Classes/PHGPostHog.h
```diff
@@ -116,10 +116,8 @@ NS_ASSUME_NONNULL_BEGIN
 
  @abstract
  Merge two user identities, effectively connecting two sets of user data as one.
- This may not be supported by all integrations.
 
- @param alias         The new ID you want to alias the existing ID to. The existing ID will be either the
- previousId if you have called identify, or the anonymous ID.
+ @param alias         The alias ID that you want to attach to the user
```
tiina303, line 120: let's be more clear here - this needs to be the anonymous ID (that hasn't been merged with anything else). Please see https://posthog.com/docs/integrate/server/node#alias for example. what "alias ID" is is not clear.

lharries: I didn't realize this was the case, I've asked a question here: https://github.com/PostHog/posthog-python/pull/82#discussion_r1085869479

marandaneto:
Closed in favor of https://github.com/PostHog/posthog-ios/releases/tag/3.0.0-alpha.2

PR #37, opened 2023-02-01
feat: feature flag payloads support

EDsCODE:
**What does this PR do?**

- https://github.com/PostHog/posthog/issues/13627
- Extends the spec here: getFeatureFlag<type>Payload so that the return will be typed as expected

**Questions:**

### DIFF
--- a/PostHog/Classes/PHGPostHog.m
+++ b/PostHog/Classes/PHGPostHog.m
@@ -312,6 +312,13 @@ NSString *const PHGBuildKeyV2 = @"PHGBuildKeyV2";
     return isFlagEnabled;
 }
 
+- (id)getFeatureFlagPayload:(NSString *)flagKey
+{
+    NSDictionary *payloads = [self.payloadManager getFlagVariants];
+    id payload = [payloads valueForKey:flagKey];
+    return payload
+}
+
 - (void)reloadFeatureFlags
 {
     [self run:PHGEventTypeReloadFeatureFlags payload:nil];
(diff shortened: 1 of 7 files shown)


PostHog/Classes/PHGPostHog.m
```diff
@@ -312,6 +311,37 @@ - (bool)isFeatureEnabled:(NSString *)flagKey
     return isFlagEnabled;
 }
 
+- (NSString *)getFeatureFlagStringPayload:(NSString *)flagKey
```
EDsCODE, line 314: Not 100% confident on this

PostHog/Classes/PHGPostHog.m
```diff
@@ -312,6 +311,71 @@ - (bool)isFeatureEnabled:(NSString *)flagKey
...
 
+- (NSString *)getFeatureFlagStringPayload:(NSString *)flagKey defaultValue:(NSString *)defaultValue
+{
+    NSDictionary *payloads = [self.payloadManager getFeatureFlagPayloads];
+    NSString *payload = [payloads valueForKey:flagKey];
+    
+    if ([payload isKindOfClass:[NSString class]]) {
+        return payload;
+    } else {
+        NSLog(@"Could not retrieve value of type: NSString");
```
benjackwhite, line 322: I think we might want to prefix this with `[PostHog]` or something.

benjackwhite: Also, asking because I haven't followed the overall upgrade much, will there always be a payload? If the person isn't in capture group then they will likely have `nil` here right? Then we log an issue when there isn't really an issue 🤔

benjackwhite (approved):
Looks good assuming my comments are to be ignored 👍

PostHog/Classes/PHGPostHog.m
EDsCODE: Good catch!

PR #54, opened 2023-05-31
Stop using UIScreen.mainScreen to fix accent color

### LINKED ISSUE #53: Initializing PostHog in a SwiftUI app results in accent color not being respected
Similar to https://github.com/getsentry/sentry-cocoa/issues/2741, there is an issue when using PostHog in a SwiftUI app. Steps to reproduce:

- Make a new SwiftUI app in Xcode and customise the AccentColor in the Assets catalog to pink.
- Run the app and see the accent colour is pink
- Add PostHog as a Swift package to the project.
- Add an init to the top-level `App` struct with the following code:
`PHGPostHog(configuration: .init(apiKey: ""))`
- Run the app again. Notice that the accent colour is now blue.

The fix is the same as Sentry [implemented](https://github.com/getsentry/sentry-cocoa/pull/2762) and stop [using](https://github.com/PostHog/posthog-ios/blob/431bb9a78b2dc7f8c3cd2a5da07e278bb4f6221b/PostHog/Internal/PHGPostHogIntegration.m#L152) the now deprecated `UIScreen.main`. The simplest solution if there isn't access to an actual view is to do this via UIApplication's `windows` property.

pixlwave:
**What does this PR do?**
`UIScreen.mainScreen` is deprecated and SwiftUI apps lose their accent colour if the screen is accessed this way before the app's body is evaluated. This PR switches to reading the screen size from the app's first window and fixes #53.

Test it against a sample app.

Read the steps to reproduce the issue in #53 and test this PR against that.

No

#53

**Questions:**
- Does the docs need an update? No
- Are there any security concerns? No
- Do we need to update engineering / success? No

### DIFF
--- a/PostHog/Internal/PHGPostHogIntegration.m
+++ b/PostHog/Internal/PHGPostHogIntegration.m
@@ -149,9 +149,13 @@ static CTTelephonyNetworkInfo *_telephonyNetworkInfo;
     dict[@"$os_name"] = device.systemName;
     dict[@"$os_version"] = device.systemVersion;
 
-    CGSize screenSize = [UIScreen mainScreen].bounds.size;
-    dict[@"$screen_width"] = @(screenSize.width);
-    dict[@"$screen_height"] = @(screenSize.height);
+    // Access the screen via a window as UIScreen.mainScreen is deprecated
+    // and using it messes with the accent color in SwiftUI apps.
+    UIScreen *appScreen = UIApplication.sharedApplication.windows.firstObject.screen;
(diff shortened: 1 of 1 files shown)


benjackwhite:
Going to check this out locally as I don't think this generally works. I believe that `sharedApplication` is not available in extensions which we technically currently support.

pixlwave:
Ah yeah good point, didn't think about that 🤔

pixlwave:
Closing in favour of #55.

PR #63, opened 2023-09-04
Update device type

bddq:
**What does this PR do?**
This updates the device type property sent by SDK to conform examples seen in documentation.

Just check `PHGPostHogIntegration` file.

Trigger any event and check that `$device_type` is set with one of the implemented values.

Old value "ios" doesn't represent the a device type. Even less for tvOS apps.


bddq:
Any news @marandaneto?

marandaneto:
> Any news @marandaneto?

Sorry for the delay, there's a similar issue here https://github.com/PostHog/posthog-flutter/issues/49
I believe the change makes sense, but not sure if using the current approach.

I've been doing this with compile flags, eg.

```
#if TARGET_OS_OSX || TARGET_OS_MACCATALYST
    ..
#elif TARGET_OS_IOS
    ..
#elif TARGET_OS_TV
    ..
#elif TARGET_OS_WATCH
    ..
#endif
```

They come from https://github.com/apple/swift-corelibs-foundation/blob/main/CoreFoundation/Base.subproj/SwiftRuntime/TargetConditionals.h IIRC, would you mind testing this out?

but never used `userInterfaceIdiom`.

> Constants that indicate the interface type for the device or an object that has a trait environment, such as a view and view controller.

So it might be that this gives a false positive, what do you think?

bddq:
Conditional flags mentioned can work but cannot identify a CarPlay device (maybe PostHog will support that).

Because `userInterfaceIdiom` is read from the current device, it's safe to rely on that value to determine the device type.
It can also help the user to differentiate an iPhone and an iPad that are two types of iOS devices.

marandaneto:
> Conditional flags mentioned can work but cannot identify a CarPlay device (maybe PostHog will support that).
> 
> Because `userInterfaceIdiom` is read from the current device, it's safe to rely on that value to determine the device type. It can also help the user to differentiate an iPhone and an iPad that are two types of iOS devices.

That's also possible with the `$device_model`, so I'd be fine with both solutions.
@benjackwhite anything to add? Since this is a breaking change and add new values that are yet not documented, eg `CarPlay`.

marandaneto (approved):

bddq:
@marandaneto Done!

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 - CoreTelephony should not be added to tvOS builds [#67](https://github.com/PostHog/posthog-ios/pull/67)
 - Remote notifications methods do not throw if no default implementation [#67](https://github.com/PostHog/posthog-ios/pull/67)
+- Update device type [#63](https://github.com/PostHog/posthog-ios/pull/63)
```
marandaneto, line 5: Add as a `## Next` on the top since 2.0.4 is already published.

Eg:

## Next

- Update device type [#63](https://github.com/PostHog/posthog-ios/pull/63)

marandaneto: I tried to push the fix into your branch but its protected.

bddq: Updated. I'll check about branch protection.

PR #69, opened 2023-09-13
Prevent the accidental generation of multiple anon_distinct_id, causi…

inbreaks:
Prevent the accidental generation of multiple anon_distinct_id, causing unnecessary increase in person


marandaneto:
@inbreaks thanks for the PR, could we add some tests to avoid regression in the future as well?

marandaneto:
Closed because of the new major version that should have addressed this use case
https://github.com/PostHog/posthog-ios/pull/75

PR #71, opened 2023-10-03
Add loadFeatureFlagsOnStart configuration flag

marioradonic:
**What does this PR do?**

It adds a configuration flag similar to JS library's `advanced_disable_feature_flags_on_first_load` with which a first load of feature flags can be disabled.

Not much to this PR. Variable name can be discussed and I'm not sure if this is the best way to write the test, but I've tried to follow current structure. I didn't wanna name the variable the same as it's in JS lib since in iOS we don't have a "page load" concept.

I tested it manually by intercepting http requests using Charles proxy and confirming that /decide call is not made.

In our project we want to be in control of when to reload feature flags. The flag is true by default so it's backward compatible.

/

/

**Questions:**

Could be useful, but not my job.

I don't see any.

/


PostHog/Classes/PHGPostHogConfiguration.h
```diff
@@ -123,6 +123,13 @@ typedef NSMutableURLRequest *_Nonnull (^PHGRequestFactory)(NSURL *_Nonnull);
  */
 @property (nonatomic, assign) BOOL shouldSendDeviceID;
 
+/**
+ * Whether the posthog client should load feature flags when initialized.
+ * Changing the value of this property after initializing the client will have no effect.
+ * The default value is `YES`.
+ */
+@property (nonatomic, assign) BOOL loadFeatureFlagsOnStart;
```
marandaneto, line 131:
```suggestion
@property (nonatomic, assign) BOOL preloadFeatureFlags;
```

Let's call it `preloadFeatureFlags` as the other SDKs.

PostHog/Classes/PHGPostHogConfiguration.h
```diff
@@ -123,6 +123,13 @@ typedef NSMutableURLRequest *_Nonnull (^PHGRequestFactory)(NSURL *_Nonnull);
  */
 @property (nonatomic, assign) BOOL shouldSendDeviceID;
 
+/**
+ * Whether the posthog client should load feature flags when initialized.
+ * Changing the value of this property after initializing the client will have no effect.
```
marandaneto, line 128: We can add a comment that they can be reloaded by calling the `reloadFeatureFlags` method.

PostHog/Classes/PHGPostHogConfiguration.h
marioradonic: Updated.

PostHog/Classes/PHGPostHogConfiguration.h
marioradonic: Updated.

marandaneto (approved):

PR #73, opened 2023-10-06
`$active_feature_flags` should filter non active flags

marandaneto:
**What does this PR do?**
Similar to this one https://github.com/PostHog/posthog-android/pull/41
Closes https://github.com/PostHog/posthog-android/issues/40

there is no unit test at all for that, I will add in the new version of the iOS SDK, but for now this should be fine.

decide v3 returns all the feature flags and not only the active ones

**Questions:**


neilkakkar (approved):
A test or two would be nice, and give me confidence that the type conversion works as expected, but otherwise LGTM!

marandaneto:
> A test or two would be nice, and give me confidence that the type conversion works as expected, but otherwise LGTM!

I made sure to run locally on an emulator, so it's fine, let me check if it's easy to set up a mock HTTP cal, etc.

neilkakkar:
not a requirement! happy as long as it works, given you're going to revamp soon

PR #74, opened 2023-10-08
isFeatureEnabled now returns false if disabled

marandaneto:
**What does this PR do?**

`isFeatureEnabled` should return `false` if feature flag is disabled

`PHGPostHog.m`

Demo project has one flag for each case, you can just call `isFeatureEnabled` with them.

Relates to https://github.com/PostHog/posthog-android/pull/42

**Questions:**

### DIFF
--- a/Examples/CarthageExample/CarthageExample/AppDelegate.swift
+++ b/Examples/CarthageExample/CarthageExample/AppDelegate.swift
@@ -2,7 +2,7 @@ import UIKit
 import PostHog
 
 // Use your own apiKey!
-let PostHog = PHGPostHog(configuration: PHGPostHogConfiguration(apiKey: "b2XoQDHA5YhWQAjr2FZ4qIckgyEqXyY8"))
+let PostHog = PHGPostHog(configuration: PHGPostHogConfiguration(apiKey: "_6SG-F7I1vCuZ-HdJL3VZQqjBlaSb1_20hDPwqMNnGI"))
 
 @UIApplicationMain
 class AppDelegate: UIResponder, UIApplicationDelegate {
--- a/PostHog/Classes/PHGPostHog.m
+++ b/PostHog/Classes/PHGPostHog.m
@@ -313,8 +313,18 @@ NSString *const PHGBuildKeyV2 = @"PHGBuildKeyV2";
 
 - (bool)isFeatureEnabled:(NSString *)flagKey options:(NSDictionary *)options
 {
-    NSArray *keys = [self.payloadManager getFeatureFlags];
-    BOOL isFlagEnabled = [keys containsObject: flagKey];
+    NSDictionary *flags = [self.payloadManager getFlagVariants];
+
+    bool isFlagEnabled = true;
+    id value = [flags valueForKey:flagKey];
+    
+    if (value != nil) {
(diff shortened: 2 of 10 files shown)


Examples/CarthageExample/CarthageExample/AppDelegate.swift
```diff
@@ -2,7 +2,7 @@ import UIKit
 import PostHog
 
 // Use your own apiKey!
-let PostHog = PHGPostHog(configuration: PHGPostHogConfiguration(apiKey: "b2XoQDHA5YhWQAjr2FZ4qIckgyEqXyY8"))
+let PostHog = PHGPostHog(configuration: PHGPostHogConfiguration(apiKey: "_6SG-F7I1vCuZ-HdJL3VZQqjBlaSb1_20hDPwqMNnGI"))
```
marandaneto, line 5: Defaults to "Demo".

PostHogTests/FeatureFlagTests.swift
```diff
@@ -27,13 +27,24 @@ class FeatureFlagTests: QuickSpec {
     it("checks flag is enabled") {
       _ = stubRequest("POST", "https://app.posthog.test/decide/?v=3" as LSMatcheable)
         .andReturn(200)?
-        .withBody("{\"featureFlags\":{\"some-flag\":\"true\"}}" as LSHTTPBody);
+        .withBody("{\"featureFlags\":{\"some-flag\": true}}" as LSHTTPBody);
```
marandaneto, line 30: String escaping was making the test wrong, fixed all of them.

PostHogTests/PostHogTests.swift
```diff
@@ -97,22 +97,22 @@ class PostHogTests: QuickSpec {
 
     it("fires Application Opened during UIApplicationWillEnterForeground") {
       testMiddleware.swallowEvent = true
-      NotificationCenter.default.post(name: UIApplication.willEnterForegroundNotification, object: testApplication)
+      NotificationCenter.default.post(name: NSNotification.Name.UIApplicationWillEnterForeground, object: testApplication)
```
marandaneto, line 100: `UIApplication.x` is deprecated in favor of `NSNotification.Name.x`

PostHog/Classes/PHGPostHog.m
```diff
@@ -313,8 +313,18 @@ - (bool)isFeatureEnabled:(NSString *)flagKey
...
+    bool isFlagEnabled = true;
+    id value = [flags valueForKey:flagKey];
+    
+    if (value != nil) {
+        if ([value isKindOfClass:[NSNumber class]]) {
+            isFlagEnabled = [value boolValue];
+        }
+    } else {
+        isFlagEnabled = false;
+    }
```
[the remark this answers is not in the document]
marandaneto: @benjackwhite context is her https://github.com/PostHog/posthog-android/pull/42#issuecomment-1751966655
If a flag is `true` or any string, it returns `true`.
if a flag is `false`, it should return `false`, which is not the case before this change.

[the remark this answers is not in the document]
benjackwhite: Ah fair enough with that context. It was more to say "currently the API never returns `false` - it just isn't defined"

benjackwhite (approved):

PR #78, opened 2023-10-23
Add ObjC example and add method annotations

marandaneto:
**What does this PR do?**

Add ObjC example and add method annotations
Otherwise ObjC calls would be super long and repetitive.

**Questions:**

### DIFF
--- a/PostHog/PostHogFeatureFlags.swift
+++ b/PostHog/PostHogFeatureFlags.swift
@@ -34,7 +34,7 @@ class PostHogFeatureFlags {
         distinctId: String,
         anonymousId: String,
         groups: [String: String],
-        completion: @escaping ([String: Any]?, [String: Any]?) -> Void
+        callback: @escaping () -> Void
     ) {
         isLoadingLock.withLock {
             if self.isLoadingFeatureFlags {
@@ -53,7 +53,7 @@ class PostHogFeatureFlags {
                 else {
                     hedgeLog("Error: Decide response missing correct featureFlags format")
                     self.setLoading(false)
-                    return completion(nil, nil)
+                    return callback()
                 }
                 let errorsWhileComputingFlags = data?["errorsWhileComputingFlags"] as? Bool ?? false
 
@@ -80,7 +80,7 @@ class PostHogFeatureFlags {
 
                 self.setLoading(false)
 
-                return completion(featureFlags, featureFlagPayloads)
+                return callback()
             }
         }
     }
@@ -94,13 +94,13 @@ class PostHogFeatureFlags {
         return flags
     }
 
-    func isFeatureEnabled(_ flagKey: String) -> Bool {
+    func isFeatureEnabled(_ key: String) -> Bool {
         var flags: [String: Any]?
         featureFlagsLock.withLock {
             flags = self.storage.getDictionary(forKey: .enabledFeatureFlags) as? [String: Any]
         }
 
-        let value = flags?[flagKey]
+        let value = flags?[key]
 
         if value != nil {
             let boolValue = value as? Bool ?? false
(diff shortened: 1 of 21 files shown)


PostHog/PostHogFeatureFlags.swift
```diff
@@ -34,7 +34,7 @@ class PostHogFeatureFlags {
         distinctId: String,
         anonymousId: String,
         groups: [String: String],
-        completion: @escaping ([String: Any]?, [String: Any]?) -> Void
+        callback: @escaping () -> Void
```
marandaneto, line 37: Since we have special logic to check if the feature flag is enabled and so on, it's better to not expose the raw dictionary, otherwise it's too easy to misuse the API, I missed that before.

PostHog/PostHogSDK.swift
```diff
@@ -315,7 +372,12 @@ let maxRetryDelay = 30.0
         ))
     }
 
-    @objc public func alias(_ alias: String, properties: [String: Any]? = nil) {
+    @objc public func alias(_ alias: String) {
+        self.alias(alias, properties: nil)
+    }
+
+    @objc(aliasWithAlias:properties:)
```
marandaneto, line 379: `aliasWithAlias` is repeating because its the method name and param, but I don't have a better idea?

PostHog/PostHogSDK.swift
```diff
@@ -363,7 +425,7 @@ let maxRetryDelay = 30.0
         return mergedGroups ?? [:]
     }
 
-    @objc public func groupIdentify(type: String, key: String, groupProperties: [String: Any]? = nil) {
+    private func groupIdentify(type: String, key: String, groupProperties: [String: Any]? = nil) {
```
marandaneto, line 428: I'd rather keep it as a `group` only as the other SDKs.

PostHog/PostHogFeatureFlags.swift
benjackwhite: Agreed

PostHog/PostHogSDK.swift
```diff
@@ -219,7 +219,9 @@ let maxRetryDelay = 30.0
         return props
     }
 
-    @objc public func register(_ properties: [String: Any]) {
+    // register is a reserved word in ObjC
+    @objc(registerProperty:)
```
benjackwhite, line 223: Would it not be plural? 
```suggestion
    @objc(registerProperties:)
```

PostHog/PostHogSDK.swift
```diff
@@ -232,18 +234,39 @@ let maxRetryDelay = 30.0
...
+                         userProperties: [String: Any]? = nil)
+    {
+        identify(distinctId, properties: properties, userProperties: userProperties, userPropertiesSetOnce: nil)
+    }
+
+    @objc(identifyWithDistinctId:properties:userProperties:userPropertiesSetOnce:)
+    public func identify(_ distinctId: String,
+                         properties _: [String: Any]? = nil,
+                         userProperties: [String: Any]? = nil,
+                         userPropertiesSetOnce: [String: Any]? = nil)
```
benjackwhite, line 269: Typically in every other SDK we don't allow you to modify the normal properties of the `$identify` event. Mostly because it has special behaviour. I would remove it here, or at the very least make it the last argument in the chain, not the first.

PostHog/PostHogSDK.swift
```diff
@@ -315,7 +372,12 @@ let maxRetryDelay = 30.0
         ))
     }
 
-    @objc public func alias(_ alias: String, properties: [String: Any]? = nil) {
+    @objc public func alias(_ alias: String) {
+        self.alias(alias, properties: nil)
+    }
+
+    @objc(aliasWithAlias:properties:)
+    public func alias(_ alias: String, properties: [String: Any]? = nil) {
```
benjackwhite, line 380: Similar to $identify, I'm not sure we actually want to expose the other properties.

benjackwhite (changes requested):
Some change requests now that I've had time to think about the methods properly.

PostHog/PostHogSDK.swift
marandaneto: Good point

PostHog/PostHogSDK.swift
marandaneto: Mmm, I've taken this from https://github.com/PostHog/posthog-js-lite/blob/9ac7090d676dc2a6e5688cbe2b0ec001e2a23088/posthog-core/src/index.ts#L149 which allow you passing the properties

> Mostly because it has special behaviour

I don't have context about that, so I don't know what the right call here is, happy to remove/reorder.
JS SDK does not have it, should I remove it then since JS is the single source of truth?

PostHog/PostHogSDK.swift
marandaneto: I agree to remove it from here, JS and JS-Lite don't have it, I probably just thought that it'd be useful.

benjackwhite (approved):

PR #79, opened 2023-10-24
Modify cocoapods spec to make it work with v3

marandaneto:
**What does this PR do?**

**Questions:**


Podfile
```diff
@@ -1,37 +0,0 @@
-# Uncomment the next line to define a global platform for your project
-platform :ios, '11.0'
```
marandaneto, line 2: We use SPM for our own SDK now

PostHog/PostHogDataMode.swift
```diff
@@ -1,14 +0,0 @@
-//
```
marandaneto, line 1: That as a dupe

benjackwhite (approved):

PR #80, opened 2023-10-24
Swift PM fixes and Swift only

marandaneto:
**What does this PR do?**

SPM isn't happy with mixed languages, so we could either split the ObjC bits (NSData) in a new package or replace it altogether, I choose the later.

**Questions:**


benjackwhite (approved):
Nice cleaning here. Love that we have an example project for every target

PR #82, opened 2023-10-25
Sanitize dictionaries before serialization

marandaneto:
**What does this PR do?**

Sanitize dictionaries before serialization

I've tried multiple approaches for proper sanitization but all of them have edge cases or get overly complex.
Similar thoughts as here https://github.com/PostHog/posthog-ios/blob/19064a4f25e74e50604b9954d2e0a18f779d7397/PostHog/Models/PostHogEvent.swift#L33-L34

`Codable` types only would be awesome but the interoperability with ObjC is suboptimal.

Using Generics would be ideal as well but again bad interoperability with ObjC.
 I guess we can improve this in the future if we see it but I'd not write tons of code right now, the older version relies on`JSONSerialization` as well.

I checked https://github.com/SwiftyJSON/SwiftyJSON/blob/master/Source/SwiftyJSON/SwiftyJSON.swift
It just fails if there's a nonserializable type.
Also, https://swiftunboxed.com/stdlib/json-decoder-decodable/ but generics are not well played.
Tried to implement my own version of `AnyCodable` as well but entered the rabbit whole of https://github.com/asensei/AnyCodable/tree/master

> We came to the conclusion that there are too many edge cases, and we have since then completely moved away from fuzzy decoding of Any type.

`JSONSerialization.isValidJSONObject` always requires an Object or Array, so I cannot just use it for each value of the dict.

**Questions:**


PostHog/PostHogSDK.swift
```diff
@@ -331,10 +335,10 @@ let maxRetryDelay = 30.0
...
             event: event,
             distinctId: getDistinctId(),
-            properties: buildProperties(properties: properties,
-                                        userProperties: userProperties,
-                                        userPropertiesSetOnce: userPropertiesSetOnce,
-                                        groupProperties: groupProperties)
+            properties: buildProperties(properties: sanitizeDicionary(properties),
+                                        userProperties: sanitizeDicionary(userProperties),
+                                        userPropertiesSetOnce: sanitizeDicionary(userPropertiesSetOnce),
+                                        groupProperties: sanitizeDicionary(groupProperties))
```
benjackwhite, line 341: Could this be done in PostHogEvent instead? Feels like something that could easily get forgotten about, not to mention the code duplication

PostHog/PostHogSDK.swift
```diff
@@ -723,4 +733,12 @@ let maxRetryDelay = 30.0
...
         capture("Application Backgrounded")
     }
+
+    private func sanitizeDicionary(_ dict: [String: Any]?) -> [String: Any]? {
+        if dict == nil || dict!.isEmpty || !JSONSerialization.isValidJSONObject(dict!) {
+            return nil
+        }
+
+        return dict
+    }
```
benjackwhite, line 743: 1. If it isn't serializable, we probably want to hedgeLog something about that for debugging purposes ("why are my events missing properties??")
2. We could do this on a per-property basis, filtering out any properties that are not json serializable. This would allow some properties to get through fine (`$screen_name` for example) even if one property was not serializable.
3. If we don't do 2. then I think we need to actually drop the whole event. Otherwise, it is likely events will come through with completely meaningless properties which feels worse than coming through with only one property missing
4. We could even add a special property like `$serialization_error_for_keys: ['my_bad_property']` so that we can indicate in the UI that something went wrong

benjackwhite (commented):
Some thoughts on how we should best handle this, thinking ahead to what the feedback will be if someone gets into this situation.

PostHog/PostHogSDK.swift
marandaneto: The reason for not doing this on a top-level function is that it is an all-in or nothing, in this case, if there's something messed up in `groupProperties`, just `groupProperties` will be dropped but everything else is fine.

PostHog/PostHogSDK.swift
marandaneto: 1. good point, will add logging.
2. that won't happen since I only call `sanitizeDicionary` for the given properties, not the ones generated by the SDK.
eg:

```
let props = [
            "$screen_name": screenTitle,
        ].merging(sanitizeDicionary(properties) ?? [:]) { prop, _ in prop }
```

3. I agree, there's just no easy way to check if every property is serializable or not in an easy way.
The only easy solution I found is to wrap a property in a list or dictionary and call `JSONSerialization.isValidJSONObject` but then there's a huge overhead, besides the need to be recursive.
I am open to ideas, all the other ideas compatible with ObjC and Swift would be tons of code or 3rd party libs, See `Any background context you want to provide?` section.

4. That would be awesome, but again, see the item 3.

marandaneto: For context, iOS SDK v2 does not do it either, just document via an annotation https://github.com/PostHog/posthog-ios/blob/8500ffd930fb53670f6ad8c10f43d2efe3321089/PostHog/Classes/PHGSerializableValue.h#L13 + that https://github.com/PostHog/posthog-ios/blob/8500ffd930fb53670f6ad8c10f43d2efe3321089/PostHog/Internal/PHGPostHogUtils.m#L107 which is very slow for nested properties.

benjackwhite: I don't think we need to do it nested, just top level properties would be fine and then I don't think the overhead should be that much (but happy to be proved wrong) to iterate 10 properties?

marandaneto: @benjackwhite If one of the nested properties is not serializable, the whole JSONSerialization fails anyway, so we have to do it, assuming that all Dicts are quite small, it should be fine, let me see if I can get something quickly without a huge amount of LoC that works for ObjC as well

marandaneto: @benjackwhite pushed some changes, I still only do for the given properties since the properties set by the SDK are always serializable, so this spares a bit of the overhead.
Also, only for high-level properties, if a nested dict or array isn't serializable, we drop it and log it out.
I also added the `URL` and `Date` parsing to be compatible with the older version.

benjackwhite (changes requested):
This looks great functionally but feels like something that 100% needs tests

benjackwhite (approved):

PR #83, opened 2023-10-27
Compatibility with iOS SDK v2

marandaneto:
**What does this PR do?**

Compatibility with iOS SDK v2, move the old queue to the new queue and delete the old files

The iOS SDK previously didn't support multiple SDKs running at the same time since there's no `apiKey` or `tag` within the caching path (well they would share the same folder and it'd be a mess).
If we want to add support for it, we can make it in another PR, but I'd not block the first release because of that since there's no support already.

**Questions:**

### DIFF
--- a/PostHog/PostHogQueue.swift
+++ b/PostHog/PostHogQueue.swift
@@ -45,7 +45,7 @@ class PostHogQueue {
         self.storage = storage
         self.api = api
         self.reachability = reachability
-        fileQueue = PostHogFileBackedQueue(url: storage.url(forKey: .queue))
+        fileQueue = PostHogFileBackedQueue(queue: storage.url(forKey: .queue), oldQueue: storage.url(forKey: .oldQeueue))
     }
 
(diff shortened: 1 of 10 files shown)


PostHog/PostHogQueue.swift
```diff
@@ -45,7 +45,7 @@ class PostHogQueue {
         self.storage = storage
         self.api = api
         self.reachability = reachability
-        fileQueue = PostHogFileBackedQueue(url: storage.url(forKey: .queue))
+        fileQueue = PostHogFileBackedQueue(queue: storage.url(forKey: .queue), oldQueue: storage.url(forKey: .oldQeueue))
```
benjackwhite, line 48: Personally I would choose to have the "legacy" logic as it's own standalone class, which takes the fileQueue as an argument. That way we can keep it clean and separate. Not going to block you so up to you if you agree.

Would just make it easier in the future to know what the code is for, and simpler to read.

benjackwhite (approved):
Approved as it looks like it likely works, just a suggestion left

PR #84, opened 2023-10-30
Unit tests and refactoring to make it testable

marandaneto:
**What does this PR do?**

Add tests for almost every class, also has a lot of fixing found during testing.

More tests can be added specifically for `PostHogQueue`, but its ok to add it later as well, its a bit complicated due to time advancing, and the `URLSession` callback approach, but I can expand it in another PR so we can unblock this PR that is getting quite big already.

**Questions:**


.github/workflows/build-examples.yml
```diff
@@ -0,0 +1,18 @@
+name: Build Examples
```
marandaneto, line 1: The separation speeds up the build due to running in parallel

PostHog/PostHogApi.swift
```diff
@@ -49,7 +49,7 @@ class PostHogApi {
 
         do {
             data = try JSONSerialization.data(withJSONObject: toSend)
-        } catch let error as NSError {
+        } catch {
```
marandaneto, line 52: No need to cast it, otherwise the casting might lose the precision

PostHog/PostHogFeatureFlags.swift
```diff
@@ -135,14 +145,15 @@ class PostHogFeatureFlags {
             return value
         }
 
-        // The payload value is stored as a string and is not pre-parsed...
-        // We need to mimic the JSON.parse of JS which is what posthog-js uses
-        let jsonData = try? JSONSerialization.jsonObject(with: stringValue.data(using: .utf8)!, options: .fragmentsAllowed)
-
-        if jsonData == nil {
-            return value
+        do {
```
marandaneto, line 148: the do/catch approach always gives us the chance to hedgeLog.

PostHog/PostHogQueue.swift
```diff
@@ -75,41 +73,51 @@ class PostHogQueue {
...
-            self.pausedLock.withLock {
-                if self.config.dataMode == .wifi, reachability.connection != .wifi {
-                    hedgeLog("Queue is paused because its not in WiFi mode")
-                    self.paused = true
-                } else {
-                    self.paused = false
+    func start(disableReachabilityForTesting: Bool,
+               disableQueueTimerForTesting: Bool)
+    {
+        if !disableReachabilityForTesting {
```
marandaneto, line 79: Those flags are always enabled but during tests the `reachability` and `timer` just give us a hard time during debugging, also because the reachability goes crazy, we'd need a deboucer just for testing.

PostHog/PostHogQueue.swift
```diff
@@ -157,36 +172,22 @@ class PostHogQueue {
...
-
-            if processing.isEmpty {
-                return
-            }
-
             completion(PostHogConsumerPayload(events: processing) { success in
                 hedgeLog("Completed!")
                 if success {
-                    _ = self.fileQueue.pop(processing.count)
+                    self.fileQueue.pop(items.count)
```
marandaneto, line 190: We pop the total items from the peek method, even if some of them failed to be parsed, so we don't run into the problem with the deletion, the index would be moving and its just harder.

PostHog/PostHogStorage.swift
```diff
@@ -144,20 +149,6 @@ class PostHogStorage {
         setJson(forKey: key, json: contents)
     }
 
-    public func getNumber(forKey key: StorageKey) -> Double? {
```
marandaneto, line 147: Some methods here were not used at all so I removed them

PostHogTests/TestUtils/MockPostHogServer.swift
```diff
@@ -14,73 +14,104 @@ import OHHTTPStubsSwift
...
+        decideRequests = []
+        batchExpectation = XCTestExpectation(description: "\(batchCount) batch requests to occur")
+        decideExpectation = XCTestExpectation(description: "1 decide requests to occur")
+        batchExpectationCount = batchCount
+        errorsWhileComputingFlags = false
+        return500 = false
     }
 
-    func parseBatchRequest(_ context: URLRequest) -> [String: Any]? {
+    func parseRequest(_ context: URLRequest, gzip: Bool = true) -> [String: Any]? {
```
marandaneto, line 107: I wonder if decide can be `gzip` as well? but likely not important since its just a small payload.

Package.swift
```diff
@@ -25,7 +28,13 @@ let package = Package(
...
         .testTarget(
             name: "PostHogTests",
-            dependencies: ["PostHog"],
+            dependencies: [
+                "PostHog",
+                "Quick",
+                "Nimble",
+                "OHHTTPStubs",
+                .product(name: "OHHTTPStubsSwift", package: "OHHTTPStubs"),
+            ],
```
marandaneto, line 37: Here is what had to be done to make it work in CI, in combination with `swift test` instead of the xcode command for some reason.

marandaneto: Xcode already defines those dependencies as SPM in its projects, but apparently, CI wasn't picking them up, maybe a Xcode version mismatch, an empty project worked just fine.

PostHog/PostHogSDK.swift
```diff
@@ -213,7 +226,7 @@ let maxRetryDelay = 30.0
     }
 
     private func getRegisteredProperties() -> [String: Any] {
-        guard let props = storage?.getDictionary(forKey: .registerProperties) as? [String: String] else {
+        guard let props = storage?.getDictionary(forKey: .registerProperties) as? [String: Any] else {
```
marandaneto, line 229: I am unsure if `$groups` is always `String: String` or `String: Any`, apparently its the former but JS does `String: Int` as well for some reason.

PostHogTests/TestUtils/MockPostHogServer.swift
benjackwhite: Nope decide can't be gzipped, only ingestion currently (although it sounds like we might move away from that)

PostHog/PostHogSDK.swift
benjackwhite: I guess we don't stop people from passing a number like `{ project: 123 }` but I'm not sure what actual effect that has downstream...

benjackwhite (approved):
Mostly skimmed the tests, but the actual functional changes all seem reasonable. Mammoth effort!

PostHog/PostHogSDK.swift
marandaneto: Will keep `String: Any` for safety and let ingestion decide if its ok or not, the SDK only sets `String: String` anyway, this would happen if somebody passes `$groups` manually.

PR #85, opened 2023-11-02
New usage md and breaking changes list

marandaneto:
**What does this PR do?**

_#skip-changelog_

**Questions:**


benjackwhite (approved):
Traditionally we have always pushed people to our docs rather than including instructions in github - trying to avoid duplication.
Given this is pre-release though, makes sense and then we can make a call after release

marandaneto:
> Traditionally we have always pushed people to our docs rather than including instructions in github - trying to avoid duplication. Given this is pre-release though, makes sense and then we can make a call after release

I agree, but right now the docs are structured to only support stable versions unfortunately, no plans to add a dropdown versioning on the docs for now, so better to keep it here till we get to RC and/or stable.
I have a draft PR on the docs replacing the code snippets locally but it'd just cause issues for current customers if merged.

PR #89, opened 2023-11-10
Fix Tests compilation for watchOS

marandaneto:
_#skip-changelog_

Tests for watchOS were failing.
Also added a macOS example since the app delegate is different, also the targets.

## :pencil: Checklist


Package.swift
```diff
@@ -4,7 +4,7 @@ import PackageDescription
 let package = Package(
     name: "PostHog",
     platforms: [
-        .macOS(.v10_14), .iOS(.v13), .tvOS(.v13), .watchOS(.v6),
+        .macOS(.v10_15), .iOS(.v13), .tvOS(.v13), .watchOS(.v6),
```
marandaneto, line 7: Even our testing targets are min. 10.15 already, so we cannot even test unless we change our testing infra, since it's very old, it's ok to drop it.

benjackwhite (approved):
👍👍👍👍

Issue #97, opened 2024-01-19
`$session_id` support


marandaneto:
### Description

Android already supports the `$session_id`
iOS not yet, this will be done with iOS session recording

Relates to https://github.com/PostHog/posthog-ios/pull/32 but this PR is table since the SDK was rewritten.

mfclarke-cnx:
Hi @marandaneto, do you have an idea of when this will be shipped? We're very keen to make the switch, but the lack of session support is a total deal breaker for us.

marandaneto:
Hi, yes, this is planned for the next week.

PR #100, opened 2024-02-05
Add session tracking

mfclarke-cnx:
We would like to switch to PostHog, but the lack of session tracking on iOS was a dealbreaker. Although @marandaneto indicated it would be implemented this week in  #97 , I got curious after looking at the old PR.

This is still a draft PR. I would like feedback from @marandaneto if they would like me to continue to finish it (refine logic, add tests). Otherwise I'm happy for this to be closed and wait for the official thing.

Closes https://github.com/PostHog/posthog-ios/issues/97

Simply ran our app integrated with my local posthog-ios and saw that session ID appeared in the events, and I was able to track session related metrics https://us.posthog.com/shared/HeXYA1N8aHnvLpGfGrwpBzllJiswgw

I would like to add to the test suite if this PR is deemed workable.

## :pencil: Checklist

### DIFF
--- a/PostHog/PostHogSDK.swift
+++ b/PostHog/PostHogSDK.swift
@@ -15,7 +15,8 @@ import Foundation
 
 let retryDelay = 5.0
 let maxRetryDelay = 30.0
-private let sessionChangeThreshold: TimeInterval = 1800
+// 30 minutes in seconds
+private let sessionChangeThreshold: TimeInterval = 60 * 30
 
 // renamed to PostHogSDK due to https://github.com/apple/swift/issues/56573
 @objc public class PostHogSDK: NSObject {
--- a/PostHog/PostHogSDK.swift
+++ b/PostHog/PostHogSDK.swift
@@ -696,16 +696,16 @@ private let sessionChangeThreshold: TimeInterval = 60 * 30
                                       object: nil)
         #elseif os(macOS)
             defaultCenter.addObserver(self,
-                                      selector: #selector(handleAppOpened),
+                                      selector: #selector(handleAppDidFinishLaunching),
                                       name: NSApplication.didFinishLaunchingNotification,
                                       object: nil)
             // macOS does not have didEnterBackgroundNotification, so we use didResignActiveNotification
             defaultCenter.addObserver(self,
-                                      selector: #selector(captureAppBackgrounded),
+                                      selector: #selector(handleAppDidEnterBackground),
                                       name: NSApplication.didResignActiveNotification,
                                       object: nil)
             defaultCenter.addObserver(self,
-                                      selector: #selector(handleAppOpened),
+                                      selector: #selector(handleAppDidBecomeActive),
                                       name: NSApplication.didBecomeActiveNotification,
                                       object: nil)
         #endif
--- a/PostHog/PostHogSDK.swift
+++ b/PostHog/PostHogSDK.swift
@@ -40,6 +40,7 @@ private let sessionChangeThreshold: TimeInterval = 60 * 30
     #if !os(watchOS)
         private var reachability: Reachability?
     #endif
+    var now: () -> Date = { Date() }
     private var flagCallReported = Set<String>()
     private var featureFlags: PostHogFeatureFlags?
     private var context: PostHogContext?
@@ -158,7 +159,7 @@ private let sessionChangeThreshold: TimeInterval = 60 * 30
         if groups != nil, !groups!.isEmpty {
             properties["$groups"] = groups!
         }
-        
+
         var theSessionId: String?
         sessionLock.withLock {
             theSessionId = sessionId
@@ -375,11 +376,12 @@ private let sessionChangeThreshold: TimeInterval = 60 * 30
         guard let queue = queue else {
             return
         }
-        
+
         // If events fire in the background after the threshold, they should no longer have a sessionId
         if isInBackground && sessionId != nil,
            let sessionLastTimestamp = sessionLastTimestamp,
-           Date().timeIntervalSince1970 - sessionLastTimestamp > sessionChangeThreshold {
+           now().timeIntervalSince1970 - sessionLastTimestamp > sessionChangeThreshold
+        {
             sessionLock.withLock {
                 sessionId = nil
             }
@@ -611,14 +613,14 @@ private let sessionChangeThreshold: TimeInterval = 60 * 30
             return
         }
 
-        if Date().timeIntervalSince1970 - sessionLastTimestamp > sessionChangeThreshold {
+        if now().timeIntervalSince1970 - sessionLastTimestamp > sessionChangeThreshold {
             rotateSession()
         }
     }
 
     private func rotateSession() {
         let newSessionId = UUID().uuidString
-        let newSessionLastTimestamp = Date().timeIntervalSince1970
+        let newSessionLastTimestamp = now().timeIntervalSince1970
 
         sessionLock.withLock {
             sessionId = newSessionId
@@ -823,7 +825,7 @@ private let sessionChangeThreshold: TimeInterval = 60 * 30
         captureAppBackgrounded()
 
         sessionLock.withLock {
-            sessionLastTimestamp = Date().timeIntervalSince1970
+            sessionLastTimestamp = now().timeIntervalSince1970
         }
 
         isInBackground = true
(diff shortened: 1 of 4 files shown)


PostHog/PostHogSDK.swift
```diff
@@ -15,6 +15,7 @@ import Foundation
 
 let retryDelay = 5.0
 let maxRetryDelay = 30.0
+private let sessionChangeThreshold: TimeInterval = 1800
```
mfclarke-cnx, line 18: 30 mins is the change threshold from the older PR

marandaneto, line 18:
```suggestion
// 30 minutes in seconds
private let sessionChangeThreshold: TimeInterval = 60 * 30
```

PostHog/PostHogSDK.swift
```diff
@@ -99,7 +100,7 @@ let maxRetryDelay = 30.0
                 let optOut = theStorage.getBool(forKey: .optOut)
                 config.optOut = optOut ?? config.optOut
             }
-
+
```
marandaneto, line 103:
```suggestion

```

PostHog/PostHogSDK.swift
```diff
@@ -142,14 +145,18 @@ let maxRetryDelay = 30.0
 
     private func dynamicContext() -> [String: Any] {
         var properties = getRegisteredProperties()
-
+
```
marandaneto, line 148:
```suggestion

```

PostHog/PostHogSDK.swift
```diff
@@ -245,6 +252,22 @@ let maxRetryDelay = 30.0
...
+        storage?.setString(forKey: .sessionId, contents: newValue)
+    }
+    
+    private func getSessionLastTimestamp() -> TimeInterval? {
+        storage?.getDouble(forKey: .sessionLastTimestamp)
+    }
+    
+    private func setSessionLastTimestamp(_ newValue: TimeInterval) {
+        storage?.setDouble(forKey: .sessionLastTimestamp, contents: newValue)
+    }
```
marandaneto, line 270: This can be all in-memory, if the app is closed/killed by the OS, it's a brand new session.

For reference https://github.com/PostHog/posthog-android/blob/main/posthog-android/src/main/java/com/posthog/android/internal/PostHogLifecycleObserverIntegration.kt

PostHog/PostHogStorage.swift
```diff
@@ -144,7 +146,7 @@ class PostHogStorage {
         }
         return nil
     }
-
+
```
marandaneto, line 149:
```suggestion

```

marandaneto:
One main difference is that on Android I used a timer to end the session after 30 minutes of inactivity:

https://github.com/PostHog/posthog-android/blob/5138d79a25801574a964b98020776680f6b35282/posthog-android/src/main/java/com/posthog/android/internal/PostHogLifecycleObserverIntegration.kt#L101

It is also ok to do that once the app is back in the foreground, but if background events are captured while the app is in the background for more than 30 minutes, technically those events will be part of that session still but it's not correct, maybe we do the same approach as on Android so the behavior is the same?

PostHog/PostHogSDK.swift
```diff
@@ -578,6 +602,28 @@ let maxRetryDelay = 30.0
...
+            rotateSession()
+        }
+    }
+    
+    private func rotateSession() {
+        let sessionId = UUID()
+        let sessionLastTimestamp = Date()
+        
+        setSessionId(sessionId.uuidString)
+        setSessionLastTimestamp(sessionLastTimestamp.timeIntervalSince1970)
```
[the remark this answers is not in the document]
mfclarke-cnx: Good point, I will fix this

PostHog/PostHogSDK.swift
```diff
@@ -678,8 +724,13 @@ let maxRetryDelay = 30.0
             #endif
         }
     }
+    
+    @objc func handleAppStarted() {
```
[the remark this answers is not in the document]
mfclarke-cnx: I have a vague memory of app opened/foregrounded not being called on app start a while ago. I will double check this.

PostHog/PostHogSDK.swift
mfclarke-cnx: Ok great, that simplifies things :)

mfclarke-cnx:
Absolutely happy to address these things!

Re using a timer: on iOS they restrict how you can execute code in the background. There's no way to set a timer and guarantee that it will fire when you want it too. For some more info, see: https://forums.developer.apple.com/forums/thread/114859

So what I will do is set the sessionLastTime when the app is backgrounded, and clear it when the app is foregrounded. If a sessionLastTime exists and events are being logged, it will check sessionLastTime against the threshold and start a new session if needed.

This does beg the question: if events fire in the background after 30 mins, is it a new session? Or is there now no session?

And another question: I believe Flurry makes this configurable. So you can opt to not include background events as part of sessions (keeping session as a purely "user using the app" based construct).

mfclarke-cnx:
@marandaneto please take a look at my comment above and the supporting changes. Thank you!

PostHog/PostHogSDK.swift
```diff
@@ -367,6 +380,12 @@ let maxRetryDelay = 30.0
                                         userPropertiesSetOnce: sanitizeDicionary(userPropertiesSetOnce),
                                         groupProperties: sanitizeDicionary(groupProperties))
         ))
+
+        if !isInBackground {
+            sessionLock.withLock {
+                sessionLastTimestamp = Date().timeIntervalSince1970
+            }
+        }
```
marandaneto, line 388: You just need this when the app goes to the background/foreground, and not for every event, the inactivity only counts if the app is in the background.

marandaneto:
@mfclarke-cnx Good point with the background tasks.

if events are fired during the background within the 30-minute threshold, it's part of the session, after that, not anymore, so the session_id should not be set in this case.

I think caching the time that the app went to the background and checking this timestamp only when its back to the foreground is less performance overhead and simpler than doing this during every event.

PostHog/PostHogSDK.swift
```diff
@@ -151,6 +159,10 @@ let maxRetryDelay = 30.0
             properties["$groups"] = groups!
         }
 
+        if let sessionId = sessionId {
+            properties["$session_id"] = sessionId
+        }
```
marandaneto, line 164: Those checks have to be locked as well, because between the `sessionId` check and assigning from `sessionId`, the value might have changed already, the chance is almost nothing but it's possible so better to lock to avoid racy conditions that would be very hard to debug.

PostHog/PostHogSDK.swift
```diff
@@ -761,7 +812,12 @@ let maxRetryDelay = 30.0
         capture("Application Opened", properties: props)
     }
 
-    @objc func captureAppBackgrounded() {
+    @objc func handleAppDidEnterBackground() {
```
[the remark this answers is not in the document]
mfclarke-cnx: Ah yes you're right. I was thinking the background event would cause the timestamp to update, but the user could have `captureApplicationLifecycleEvents` false in which case backgrounding wouldn't cause an event to fire. Will update.

PostHog/PostHogSDK.swift
mfclarke-cnx: Yes you're right, we can just set it once on backgrounding. Will update.

marandaneto:
I made it a draft and removed draft com the title, since its a GH feature anyway.

mfclarke-cnx:
> @mfclarke-cnx Good point with the background tasks.
> 
> if events are fired during the background within the 30-minute threshold, it's part of the session, after that, not anymore, so the session_id should not be set in this case.
> 
> I think caching the time that the app went to the background and checking this timestamp only when its back to the foreground is less performance overhead and simpler than doing this during every event.

Right. And in addition, we need to check it if events fire and the app is in the background so we can nil the sessionId from that point on.

mfclarke-cnx:
I think this is in a good place now @marandaneto

marandaneto:
> I think this is in a good place now @marandaneto

awesome, just a listing issue:
PostHogSDK.swift:381:1: warning: (andOperator) Prefer comma over && in if, guard or while conditions.

and add a new entry to the [changelog](https://github.com/PostHog/posthog-ios/blob/main/CHANGELOG.md) -> 3.1.0 and ready to roll.

mfclarke-cnx:
> awesome, just a linting issue: PostHogSDK.swift:381:1: warning: (andOperator) Prefer comma over && in if, guard or while conditions.

Ah, I think your lint step is out of sync with the local swiftlint rules? This doesn't come up locally for me. Will fix now.

marandaneto (approved):

Issue #103, opened 2024-02-08
Batch HTTP 400


bddq:
### Version

3.1.0

### Steps to Reproduce

Trigger any event. 
Wait for send to the `/batch` endpoint.

### Expected Result

Expected HTTP 2xx code after send.
Expected to see some events in Activity feed.

### Actual Result

No logs are shown in the Xcode console.
No events are shown in Activity feed on website.

Debug with breakpoint indicates a response status code is always `400` (`PostHogApi.batch()`) without any error.

SDK is initialised with:

```
let configuration = PostHogConfig(apiKey: postHogKey, host: "https://eu.posthog.com")

configuration.captureApplicationLifecycleEvents = false
configuration.captureScreenViews = false
configuration.debug = true

PostHogSDK.shared.setup(configuration)
```

marandaneto:
@bddq I just checked it and it's working alright for me.
Usually, a wrong apiKey returns 401, but not 400.

What do you see in the logs if enabling `configuration.debug = true`?

bddq:
@marandaneto As mentioned, logs are not displayed at all in Xcode console.
I'm using Xcode 14.2.

I checked the API key used. I use the right one.
I'm still having an HTTP 400 error.

I upgraded to 3.1.1, same issue.

<img width="968" alt="Capture 2024-02-08 at 17 01 12" src="https://github.com/PostHog/posthog-ios/assets/342063/48f29238-6a8e-4047-b0a5-49ad85b5fb70">

marandaneto:
@bddq we had some DNS issues yesterday, not sure if its the same thing but I will double check.
https://status.posthog.com/incidents/qqslrk8p86hx

I will check why there are no logs as well.

bddq:
@marandaneto I got same behaviour with default host.

BTW, I tested the same source code on another device and it works.
Device with PostHog working: iPhone 11 Pro, iOS 15.6, WiFi.
Device with PostHog *not* working: iPhone SE (3rd gen), iOS 16.2, WiFi.

Both tests were done with PostHog SDK v3.1.1.

Can it be an issue with Gzip on iOS 16 or something else making the request bad?

marandaneto:
@bddq https://github.com/PostHog/posthog-ios/pull/104
After this PR we'd have better logging, not sure, `gzip` is wrapped in a try-catch, so it'd throw and return early instead of calling the API and returning 400.
Let me get this merged, so you can test and check the logs.

[...]

PR #106, opened 2024-02-09
fix ISO8601 formatter to always use the 24h format

marandaneto:
Closes https://github.com/PostHog/posthog-ios/issues/103

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 3.1.3 - 2024-02-09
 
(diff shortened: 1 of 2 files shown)


marandaneto:
@bddq are you able to test from a branch?
the format was `HH` already and not `hh` so I believe it could be the default timezone and locale.
I cannot change my emulator to 12h only for some reason, the settings are blocked.

bddq (approved):
The `locale` did the trick. It's working as expected on my side with a 12-hour time format.

PR #122, opened 2024-03-25
chore: session recording view tree off the main thread

marandaneto:
## :pencil: Checklist


marandaneto:
Not possible because most components will throw with `X (eg UITextView.isEditable) must be used from main thread only

Issue #127, opened 2024-04-12
Swift 6 Compatibility and class property warning


adilrc:
### Problem Statement

I'm using your SDK with Xcode 15.3, and I'm getting a warning about a static property not being concurrency-safe. Can you please update the SDK to make it Swift 6 compatible and fix the warning?

Here's the warning I'm getting in _PostHogSDK.swift:54_:
`Class property 'shared' is not concurrency-safe because it is not either conforming to 'Sendable' or isolated to a global actor; this is an error in Swift 6`

Thanks!

marandaneto:
@adilrc thanks for reporting, did you change any configuration to get this warning? I am also using Xcode 15.3 but I don't get this warning.

adilrc:
Yes I do have [strict concurrency checks](https://www.swift.org/documentation/concurrency/) enabled in my package along with Swift 5.10. 

```swift
.target(
  ...
  swiftSettings: [
    .enableUpcomingFeature("StrictConcurrency")
  ]
)
```

dehesa:
I believe @adilrc  made a small typo. The SPM setting is:

```swift
.enableExperimentalFeature("StrictConcurrency"),
```

You can control the type of strict concurrency by following the attribute with `=` and then the level; i.e. `minimal`, `targeted`, `complete`. The absence of any mean `complete`. For example:

```
.enableExperimentalFeature("StrictConcurrency=targeted"),
```

If I may offer some advice in how to go about removing the concurrency warnings and making `PostHogSDK` safer:
- Ideally you would make `PostHogSDK` `Sendable`. This way the SDK can cross thread boundaries. I already see you are using locks when setting up some functions; therefore you could mark the whole class `PostHogSDK` as `@unchecked Sendable`. However, I think you are forwarding mutations on places unprotected by locks (e.g. `getDistinctId()` calls `isEnabled()` which relies in the _unprotected_ `enabled` private variable).
- Since making `PostHogSDK` sendable is probably too much work. You could mark the static `PostHog.shared` constant as `nonisolated(unsafe)`. This will remove the warning for users including the SDK as a static library while not making any _untrue_ promises to the compiler. By the way, you don't need the self-executing closure there.

```swift
public class PostHogSDK: NSObject {
  @objc public nonisolated(unsafe) static let shared = PostHogSDK(PostHogConfig(apiKey: ""))
}
```

marandaneto:
[...]

PR #129, opened 2024-04-15
chore: silence shared warning for strict concurrency

marandaneto:
Closes https://github.com/PostHog/posthog-ios/issues/127
Draft PR trying to silence all warnings for the SDK as well https://github.com/PostHog/posthog-ios/pull/128
That will likely require a iOS SDK v4 with major changes, but not worth for now until we are able to drop older iOS/Swift versions support and start using async/await, all sendable, etc.
`nonisolated(unsafe)` would be better as well instead of the whole class but that requires Swift >= 5.10 sadly.
Edit: `nonisolated(unsafe)` is possible using a compile directive and checking the Swift version

## :pencil: Checklist


pauldambra (approved):
🚢

PR #133, opened 2024-05-20
allow anonymous id generation to be configurable

jimexist:

this makes it aligned with the js/ts sdk

swift test

## :pencil: Checklist


PostHog/PostHogConfig.swift
```diff
@@ -26,6 +26,7 @@ import Foundation
     @objc public var captureScreenViews: Bool = true
     @objc public var debug: Bool = false
     @objc public var optOut: Bool = false
+    @objc public var getDeviceId: ((UUID) -> UUID) = { uuid in uuid }
```
marandaneto, line 29: @jimexist this function is actually used for the `$device_id` property and not the `$anon_distinct_id` property (in JS).
if the intention is that the anonymous ID is configurable, it will make more sense to call it `getAnonymousId` instead

jimexist: fixed and pushed

marandaneto:
Thanks for the PR @jimexist 
Could you add a test to validate the behavior? PostHogSessionManagerTest would be the file I think

marandaneto:
For the lint issue: https://github.com/PostHog/posthog-ios/pull/134

PostHog/Replay/URLSessionSwizzler.swift
```diff
@@ -60,192 +60,192 @@
 
         /// Swizzles the `URLSession.dataTask(with:completionHandler:)` for `URLRequest`.
         class DataTaskWithURLRequestAndCompletion: MethodSwizzler<
-        @convention(c) (URLSession, Selector, URLRequest, CompletionHandler?) -> URLSessionDataTask,
+            @convention(c) (URLSession, Selector, URLRequest, CompletionHandler?) -> URLSessionDataTask,
```
marandaneto, line 63: If there was no changes here, you can rollback, the linter did it?

PostHogTests/PostHogSessionManagerTest.swift
```diff
@@ -45,5 +45,16 @@ class PostHogSessionManagerTest: QuickSpec {
 
             sut.reset()
         }
+
+        it("Can can accept id customization via config") {
```
marandaneto, line 49:
```suggestion
        it("Can accept id customization via config") {
```

PostHog/Replay/URLSessionSwizzler.swift
jimexist: > If there was no changes here, you can rollback, the linter did it?

all the changes to this file was done by `make swiftFormat`

PostHog/PostHogSessionManager.swift
```diff
@@ -22,7 +25,7 @@ class PostHogSessionManager {
             anonymousId = storage.getString(forKey: .anonymousId)
 
             if anonymousId == nil {
-                anonymousId = UUID().uuidString
+                anonymousId = idGen(UUID()).uuidString
```
marandaneto, line 28:
```suggestion
                anonymousId = config.getAnonymousId(UUID()).uuidString
```

Better to use `config.getAnonymousId` directly

PostHog/PostHogSessionManager.swift
```diff
@@ -12,8 +12,11 @@ class PostHogSessionManager {
 
     private let anonLock = NSLock()
     private let distinctLock = NSLock()
+    private let idGen: (UUID) -> UUID
+
     init(_ config: PostHogConfig) {
         storage = PostHogStorage(config)
+        idGen = config.getAnonymousId
```
marandaneto, line 19:
```suggestion

    init(_ config: PostHogConfig) {
        storage = PostHogStorage(config)
```

PostHog/Replay/URLSessionSwizzler.swift
jimexist: this is strange because the submitted one fails the ci.

i guess it might be due to versioning. let me rollback.

marandaneto: yeah could be because the CI is happy with the current version.

PostHog/PostHogSessionManager.swift
marandaneto: oh config wasnt cached as a global env., makes sense.

PostHogTests/PostHogSessionManagerTest.swift
```diff
@@ -45,5 +45,16 @@ class PostHogSessionManagerTest: QuickSpec {
...
+            let config = PostHogConfig(apiKey: "123")
+            let fixedUuid = UUID()
+            config.getAnonymousId = { _ in fixedUuid }
+            let sut = PostHogSessionManager(config)
+            let anonymousId = sut.getAnonymousId()
+            expect(anonymousId) == fixedUuid
```
marandaneto, line 55:
```suggestion
            expect(anonymousId) == fixedUuid.uuidString
```

marandaneto (approved):

PostHog/PostHogSessionManager.swift
jimexist: @marandaneto was not able to do this because `config` was not saved as a field so the private method below cannot refer to it.

PR #142, opened 2024-06-11
chore: replay screenshot

marandaneto:
## :pencil: Checklist

### DIFF
--- a/PostHog/Replay/PostHogReplayIntegration.swift
+++ b/PostHog/Replay/PostHogReplayIntegration.swift
@@ -144,7 +144,7 @@
             let style = RRStyle()
 
             // no parent id means its the root
-            if parentId == nil, config.sessionReplayConfig.screenshot {
+            if parentId == nil, config.sessionReplayConfig.screenhshotMode {
                 if let image = view.toImage() {
                     wireframe.base64 = imageToBase64(image)
                 }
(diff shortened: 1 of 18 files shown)


PostHogExample/Api.swift
```diff
@@ -20,18 +20,18 @@ class Api: ObservableObject {
     var logger = Logger(subsystem: Bundle.main.bundleIdentifier!, category: "main")
 
     func listBeers(completion _: @escaping ([PostHogBeerInfo]) -> Void) {
-        guard let url = URL(string: "https://api.punkapi.com/v2/beers") else {
-            return
-        }
-
-        logger.info("Requesting beers list...")
-        URLSession.shared.dataTask(with: url) { _, _, _ in
+//        guard let url = URL(string: "https://api.punkapi.com/v2/beers") else {
```
marandaneto, line 23: this API does not exist anymore, the sample has to be changed

PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -274,6 +286,9 @@
         }
 
         @objc private func snapshot() {
+            // TODO: add debouncer with debouncerDelayMs to take into account how long it takes to execute the
+            // snapshot method
```
marandaneto, line 290: since you cannot offload this from the main thread, a debouncer mechanism would be handy, can be separated to PR to get this out of the door since the timer calls `snapshot` every 500ms already instead of for every frame

PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -96,6 +96,8 @@
                 windowViews.setObject(snapshotStatus, forKey: view)
             }
 
+            // TODO: IncrementalSnapshot, type=2
```
marandaneto, line 99: probably not needed if we make the `screenshot` the default option since it is always a single image anyway, unless at some point we try to optimize the data transfer by breaking a full image in tiles and only sending the ones that have changed.

PostHog/Replay/PostHogSessionReplayConfig.swift
```diff
@@ -23,6 +23,13 @@
         /// Default: true
         @objc public var captureNetworkTelemetry: Bool = true
 
+        /// By default Session replay will capture all the views on the screen as a wireframe,
+        /// By enabling this option, PostHog will capture the screenshot of the screen.
+        /// The screenshot may contain sensitive information, use with caution.
+        /// Experimental support
+        /// Default: false
+        @objc public var screenshot: Bool = false
```
pauldambra, line 31: nit: maybe a nicer name here, `screenhshotMode` ???, despite being experimental we're a little stuck with it once it's in the wild 🤷

pauldambra (approved):
👍 ⚡ 📷

PR #146, opened 2024-06-25
chore: screenshot masking

marandaneto:
Tests running with masking enabled and disabled, with SwiftUI and UIKit

## :pencil: Checklist

### DIFF
--- a/PostHog/Replay/PostHogReplayIntegration.swift
+++ b/PostHog/Replay/PostHogReplayIntegration.swift
@@ -22,6 +22,11 @@
         private let urlInterceptor: URLSessionInterceptor
         private var sessionSwizzler: URLSessionSwizzler?
 
+        // SwiftUI image types
+        private let swiftUIImageTypes = ["_TtCOCV7SwiftUI11DisplayList11ViewUpdater8Platform13CGDrawingView",
+                                         "_TtC7SwiftUIP33_A34643117F00277B93DEBAB70EC0697122_UIShapeHitTestingView",
+                                         "SwiftUI._UIGraphicsView", "SwiftUI.ImageLayer"].compactMap { NSClassFromString($0) }
+
         init(_ config: PostHogConfig) {
             self.config = config
             urlInterceptor = URLSessionInterceptor(self.config)
@@ -170,6 +175,11 @@
                 }
             }
 
+            if swiftUIImageTypes.contains(where: { view.isKind(of: $0) }) {
+                maskableWidgets.append(view.frame)
+                return
+            }
+
             if !view.subviews.isEmpty {
                 for child in view.subviews {
                     if !child.isVisible() {
@@ -230,12 +240,6 @@
         }
 
         private func isImageViewSensitive(_ view: UIImageView) -> Bool {
-            // TODO: redact SwiftUI images:
-            // _TtCOCV7SwiftUI11DisplayList11ViewUpdater8Platform13CGDrawingView
-            // _TtC7SwiftUIP33_A34643117F00277B93DEBAB70EC0697122_UIShapeHitTestingView
-            // SwiftUI._UIGraphicsView
-            // SwiftUI.ImageLayer
-
             var isAsset = false
             if let image = view.image {
                 isAsset = isAssetsImage(image)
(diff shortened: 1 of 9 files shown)


PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -129,33 +134,133 @@
...
+                    context.cgContext.interpolationQuality = .none
+                    image.draw(at: .zero)
+
+                    for rect in maskableWidgets {
+                        let path = UIBezierPath(roundedRect: rect, cornerRadius: 10)
+                        UIColor.black.setFill()
+                        path.fill()
+                    }
+                }
+                wireframe.base64 = imageToBase64(redactedImage)
```
marandaneto, line 217: possible optimization for the future, this can be done async but requires changing a few things and adjusting timestamps, if performance is an issue, this is a place to look at.

PostHog/Replay/UIView+Util.swift
```diff
@@ -44,5 +44,10 @@
 
             return image
         }
+
+        // you need this because of SwiftUI otherwise the coordinates always zeroed for some reason
```
marandaneto, line 48: took me a few hours and a bit of digging to find this, otherwise x, and y are always zeroed for SwiftUI componenets

PostHog/Replay/PostHogReplayIntegration.swift
marandaneto: I've started doing this optimization but led me to do other things that also required to be in the main thread so the changes didn't make much effect, most likely we'll need to pay the memory footprint price instead of processing the images right away, so we can batch the process in background threads but pay the price of holding a copy of the images in memory which can lead to OOM quite quickly as well.

PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -22,6 +22,11 @@
         private let urlInterceptor: URLSessionInterceptor
         private var sessionSwizzler: URLSessionSwizzler?
 
+        // SwiftUI image types
+        private let swiftUIImageTypes = ["_TtCOCV7SwiftUI11DisplayList11ViewUpdater8Platform13CGDrawingView",
+                                         "_TtC7SwiftUIP33_A34643117F00277B93DEBAB70EC0697122_UIShapeHitTestingView",
+                                         "SwiftUI._UIGraphicsView", "SwiftUI.ImageLayer"].compactMap { NSClassFromString($0) }
```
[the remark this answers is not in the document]
daibhin: I suppose we won't know until we release but is there a chance that there are more auto-generated types in a customers app?

PostHog/Replay/UIView+Util.swift
daibhin: "This one trick will save you 4 hours building mobile session replay" - A buzzfeed clickbait article by PostHog

PostHogExample/AppDelegate.swift
```diff
@@ -17,18 +17,22 @@ class AppDelegate: NSObject, UIApplicationDelegate {
...
         config.debug = true
         config.sendFeatureFlagEvent = false
         config.sessionReplay = true
         config.sessionReplayConfig.screenshotMode = true
+        config.sessionReplayConfig.maskAllTextInputs = true
+        config.sessionReplayConfig.maskAllImages = true
 
         PostHogSDK.shared.setup(config)
-        PostHogSDK.shared.debug()
+//        PostHogSDK.shared.debug()
```
daibhin, line 30: Do we want to add these back?

daibhin (approved):
Looks great! Could certainly be a blog post or X thread. It's a cool technical solution that's pretty well contained in a single PR and easy to follow with a relative understanding of Swift

PostHogExample/AppDelegate.swift
marandaneto: Nope because `config.debug = true`, so it was just repeating.

PostHog/Replay/PostHogReplayIntegration.swift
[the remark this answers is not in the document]
marandaneto: Since the types are auto-generated, we cannot just do `view as? SwiftUI.Text` similarly to UIKit.
So yeah I believe this is a moving thing that we'll discover more over time or new ones will come up within the next few SwiftUI versions, but those are the "root" auto-generated types that work well with the basic widgets at least.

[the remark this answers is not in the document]
marandaneto: Some of them are mapped to the UIKit though, for example:
TextField -> `UITextField`.
navigationTitle -> `UILabel`

PR #147, opened 2024-06-26
recording: improvements to screenshot masking

marandaneto:
* only mask if there's any content (text, image)
* added missing components that are placeholders for wireframes but not for screenshots

## :pencil: Checklist

### DIFF
--- a/PostHog/Replay/PostHogReplayIntegration.swift
+++ b/PostHog/Replay/PostHogReplayIntegration.swift
@@ -179,7 +179,7 @@
             }
 
             if let webView = view as? WKWebView {
-                // since we cannot the webview content, if masking texts or images are enabled
+                // since we cannot mask the webview content, if masking texts or images are enabled
                 // we mask the whole webview as well
                 if isAnyInputSensitive(webView) {
                     maskableWidgets.append(view.toAbsoluteRect(parent))
--- a/PostHog/Replay/PostHogReplayIntegration.swift
+++ b/PostHog/Replay/PostHogReplayIntegration.swift
@@ -171,8 +171,38 @@
                 }
             }
 
-            if view is UILabel {
-                if isTextInputSensitive(view) {
+            if let label = view as? UILabel {
+                if isLabelSensitive(label) {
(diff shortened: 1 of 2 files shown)


PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -171,8 +171,38 @@
...
+            }
+
+            if let theSwitch = view as? UISwitch {
+                if isSwitchSensitive(theSwitch) {
+                    maskableWidgets.append(view.toAbsoluteRect(parent))
+                    return
+                }
+            }
+
+            if let picker = view as? UIPickerView {
```
marandaneto, line 204: Added a few missing items that are placeholders for wireframes but not for screenshots and can be sensitive

PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -225,29 +255,63 @@
...
+            }
         }
 
         private func isImageViewSensitive(_ view: UIImageView) -> Bool {
             var isAsset = false
             if let image = view.image {
                 isAsset = isAssetsImage(image)
+            } else {
+                // if there's no image, there's nothing to mask
+                return false
```
marandaneto, line 314: Even if masking is enabled, only mask if there's actually any content (text|image)

daibhin (approved):
Random thought: I don't recall seeing any exit early conditions. I know rrweb will not check child nodes if a parent has been masked which could help some of the extra processing time here

marandaneto:
> Random thought: I don't recall seeing any exit early conditions. I know rrweb will not check child nodes if a parent has been masked which could help some of the extra processing time here

`findMaskableWidgets` (screenshot masking) returns if any condition is met and doesn't check the children's list, so I mask the whole X-Y no matter what.
`toWireframe` indeed does not do it, but the reason is that a component can be composed of different types, eg a Button contains a Label but I have to cast to both types to read the property of each of them, so it's just easier to let it run thru the whole thing, but it's a place to improve if the wireframe has performance issues as well for sure, good point.

PR #150, opened 2024-07-25
fix: respect session replay project settings from app start

marandaneto:
The customer reported it https://posthoghelp.zendesk.com/agent/tickets/16125 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/18624)
The flag was already respected but only after flags were loaded.
Now it does respect with the cached value until the flags are loaded.
It might still send a recording if the cached value is enabled and the session project settings is disabled until it loads but on the next app start or session start it should not happen anymore since the new value will be updated.

The very same patch has to be done on Android

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Next
 
+- recording: fix: respect session replay project settings from app start ([#150](https://github.com/PostHog/posthog-ios/pull/150))
+
 ## 3.6.2 - 2024-07-25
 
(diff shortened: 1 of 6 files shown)


marandaneto:
@PostHog/team-replay is still a draft, but you can double-check if the approach is fine.
have to fix the tests and add new tests.
the web does it differently though because it does not cache any recording data

daibhin (approved):

PR #156, opened 2024-08-08
Update device name

bddq:
Currently the SDK reports a device name that not match the user expectation when reading PostHog documentation.
Using my external package updates the device name to the "real" device name.

I use a fork of your repo with the modification in a production project for months.

## :pencil: Checklist

### DIFF
--- a/PostHog/PostHogContext.swift
+++ b/PostHog/PostHogContext.swift
@@ -6,6 +6,7 @@
 //
 
 import Foundation
+import LocalizedDeviceModel
 
 #if os(iOS) || os(tvOS)
     import UIKit
@@ -44,8 +45,7 @@ class PostHogContext {
 
         #if os(iOS) || os(tvOS)
             let device = UIDevice.current
-            // use https://github.com/devicekit/DeviceKit
-            properties["$device_name"] = device.model
+            properties["$device_name"] = device.productName
             properties["$os_name"] = device.systemName
             properties["$os_version"] = device.systemVersion
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Next
 
+- Update device name to a more explicit name.
+
 ## 3.7.0 - 2024-08-07
 
 - chore: Support the `propertiesSanitizer` config ([#154](https://github.com/PostHog/posthog-ios/pull/154))
--- a/Package.resolved
+++ b/Package.resolved
@@ -2,6 +2,15 @@
   "object": {
     "pins": [
       {
+        "package": "LocalizedDeviceModel",
+        "repositoryURL": "https://github.com/bixcorp/BDLocalizedDevicesModels",
+        "state": {
+          "branch": null,
+          "revision": "d98a5668573829560fd89a77c5b44373ad5db6a6",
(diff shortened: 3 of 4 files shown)


PostHog/PostHogContext.swift
```diff
@@ -6,6 +6,7 @@
 //
 
 import Foundation
+import LocalizedDeviceModel
```
marandaneto, line 9: Can we vendor this extension instead of depending on it directly? this avoids conflicts with dependencies.
We can vendor the class here https://github.com/PostHog/posthog-ios/tree/main/PostHog/Utils and add the license attribution as a comment.

marandaneto:
@bddq 

> Currently the SDK reports a device name that not match the user expectation when reading PostHog documentation.
> Using my external package updates the device name to the "real" device name.

can you link me which prt of the posthog documentation?
can you give me an example of the current value and whats gonna be the new value?

bddq:
> @bddq
> 
> > Currently the SDK reports a device name that not match the user expectation when reading PostHog documentation.
> > Using my external package updates the device name to the "real" device name.
> 
> can you link me which prt of the posthog documentation? can you give me an example of the current value and whats gonna be the new value?

Documentation is the one provided in the tooltip on Activity explorer.
Capture d’écran 2024-08-12 à 09 52 44

Currently, it returns `iPhone10,3` for an iPhone X.

PostHog/PostHogContext.swift
bddq: I would not recommend to vendor the extension because it depends of a localised file to get the "commercial" model name from the device identifier and because each time a new model is released by Apple, you will miss it.

bddq: BTW if you want to embed it, that ok for the attribution.

marandaneto:
> > @bddq
> > > Currently the SDK reports a device name that not match the user expectation when reading PostHog documentation.
> > > Using my external package updates the device name to the "real" device name.
> > 
> > 
> > can you link me which prt of the posthog documentation? can you give me an example of the current value and whats gonna be the new value?
> 
> Documentation is the one provided in the tooltip on Activity explorer. Capture d’écran 2024-08-12 à 09 52 44
> 
> Currently, it returns `iPhone10,3` for an iPhone X.

Alright, makes sense, we have this issue tracking this https://github.com/PostHog/posthog/issues/18025
but I wasnt aware its possible to be done on the SDK side without a huge list similar to https://github.com/tlacroix/mobile-devices-model-mapping-data#data-sources that requires constant updates on the SDK.

PostHog/PostHogContext.swift
marandaneto: That's why we have https://github.com/PostHog/posthog/issues/18025
The goal is to not depend on SDK updates
If this is important for you right now, you can use the props sanitizer https://github.com/PostHog/posthog-ios/blob/b0c6c1847228ce1fb78438be21cfe39cddc55254/PostHog/PostHogConfig.swift#L32C22-L32C41
read the current value, check what's the model name, and update with the human readable model name.

marandaneto:
reason and workaround: https://github.com/PostHog/posthog-ios/pull/156#discussion_r1713318579

Issue #164, opened 2024-08-14
Access sessionId


colin-persona:
### Problem Statement

I want to get and use sessionId for my own purposes (correlating your sessions with my server data). But, it's private. Could you make an accessor, or public?

`private var sessionId: String?`

### Solution Brainstorm

Make an accessor function.

marandaneto:
@colin-persona that is fine, would you like to open a PR and make a read-only getter?

colin-persona:
Yes. Yes I would, thanks! Here it is: https://github.com/PostHog/posthog-ios/pull/165

colin-persona:
I didn't add to change log or bump version... figured you should do that?

marandaneto:
done by https://github.com/PostHog/posthog-ios/pull/170

PR #165, opened 2024-08-14
added getSessionId

colin-persona:

I need your sessionId to help correlate your events with my own server events.

https://github.com/PostHog/posthog-ios/issues/164

In my app.

## :pencil: Checklist

- [ X] I reviewed the submitted code.

NOTE: The code I wrote is this:

```
    @objc public func getSessionId() -> String {
        if !isEnabled() {
            return ""
        }
        
        return sessionLock.withLock {
            return sessionId ?? ""
        }
    }
```
    
 which would have been better as this (return optional, use guard), but I did the above to match your other similar functions...
 
```
     @objc public func getSessionId() -> String? {
        guard isEnabled() { return nil }
        return sessionLock.withLock {
            return sessionId
        }
    }
```

Anyway,  this will help me, thanks!

### DIFF
--- a/PostHog/PostHogSDK.swift
+++ b/PostHog/PostHogSDK.swift
@@ -176,6 +176,16 @@ private let sessionChangeThreshold: TimeInterval = 60 * 30
         return sessionManager?.getAnonymousId() ?? ""
     }
 
+    @objc public func getSessionId() -> String {
+        if !isEnabled() {
+            return ""
+        }
+        
+        return sessionLock.withLock {
+            return sessionId ?? ""
+        }
+    }
+
     // EVENT CAPTURE
 
     private func dynamicContext() -> [String: Any] {


colin-persona:
I didn't bump version or edit changelog - I can if you'd like, but I figured that'd be you folks' department. It'd be nice for this to get deployed with new version so spm would pull it down right away...

colin-persona:
Also, not quite sure which tests you'd want... brand new one? Shoehorn into "identify sets distinct and anon Ids"? How would it be done?

PostHog/PostHogSDK.swift
```diff
@@ -176,6 +176,16 @@ private let sessionChangeThreshold: TimeInterval = 60 * 30
...
 
+    @objc public func getSessionId() -> String {
+        if !isEnabled() {
+            return ""
+        }
+        
+        return sessionLock.withLock {
+            return sessionId ?? ""
+        }
+    }
```
marandaneto, line 187:
```suggestion
    @objc public func getSessionId() -> String? {
        if !isEnabled() {
            return nil
        }
        
        return sessionLock.withLock {
            return sessionId
        }
    }
```

marandaneto:
@colin-persona thanks for the PR.
the `guard` option is a nice idea, I can refactor all methods later.
`getSessionId` can return `nil` since it's not guaranteed that the session feature is enabled.
feel free to add a [changelog](https://github.com/PostHog/posthog-ios/blob/main/CHANGELOG.md) entry, just put it under `Next` and follow the others as an example.
You can add a test on `PostHogSDKTest` since the sessionId is created during SDK init, you just need to guarantee that the `getSessionId` returns non-nil.

marandaneto:
@colin-persona you can run `make format` (commit and push if needed) and later `make lint` to guarantee that CI is happy, and of course `make test` if you add a test.

marandaneto:
closed in favor of https://github.com/PostHog/posthog-ios/pull/170

Issue #151, opened 2024-08-01
Session replay with screenshot mode turned on forces SwiftUI redrawing


NickAtGit:
### Version

3.6.1

### Steps to Reproduce

When using the provided SwiftUI example of the SwiftFortuneWheel package with PostHog integrated and session replay with screenshot mode turned on then the example project behaves strangely. When turning off screenshot mode it works as expected.

https://github.com/sh-khashimov/SwiftFortuneWheel

```swift
        let postHogApiKey = "xxxxxxxxxxxxxx"
        let postHogHost = "https://eu.i.posthog.com"
        let config = PostHogConfig(apiKey: postHogApiKey, host: postHogHost)
        config.sessionReplay = true
        config.sessionReplayConfig.maskAllImages = false
        config.sessionReplayConfig.maskAllTextInputs = false
        config.sessionReplayConfig.screenshotMode = true
        PostHogSDK.shared.setup(config)
```

### Expected Result

The wheel should spin smoothly and come to an end.

### Actual Result
The wheel stops abruptly and the console spams some warnings. I guess some constant SwiftUI redrawing is causing the issue.

https://github.com/user-attachments/assets/4c7da8a4-c15b-4613-922e-4369e8c2952b

NickAtGit:
Thanks! An easy workaround would be to pause the session recording for some views. That would already help a lot...

marandaneto:
@NickAtGit that would be possible per Activity, but for the Single activity apps approach not really.
Maybe a `pause` and `start` option would be more useful, so you know when to start and when to pause better than the SDK.

We are still trying to optimize a few things such as the redimensioning of the screenshot out of the main thread (the screenshot has to be in the main thread though) and converting to base64, also increasing the interval from 500ms to 1s, I believe those would make much better for apps that is constantly redrawing the UI such as transitions and animations.

NickAtGit:
@marandaneto we are speaking about iOS. There are no activities here ;)

marandaneto:
> @marandaneto we are speaking about iOS. There are no activities here ;)

Ops sorry, indeed, but that applies to iOS as well, on SwiftUI for example, the screen names are auto-generated, so that would not work unless you pass a list of types during the SDK init, will take a look.

marandaneto:
@NickAtGit I have made some improvements, would you mind trying out https://github.com/PostHog/posthog-ios/releases/tag/3.7.1 ?

[...]

PR #166, opened 2024-08-16
chore: more performance improvements

marandaneto:
https://github.com/PostHog/posthog-ios/issues/151#issuecomment-2289415555
I am testing directly on this sample app and feels much better

## :pencil: Checklist

### DIFF
--- a/PostHog/Replay/UIView+Util.swift
+++ b/PostHog/Replay/UIView+Util.swift
@@ -45,7 +45,8 @@
             // This can significantly improve rendering performance because the renderer won't need to
             // process transparency.
             rendererFormat.opaque = isOpaque
-            rendererFormat.scale = 0.5
+            // Another way to improve rendering performance is to scale the renderer's content.
+            // rendererFormat.scale = 0.5
             let renderer = UIGraphicsImageRenderer(size: size, format: rendererFormat)
 
             let image = renderer.image { context in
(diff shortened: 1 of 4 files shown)


PostHog/Replay/UIView+Util.swift
```diff
@@ -34,17 +34,25 @@
...
+            let rendererFormat = UIGraphicsImageRendererFormat.default()
 
-            let image = renderer.image { context in
-                // Render the view's layer into the current context
-                layer.render(in: context.cgContext)
+            // This can significantly improve rendering performance because the renderer won't need to
+            // process transparency.
+            rendererFormat.opaque = isOpaque
+            // Another way to improve rendering performance is to scale the renderer's content.
+            // rendererFormat.scale = 0.5
```
marandaneto, line 51: I will only play with that if needed since I have to rescale the masking as well

PostHog/Replay/UIView+Util.swift
```diff
@@ -34,17 +34,25 @@
...
+            let renderer = UIGraphicsImageRenderer(size: size, format: rendererFormat)
+
+            let image = renderer.image { _ in
+                drawHierarchy(in: bounds, afterScreenUpdates: false)
```
marandaneto, line 55: This feels like a big improvement compared to `layer.render(in: context.cgContext)`

daibhin (commented):
🙌

PR #171, opened 2024-08-28
chore: expose session id setter

marandaneto:
follow up https://github.com/PostHog/posthog-ios/pull/170/

Renamed PostHogSessionManager to PostHogStorageManager because it was more about reading/caching things from the storage.
Created a new PostHogSessionManager which is responsible for all session Id related

unit tests already exist and they are green, just refactoring and exposing the session id (setter) via an internal class instead of the public API

## :pencil: Checklist


neilkakkar (approved):
looks like code 🚀

Issue #167, opened 2024-08-21
Support SwiftUI screen names through accessibilityIdentifier and friends


marandaneto:
### Problem Statement

Screen names are auto generated, so autocapture does not work well nor the detection of the current screen for session replay

### Solution Brainstorm

fallback to accessibilityIdentifier or friends if available

https://github.com/PostHog/posthog-ios/blob/e35ec56b8e25d759e285d9a7d831eed070bf3cd3/PostHog/UIViewController.swift#L55
https://github.com/PostHog/posthog-ios/blob/e35ec56b8e25d759e285d9a7d831eed070bf3cd3/PostHog/Replay/PostHogReplayIntegration.swift#L476

marandaneto:
afaik `accessibilityIdentifier` isn't available at runtime for production builds, what we could do is wrap the view:

```
var body: some View {
    PostHogView("Name"){
        List {
            ...
        }
    }
}
```

in this case, we can read the `Name` at runtime but it's not much different than calling `PostHog.screen("Name")`, all screens need to be wrapped.

marandaneto:
Another option would be to use the `objc_getAssociatedObject` and `objc_setAssociatedObject` as an extension method to UIKit and SwiftUI views

PR #180, opened 2024-09-05
chore: add SwiftUI View extensions to capture screen view and views in general (postHogViewSeen, postHogScreenView)

marandaneto:
Spike for "semi automatic" or more UX friendly https://github.com/PostHog/posthog-ios/issues/167

Screen view (applied to the root view)

```swift
var body: some Scene {
        WindowGroup {
            ContentView()
                .postHogScreenView() // will infer the class name (ContentView)
        }
    }
```

Widget view (applied to any view)

```swift
Button(action: triggerAction) {
    Text("My Text!")
}.postHogViewSeen("My Button")
```

Events will be captured when the View triggers the [onAppear](https://developer.apple.com/documentation/swiftui/view/onappear(perform:)) method

## :pencil: Checklist

### DIFF
--- a/PostHog/PostHogSwiftUIViewModifiers.swift
+++ b/PostHog/PostHogSwiftUIViewModifiers.swift
@@ -37,8 +37,8 @@
                                                        properties: properties))
         }
 
-        func postHogViewEvent(_ event: String,
-                              _ properties: [String: Any]? = nil) -> some View
+        func postHogViewSeen(_ event: String,
+                             _ properties: [String: Any]? = nil) -> some View
         {
             modifier(PostHogSwiftUIViewModifier(viewEventName: event,
                                                 screenEvent: false,
--- a/PostHog/PostHogSwiftUIViewModifiers.swift
+++ b/PostHog/PostHogSwiftUIViewModifiers.swift
@@ -1,0 +1,48 @@
+//
+//  PostHogSwiftUIViewModifiers.swift
+//  PostHog
+//
+//  Created by Manoel Aranda Neto on 05.09.24.
+//
+
+#if canImport(SwiftUI)
+    import Foundation
+    import SwiftUI
+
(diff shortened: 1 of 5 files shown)


PostHog/PostHogSwiftUIViewModifiers.swift
```diff
@@ -0,0 +1,48 @@
...
+    public extension View {
+        func postHogScreenView(_ screenName: String,
+                               _ properties: [String: Any]? = nil) -> some View
+        {
+            modifier(PostHogSwiftUIViewModifier(viewEventName: screenName,
+                                                screenEvent: true,
+                                                properties: properties))
+        }
+
+        func postHogViewEvent(_ event: String,
```
marandaneto, line 39: This is for when the widget/view is visible, but not clicked or interacted with, hence the name "viewEvent", any better ideas?

PostHog/PostHogSwiftUIViewModifiers.swift
```diff
@@ -0,0 +1,49 @@
...
+            let viewEventName = screenName ?? "\(type(of: self))"
+            return modifier(PostHogSwiftUIViewModifier(viewEventName: viewEventName,
+                                                       screenEvent: true,
+                                                       properties: properties))
+        }
+
+        func postHogViewEvent(_ event: String,
+                              _ properties: [String: Any]? = nil) -> some View
+        {
+            modifier(PostHogSwiftUIViewModifier(viewEventName: event,
```
marandaneto, line 43: I cannot use `type(of: self)` here because it will be `Button` for all buttons

PostHog/PostHogSwiftUIViewModifiers.swift
daibhin: Hmm it wasn't immediately obvious to me from reading the PR description what it did either.

Maybe something along the lines of rendered, seen, displayed?

PR #183, opened 2024-09-06
recording: send meta event again if session rotates

marandaneto:
The problem: If meta event is missing, the replay will be just a black screen because rrweb hides everything if no meta event is provided

We cache the window and the status object in a weak reference.
If the session rotates, and the window and status object are still in memory, we won't be sending the meta event again
Every time the session rotates, we clear the weak reference object so meta events will be sent again

Running

## :pencil: Checklist


pauldambra (approved):
nice 🚀

PR #189, opened 2024-09-11
chore: add urlsession extension for async methods

marandaneto:
https://posthog.slack.com/archives/C05U2CNV6DV/p1725979956451989

Swift methods that are not exposed in the objC runtime cannot be swizzled.
So if people are using those methods for network calls, they don't get network logs.
This exposes an extension that makes their life easier.

Running the sample with such methods

## :pencil: Checklist

### DIFF
--- a/PostHog/Replay/URLSessionExtension.swift
+++ b/PostHog/Replay/URLSessionExtension.swift
@@ -1,184 +1,127 @@
 #if os(iOS)
     public extension URLSession {
-        func postHogData(for request: URLRequest) async throws -> (Data, URLResponse) {
-            let before = Date()
-            var after: Date?
+        private func getMonotonicTimeInMilliseconds() -> UInt64 {
+            // Get the raw mach time
+            let machTime = mach_absolute_time()
+
+            // Get timebase info to convert to nanoseconds
+            var timebaseInfo = mach_timebase_info_data_t()
+            mach_timebase_info(&timebaseInfo)
+
+            // Convert mach time to nanoseconds
+            let nanoTime = machTime * UInt64(timebaseInfo.numer) / UInt64(timebaseInfo.denom)
+
+            // Convert nanoseconds to milliseconds
+            let milliTime = nanoTime / 1_000_000
+
+            return milliTime
+        }
+
+        private func executeRequest(request: URLRequest? = nil,
+                                    action: () async throws -> (Data, URLResponse)) async throws -> (Data, URLResponse)
+        {
+            let timestamp = Date()
+            let startMillis = getMonotonicTimeInMilliseconds()
+            var endMillis: UInt64?
             do {
-                let (data, response) = try await self.data(for: request)
-                after = Date()
-                captureData(request: request, response: response, before: before, after: after)
+                let (data, response) = try await action()
+                endMillis = getMonotonicTimeInMilliseconds()
+                captureData(request: request, response: response, timestamp: timestamp, start: startMillis, end: endMillis)
                 return (data, response)
             } catch {
-                captureData(request: request, response: nil, before: before, after: after)
+                captureData(request: request, response: nil, timestamp: timestamp, start: startMillis, end: endMillis)
                 throw error
             }
+        }
+
+        private func executeRequest(request: URLRequest? = nil,
+                                    action: () async throws -> (URL, URLResponse)) async throws -> (URL, URLResponse)
+        {
+            let timestamp = Date()
+            let startMillis = getMonotonicTimeInMilliseconds()
+            var endMillis: UInt64?
+            do {
+                let (url, response) = try await action()
+                endMillis = getMonotonicTimeInMilliseconds()
+                captureData(request: request, response: response, timestamp: timestamp, start: startMillis, end: endMillis)
+                return (url, response)
+            } catch {
+                captureData(request: request, response: nil, timestamp: timestamp, start: startMillis, end: endMillis)
+                throw error
+            }
+        }
+
+        func postHogData(for request: URLRequest) async throws -> (Data, URLResponse) {
+            try await executeRequest(request: request, action: { try await data(for: request) })
         }
 
         func postHogData(from url: URL) async throws -> (Data, URLResponse) {
-            let before = Date()
-            do {
-                let (data, response) = try await self.data(from: url)
-                let after = Date()
-                captureData(request: nil, response: response, before: before, after: after)
-                return (data, response)
-            } catch {
-                throw error
-            }
+            try await executeRequest(action: { try await data(from: url) })
         }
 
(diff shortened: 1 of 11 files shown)


cameronehrlich:
@marandaneto could you also provide `postHog*` methods for the other async methods on URLSession?  Like this:
```swift
extension URLSession {

    public func postHogData(for request: URLRequest, delegate: (any URLSessionTaskDelegate)? = nil) async throws -> (Data, URLResponse) {
        let before = Date()
        let (data, response) = try await data(for: request, delegate: delegate)
        let after = Date()
        captureData(request: request, response: response, before: before, after: after)
        return (data, response)
    }

    public func postHogData(from url: URL, delegate: (any URLSessionTaskDelegate)? = nil) async throws -> (Data, URLResponse) {
        let before = Date()
        let (data, response) = try await data(from: url, delegate: delegate)
        let after = Date()
        captureData(request: nil, response: response, before: before, after: after)
        return (data, response)
    }

    public func postHogUpload(for request: URLRequest, fromFile fileURL: URL, delegate: (any URLSessionTaskDelegate)? = nil) async throws -> (Data, URLResponse) {
        let before = Date()
        let (data, response) = try await upload(for: request, fromFile: fileURL, delegate: delegate)
        let after = Date()
        captureData(request: request, response: response, before: before, after: after)
        return (data, response)
    }

    public func postHogUpload(for request: URLRequest, from bodyData: Data, delegate: (any URLSessionTaskDelegate)? = nil) async throws -> (Data, URLResponse) {
        let before = Date()
        let (data, response) = try await upload(for: request, from: bodyData, delegate: delegate)
        let after = Date()
        captureData(request: request, response: response, before: before, after: after)
        return (data, response)
    }

    public func postHogDownload(for request: URLRequest, delegate: (any URLSessionTaskDelegate)? = nil) async throws -> (URL, URLResponse) {
        let before = Date()
        let (fileURL, response) = try await download(for: request, delegate: delegate)
        let after = Date()
        captureData(request: request, response: response, before: before, after: after)
        return (fileURL, response)
    }

    public func postHogDownload(from url: URL, delegate: (any URLSessionTaskDelegate)? = nil) async throws -> (URL, URLResponse) {
        let before = Date()
        let (fileURL, response) = try await download(from: url, delegate: delegate)
        let after = Date()
        captureData(request: nil, response: response, before: before, after: after)
        return (fileURL, response)
    }

    public func postHogDownload(resumeFrom resumeData: Data, delegate: (any URLSessionTaskDelegate)? = nil) async throws -> (URL, URLResponse) {
        let before = Date()
        let (fileURL, response) = try await download(resumeFrom: resumeData, delegate: delegate)
        let after = Date()
        captureData(request: nil, response: response, before: before, after: after)
        return (fileURL, response)
    }

    // MARK: Private methods
    private func captureData(request: URLRequest?, response: URLResponse, before: Date, after: Date) {
        // TODO: Capture the request data
    }
}
```

PostHog/Replay/URLSessionExtension.swift
```diff
@@ -0,0 +1,195 @@
...
+            return (data, response)
+        } catch {
+            captureData(request: request, response: nil, before: before, after: after)
+            throw error
+        }
+    }
+
+    @available(macOS 12.0, iOS 15.0, watchOS 8.0, tvOS 15.0, *)
+    func postHogUpload(for request: URLRequest, from bodyData: Data, delegate: (any URLSessionTaskDelegate)? = nil) async throws -> (Data, URLResponse) {
+        let before = Date()
```
pauldambra, line 103: i assume this is the best clock to use? JS has "performance" clock that is (more) guaranteed to monotonically increase etc

PostHog/Replay/URLSessionExtension.swift
```diff
@@ -0,0 +1,195 @@
...
+            return (data, response)
+        } catch {
+            captureData(request: nil, response: nil, before: before, after: after)
+            throw error
+        }
+    }
+
+    @available(macOS 12.0, iOS 15.0, watchOS 8.0, tvOS 15.0, *)
+    func postHogUpload(for request: URLRequest, fromFile fileURL: URL, delegate: (any URLSessionTaskDelegate)? = nil) async throws -> (Data, URLResponse) {
+        let before = Date()
```
pauldambra: these bodies all seem relatively identical... would they all change at the same time - is it worth extracting "handlers" so it's easier to see what's the same and what's different?

pauldambra (approved):
👍 my extensive Swift experience being brought to bear 🤣

PostHog/Replay/URLSessionExtension.swift
marandaneto: yes and no, https://developer.apple.com/documentation/swift/continuousclock
only from iOS >= 16 and we support iOS >= 13

marandaneto: Let me check if there's anything available in C

PostHog/Replay/URLSessionExtension.swift
marandaneto: The signature has to be identical to the original ones.
I extracted everything to `captureData` except what has to be read as early as possible such as the `before` and `after`.
Since the `captureData` spawns a background thread to capture the network data, there's a bit of duplication, indeed.
What do you mean by `handlers`?

pauldambra: so... the content of each method appears to be almost identical

so I'd almost expect (Assuming you can pass functions around in swift)

```

onNetwork(networkFn) {
        let before = Date()
        var after: Date?
        do {
            let (a, b) = networkFn()
            after = Date()
            captureData(request: request, response: response, before: before, after: after)
            return (a, b)
        } catch {
            captureData(request: request, response: nil, before: before, after: after)
            throw error
        }

```

so then you'd have

```
@available(macOS 12.0, iOS 15.0, watchOS 8.0, tvOS 15.0, *)
    func postHogDownload(resumeFrom resumeData: Data, delegate: (any URLSessionTaskDelegate)? = nil) async throws -> (URL, URLResponse) {
        onNetwork(() => try await download(resumeFrom: resumeData, delegate: delegate))
    }
```

since they all seem to be methods that wrap a function call with slightly different args and a destructed response 

so when I'm reading the file I can see that all of the swizzled functions do the same thing apart from how they make the network call

but that's coming from a place of basically no swift knowledge so feel free to ignore 😊

marandaneto: ah gotcha what you mean

PostHog/Replay/URLSessionExtension.swift
marandaneto: used something from the kernel, hopefully it works on all platforms https://developer.apple.com/documentation/kernel/1462446-mach_absolute_time

PostHog/Replay/URLSessionExtension.swift
marandaneto: done, good idea.

PR #191, opened 2024-09-11
chore: recordings sdk info

marandaneto:
Relates to https://github.com/PostHog/posthog-android/pull/172

_#skip-changelog_ since its not a user facing change

Capturing both type of events

## :pencil: Checklist

### DIFF
--- a/PostHog/PostHogSDK.swift
+++ b/PostHog/PostHogSDK.swift
@@ -1,4 +1,4 @@
-// swiftlint:disable file_length
+// swiftlint:disable file_length cyclomatic_complexity
 
 //
 //  PostHogSDK.swift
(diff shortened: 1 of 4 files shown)


PostHog/PostHogSDK.swift
```diff
@@ -301,18 +301,23 @@ let maxRetryDelay = 30.0
...
+        }
+
         if let sessionId = PostHogSessionManager.shared.getSessionId() {
             props["$session_id"] = sessionId
-            // Session replay requires $window_id, so we set as the same as $session_id.
+            // only Session replay requires $window_id, so we set as the same as $session_id.
             // the backend might fallback to $session_id if $window_id is not present next.
             #if os(iOS)
-                if config.sessionReplay {
+                if !appendSharedProps, config.sessionReplay {
```
marandaneto, line 314: @pauldambra just to confirm
events only need `$session_id` and not `$window_id`
recordings need both, right?

pauldambra: we put $window_id on all events from JS.

I can't believe that we put $window_id on all events from e.g. the Go SDK so it _should_ be safe to not add $window_id where it's not meaningful

I think either in capture or at query time we'd be safe to say $window_id == $session_id if not present

pauldambra (approved):
👍

PR #201, opened 2024-09-23
chore: make storage manager public to the rn sdk

marandaneto:
required for https://github.com/PostHog/posthog-js-lite/pull/269
_#skip-changelog_

## :pencil: Checklist

### DIFF
--- a/PostHog/PostHogConfig.swift
+++ b/PostHog/PostHogConfig.swift
@@ -53,6 +53,8 @@ import Foundation
     // only internal
     var disableReachabilityForTesting: Bool = false
     var disableQueueTimerForTesting: Bool = false
+    // internal
+    public var storageManager: PostHogStorageManager?
 
     @objc(apiKey:)
     public init(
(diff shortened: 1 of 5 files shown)


PostHog/PostHogConfig.swift
```diff
@@ -53,6 +53,8 @@ import Foundation
     // only internal
     var disableReachabilityForTesting: Bool = false
     var disableQueueTimerForTesting: Bool = false
+    // internal
+    public var storageManager: PostHogStorageManager?
```
marandaneto, line 57: I will refactor the access to such global properties later on since this pollutes the public API and people can misuse that even with "internal" comments, but for now, I'd like to quickly fix the bug and that requires access to the storage manager.

daibhin: Maybe worth adding a TODO comment to this effect

marandaneto: will make an issue instead.

Issue #162, opened 2024-08-14
Support `ph-no-capture` for SwiftUI


marandaneto:
### Description

See note https://posthog.com/docs/session-replay/ios

`.accessibilityIdentifier("ph-no-capture")`

when reading the `accessibilityIdentifier` property, it always returns nil

marandaneto:
See https://github.com/PostHog/posthog-ios/issues/167#issuecomment-2310132354
Probably the same technique is possible

marandaneto:
Found a workaround for manually tagging views for SwiftUI https://github.com/PostHog/posthog-ios/pull/202

PR #202, opened 2024-09-23
chore: manually mask swiftui views

marandaneto:
Alternative solution for https://github.com/PostHog/posthog-ios/issues/162
You inject a synthetic view into the marked view and get the X and Y positions based on that component.
You don't have access to the original view but you know where it should be masked.

## :pencil: Checklist


marandaneto:
@ioannisj you can pick this up if you wish, the PoC works, just need to polish and finish it

CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 ## Next
+- add `postHogMask` view modifier to manually mask a SwiftUI view ([#202](https://github.com/PostHog/posthog-ios/pull/202))
```
marandaneto, line 2:
```suggestion

- add `postHogMask` view modifier to manually mask a SwiftUI view ([#202](https://github.com/PostHog/posthog-ios/pull/202))
```

PostHog/PostHogMaskViewModifier.swift
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    private struct PostHogMaskViewTagger: UIViewRepresentable {
+        func makeUIView(context _: Context) -> PostHogMaskViewTaggerView {
+            PostHogMaskViewTaggerView()
+        }
+
+        func updateUIView(_: PostHogMaskViewTaggerView, context _: Context) {
+            // nothing
+        }
```
marandaneto, line 25: can we remove this method or is it mandatory?

PostHog/PostHogMaskViewModifier.swift
```diff
@@ -0,0 +1,68 @@
...
+        @ViewBuilder
+        private var viewTagger: some View {
+            if enabled {
+                PostHogMaskViewTagger()
+            }
+        }
+    }
+
+    private class PostHogMaskViewTaggerView: UIView {
+        override func didMoveToSuperview() {
```
marandaneto, line 45: is there a method to override that we would set `phIsManuallyMasked` to false? so maybe fixes the issues with the open sheet

PostHog/Replay/PostHogScreenshotMasker.swift
```diff
@@ -0,0 +1,39 @@
...
+//  PostHog
+//
+//  Created by Manoel Aranda Neto on 23.09.24.
+//
+
+#if os(iOS) && canImport(SwiftUI)
+    import Foundation
+    import SwiftUI
+
+    class PostHogScreenshotMasker {
```
marandaneto, line 12: should this file be removed then?

PostHogExample/ContentView.swift
```diff
@@ -113,7 +113,7 @@ struct ContentView: View {
                         RepresentedExampleUIView()
                     }
 
-                    Text("Sensitive text!!").accessibilityIdentifier("ph-no-capture")
+                    Text("Sensitive text!!").postHogMask()
```
marandaneto, line 116: I think we can clean all the accessibilityIdentifier in the swiftui samples since it will not work anyway, maybe replace all with posthogMask

PostHogExample/AppDelegate.swift
```diff
@@ -23,8 +23,8 @@ class AppDelegate: NSObject, UIApplicationDelegate {
         config.sendFeatureFlagEvent = false
         config.sessionReplay = true
         config.sessionReplayConfig.screenshotMode = true
-        config.sessionReplayConfig.maskAllTextInputs = true
-        config.sessionReplayConfig.maskAllImages = true
+        config.sessionReplayConfig.maskAllTextInputs = false
+        config.sessionReplayConfig.maskAllImages = false
```
marandaneto, line 27: we can rollback this once its ready to LGTM

PostHog/PostHogMaskViewModifier.swift
```diff
@@ -0,0 +1,67 @@
...
+//
+//  Created by Yiannis Josephides on 09/10/2024.
+//
+
+#if os(iOS) && canImport(SwiftUI)
+
+    import SwiftUI
+
+    public extension View {
+        func postHogMask(_ value: Bool = true) -> some View {
```
marandaneto, line 13: Do we need the `value` if there's no way to force it not to be masked? probably better to leave without it

marandaneto: makes sense, not sure `isActive` is clear but its already better than `value` that has no meaning at all in this context, maybe `maskIt: Bool = true` is more clearer, we can inspire from any other swift docs btw
https://developer.apple.com/documentation/swiftui/view/mask(_:) they use `mask` only but its an object

marandaneto:
@ioannisj I cannot approve my own PR so feel free to get it approved/merged after fixing the conflicts.

PR #203, opened 2024-10-01
recording: capture network logs from dataTask requests without CompletionHandler

marandaneto:
Closes https://github.com/PostHog/posthog-js-lite/issues/276

When dataTask is called without a completion handler, `taskCompleted` isn't called, so the network log was never captured.
Since we need the data after the task is finished and we don't know when the task is finished, on every request, I call `finishAll` and send all pending requests that were finished.
This means that if a request is made without the completion handler, and the app is finished or no other request is made, this request is never captured.
We could add a timer that sends all the pending network logs, so this won't happen unless the app is closed right away before the timer triggers.
Instead of doing this hacky solution, I'd rather release this as a quick fix and work on the whole instrumentation bits, my goal is to borrow the logic from https://github.com/kasketis/netfox that instruments lower-level stuff, and with that we would have a better way to handle those use cases, this would be a new issue and a new PR later on.

Executing URLSession.shared.dataTask(with: URL) requests

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Next
 
+- recording: capture network logs from dataTask requests without CompletionHandler ([#196](https://github.com/PostHog/posthog-ios/pull/196))
+
 ## 3.12.5 - 2024-09-24
 
 - no user facing changes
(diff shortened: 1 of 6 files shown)


pauldambra (approved):
🚀

PR #207, opened 2024-10-08
feat: add appGroupIdentifier in posthog config

ioannisj:

Adds support for application group identifier so that host app and app extensions (widgets, notifications, app clips etc) can share analytics data

see: https://posthog.slack.com/archives/C03PB072FMJ/p1727978109888519

Tested with sample app projects 

- iOS: App that adds a widget with a single action to a demo app. Made sure that `getAnonymousId()` returns the same cached identifier for both the Widget app and Host app

- macOS: Seems that `FileManager.default.containerURL(forSecurityApplicationGroupIdentifier` returns a *writable* URL regardless of the App Group Identifier provided

- iOS: Seems that `FileManager.default.containerURL(forSecurityApplicationGroupIdentifier` indeed returns nil if provided App Group Identifier does not match the one in the entitlements

Apple docs: https://developer.apple.com/documentation/foundation/filemanager/1412643-containerurl/

## :pencil: Checklist

### DIFF
--- a/PostHog/PostHogStorage.swift
+++ b/PostHog/PostHogStorage.swift
@@ -122,7 +122,7 @@ class PostHogStorage {
     /**
      There are cases where applications using posthog-ios want to share analytics data between host app and an app extension, Widget or App Clip. If there's a defined `appGroupIdentifier` in configuration, we want to use a shared container for storing data so that extensions correcly identify a user (and batch process events)
      */
-    private static func getAppFolderUrl(from configuration: PostHogConfig? = nil) -> URL {
+    private static func getAppFolderUrl(from configuration: PostHogConfig) -> URL {
         /**
 
          From Apple Docs:
@@ -134,7 +134,7 @@ class PostHogStorage {
           see: https://developer.apple.com/documentation/foundation/filemanager/1412643-containerurl/
           */
         func appGroupContainerUrl() -> URL? {
-            guard let appGroupIdentifier = configuration?.appGroupIdentifier else { return nil }
+            guard let appGroupIdentifier = configuration.appGroupIdentifier else { return nil }
 
             let url = FileManager.default
                 .containerURL(forSecurityApplicationGroupIdentifier: appGroupIdentifier)?


PostHog/PostHogConfig.swift
```diff
@@ -33,6 +33,11 @@ import Foundation
     /// Determines the behavior for processing user profiles.
     @objc public var personProfiles: PostHogPersonProfiles = .identifiedOnly
 
+    /// The identifier of the App Group that should be used to store shared analytics data.
+    /// Posthog will try to get the physical location of the App Group’s shared container, otherwise fallback to the default location
```
marandaneto, line 37:
```suggestion
    /// PostHog will try to get the physical location of the App Group’s shared container, otherwise fallback to the default location
```

PostHog/PostHogStorage.swift
```diff
@@ -12,11 +12,21 @@ import Foundation
 
  posthog-ios stores data either to file or to UserDefaults in order to support tvOS. As recordings won't work on tvOS anyways and we have no tvOS users so far,
  we are opting to only support iOS via File storage.
+
+ There are cases where applications using posthog-ios want to share analytics data between host app and an app extension, Widget or App Clip. If there's a defined `appGroupIdentifier` in configuration, we want to use a shared container for storing data so that extensions correcly identify a user (and batch process events)
  */
+func getAppFolderUrl(from configuration: PostHogConfig) -> URL {
```
marandaneto, line 18: can this be private instead?

PostHog/PostHogStorage.swift
```diff
@@ -45,7 +55,7 @@ class PostHogStorage {
     init(_ config: PostHogConfig) {
         self.config = config
 
-        appFolderUrl = applicationSupportDirectoryURL() // .appendingPathComponent(config.apiKey)
+        appFolderUrl = getAppFolderUrl(from: config)
```
marandaneto, line 58: I think we could add a unit test for the final `appFolderUrl` if the `appGroupIdentifier` is set or not

PostHog/PostHogStorage.swift
```diff
@@ -129,6 +119,39 @@ class PostHogStorage {
         setData(forKey: key, contents: data)
     }
 
+    /**
+     There are cases where applications using posthog-ios want to share analytics data between host app and an app extension, Widget or App Clip. If there's a defined `appGroupIdentifier` in configuration, we want to use a shared container for storing data so that extensions correcly identify a user (and batch process events)
+     */
+    private static func getAppFolderUrl(from configuration: PostHogConfig? = nil) -> URL {
```
marandaneto, line 125: I think config does not need to be null here since we always call with non null values

marandaneto (approved):
left a comment otherwise LGTM

PR #209, opened 2024-10-10
fix: session replay respect feature flag variants

marandaneto:
JS PR https://github.com/PostHog/posthog-js/pull/1040/

## :pencil: Checklist

### DIFF
--- a/PostHog/PostHogFeatureFlags.swift
+++ b/PostHog/PostHogFeatureFlags.swift
@@ -53,13 +53,19 @@ class PostHogFeatureFlags {
         var recordingActive = false
 
         if let linkedFlag = sessionRecording["linkedFlag"] as? String,
-           let active = featureFlags[linkedFlag] as? Bool
+           let value = featureFlags[linkedFlag] as? Bool
         {
-            recordingActive = active
+            recordingActive = value
         } else if let linkedFlag = sessionRecording["linkedFlag"] as? String,
                   featureFlags[linkedFlag] != nil
         {
             recordingActive = true
+        } else if let linkedFlag = sessionRecording["linkedFlag"] as? [String: Any],
+                  let flag = linkedFlag["flag"] as? String,
+                  let variant = linkedFlag["variant"] as? String,
+                  let value = featureFlags[flag] as? String
+        {
+            recordingActive = value == variant
         }
 
         return recordingActive
--- a/PostHog/PostHogFeatureFlags.swift
+++ b/PostHog/PostHogFeatureFlags.swift
@@ -50,21 +50,25 @@ class PostHogFeatureFlags {
     }
 
     private func isRecordingActive(_ featureFlags: [String: Any], _ sessionRecording: [String: Any]) -> Bool {
-        var recordingActive = false
+        var recordingActive = true
 
+        // check for boolean flags
         if let linkedFlag = sessionRecording["linkedFlag"] as? String,
            let value = featureFlags[linkedFlag] as? Bool
         {
             recordingActive = value
+            // check for specific flag variant
         } else if let linkedFlag = sessionRecording["linkedFlag"] as? [String: Any],
                   let flag = linkedFlag["flag"] as? String,
                   let variant = linkedFlag["variant"] as? String,
                   let value = featureFlags[flag] as? String
         {
             recordingActive = value == variant
+            // check for multi flag variant (any)
         } else if let linkedFlag = sessionRecording["linkedFlag"] as? String,
                   featureFlags[linkedFlag] != nil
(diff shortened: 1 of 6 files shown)


PostHog/PostHogFeatureFlags.swift
```diff
@@ -49,6 +49,32 @@ class PostHogFeatureFlags {
...
+                  let value = featureFlags[flag] as? String
+        {
+            recordingActive = value == variant
+            // check for multi flag variant (any)
+        } else if let linkedFlag = sessionRecording["linkedFlag"] as? String,
+                  featureFlags[linkedFlag] != nil
+        {
+            // always true if the flag exists
+            recordingActive = true
+        }
```
marandaneto, line 73: I feel like this is not necessary since I have `linkedFlag` but nothing to compare so it means just active?

marandaneto:
@pauldambra would you mind taking a look? since I will need to implement this across SDKs

PostHog/PostHogFeatureFlags.swift
```diff
@@ -49,6 +49,32 @@ class PostHogFeatureFlags {
         }
     }
 
+    private func isRecordingActive(_ featureFlags: [String: Any], _ sessionRecording: [String: Any]) -> Bool {
+        var recordingActive = true
```
marandaneto, line 53: default is active since the `sessionRecording` is a Map and not a boolean unless I find the `linkedFlag` and can compare things

PostHog/Replay/URLSessionInterceptor.swift
```diff
@@ -141,10 +141,8 @@
         private func finishAll() {
             var completedTasks: [URLSessionTask: NetworkSample] = [:]
             tasksLock.withLock {
-                for item in samplesByTask {
-                    if item.key.state == .completed {
-                        completedTasks[item.key] = item.value
-                    }
+                for item in samplesByTask where item.key.state == .completed {
+                    completedTasks[item.key] = item.value
```
marandaneto, line 145: just fixing a lint issue

PostHog/PostHogFeatureFlags.swift
pauldambra: i thought the feature flag response was always `Record<string, string|bool>` so I don't know if the third branch is correct

i.e. I don't know that we can say that the recording is active if the linked flag is present without inspecting its value

but given we've covered 

* it's a bool in the first branch
* and it's a string in the second

I think you don't need the third branch

* there is no linked flag - nothing to do
* there's a linked flag but it is some format we weren't expecting - nothing to do

since we default to active I agree I think the 3rd branch is unnecessary

otherwise the logic looks right to me

marandaneto: The thing is, the 3rd branch is a valid response from the API.
You create a multi-variant flag, assign it to the recording config, and select `any`, the `linkedFlag` is a String, but you cannot check it since it is "any", it'd be a match for everything anyway, so I'd assume removing the branch is just ok since it will be active as well

PostHog/PostHogStorage.swift
```diff
@@ -120,15 +120,20 @@ class PostHogStorage {
     }
 
     /**
-     There are cases where applications using posthog-ios want to share analytics data between host app and an app extension, Widget or App Clip. If there's a defined `appGroupIdentifier` in configuration, we want to use a shared container for storing data so that extensions correcly identify a user (and batch process events)
+     There are cases where applications using posthog-ios want to share analytics data between host app and
```
marandaneto, line 123: just fixed a lint issue (line_length)

PostHog/PostHogFeatureFlags.swift
marandaneto: pushed changes and added tests

marandaneto:
@ioannisj @pauldambra done, please double check after changes, I will do Android next (Monday)

PostHog/PostHogFeatureFlags.swift
```diff
@@ -36,19 +36,46 @@ class PostHogFeatureFlags {
...
+            recordingActive = value
+            // check for specific flag variant
+        } else if let linkedFlag = sessionRecording["linkedFlag"] as? [String: Any],
+                  let flag = linkedFlag["flag"] as? String,
+                  let variant = linkedFlag["variant"] as? String,
+                  let value = featureFlags[flag] as? String
+        {
+            recordingActive = value == variant
+        }
+        // check for multi flag variant (any)
```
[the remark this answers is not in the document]
marandaneto: I don't think so since it's a valid result from the API and might help while debugging things

PR #216, opened 2024-10-16
feat(iOS): add optional distinctId parameter to capture methods

ioannisj:

We had a small number of requests from users to be able to capture an event using a custom (anonymous) id for certain events. 

Since this is already supported in the [Android SDK](https://github.com/PostHog/posthog-android/blob/85f6ef1300919ac1ee5d9bf615c87e0fe79fb247/posthog/src/main/java/com/posthog/PostHog.kt#L391) we are adding support for iOS SDK as well

Unit tests

## :pencil: Checklist

### DIFF
--- a/PostHogObjCExample/AppDelegate.m
+++ b/PostHogObjCExample/AppDelegate.m
@@ -28,9 +28,31 @@
             object:nil];
 
     PostHogConfig *config = [[PostHogConfig alloc] apiKey:@"_6SG-F7I1vCuZ-HdJL3VZQqjBlaSb1_20hDPwqMNnGI"];
-    config.preloadFeatureFlags = NO;
+    config.preloadFeatureFlags = YES;
     [[PostHogSDK shared] debug:YES];
     [[PostHogSDK shared] setup:config];
+    
+    NSString *event = @"theEvent";
+    NSString *distinctId = @"theCustomDistinctId";
+    NSDictionary *properties = @{@"source": @"iOS App", @"state": @"running"};
+    NSDictionary *userProperties = @{@"userAlive": @YES, @"userAge": @50};
+    NSDictionary *userPropertiesSetOnce = @{@"signupDate": @"2024-10-16"};
+    NSDictionary *groups = @{@"groupName": @"developers"};
+
+    [[PostHogSDK shared] captureWithEvent:event
+                               distinctId:distinctId
+                               properties:properties
+                            userProperties:userProperties
+                    userPropertiesSetOnce:userPropertiesSetOnce
+                                   groups:groups
+    ];
+    
+    [[PostHogSDK shared] captureWithEvent:event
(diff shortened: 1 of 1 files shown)


PostHogExampleWithSPM/PostHogExampleWithSPM.xcodeproj/project.pbxproj
```diff
@@ -117,7 +120,7 @@
 			);
 			mainGroup = 690FF0392AE7DB3600A0B06B;
 			packageReferences = (
-				690FF05B2AE7E03B00A0B06B /* XCLocalSwiftPackageReference ".." */,
+				DA8D37472CBF99D1005EBD27 /* XCLocalSwiftPackageReference "../../posthog-ios" */,
```
marandaneto, line 123: Is this intentional?

PostHog/PostHogSDK.swift
```diff
@@ -510,8 +520,9 @@ let maxRetryDelay = 30.0
         return false
     }
 
-    @objc(captureWithEvent:properties:userProperties:userPropertiesSetOnce:groups:)
+    @objc(captureWithEvent:distinctId:properties:userProperties:userPropertiesSetOnce:groups:)
     public func capture(_ event: String,
+                        distinctId distinctIdParam: String? = nil,
```
marandaneto, line 525: Can we just name `distinctId: String? = nil,` instead? why do we need `distinctIdParam`?

PostHog/PostHogSDK.swift
```diff
@@ -478,28 +478,38 @@ let maxRetryDelay = 30.0
         capture(event, properties: nil, userProperties: nil, userPropertiesSetOnce: nil, groups: nil)
     }
 
-    @objc(captureWithEvent:properties:)
+    @objc(captureWithEvent:distinctId:)
```
marandaneto, line 481: Can you check if this would play well with all the PostHogObjCExample?
The goal would be to not force people passing distinctId=nil in ObjC, ObjC does not handle well the optional, and they should not be forced to pass distinctId either.

marandaneto: since this is a corner case, I'd be ok offering just one overload with all options at once and `distinctId`, for Swift this is not an issue, and for objC might be a bit verbose but this type of captured event is an edge case anyway.
So i'd rather focus the UX on the APIs that are used more often, for example:

```
    @objc(captureWithEvent:distinctId,properties:userProperties:userPropertiesSetOnce:groups:)
    public func capture(_ event: String,
                        distinctId: String? = nil,
                        properties: [String: Any]? = nil,
                        userProperties: [String: Any]? = nil,
                        userPropertiesSetOnce: [String: Any]? = nil,
                        groups: [String: String]? = nil)
```

marandaneto (approved):

PR #218, opened 2024-10-17
Catch race condition in PostHogFileBackedQueue.deleteFiles

jverkoey:
In rare cases, `deleteFiles(_:)` can be invoked such that, between the check for items.isEmpty and the call to items.remove, the task gets pre-empted and items is mutated. Once the thread resumes, the `.remove(at: 0)` operation causes a crash.

Tested by using my app with this patch enabled and verifying that the files are deleted as expected.

## :pencil: Checklist

### DIFF
--- a/PostHog/PostHogFileBackedQueue.swift
+++ b/PostHog/PostHogFileBackedQueue.swift
@@ -100,9 +100,14 @@ class PostHogFileBackedQueue {
 
     private func deleteFiles(_ count: Int) {
         for _ in 0 ..< count {
-            if items.isEmpty { return }
-            let removed = items.remove(at: 0) // We always remove from the top of the queue
-
+            guard let removed: String = _items.mutate({ items in
+                if items.isEmpty {
+                    return nil
+                }
+                return items.remove(at: 0) // We always remove from the top of the queue
+            }) else {
+                continue
+            }
             deleteSafely(queue.appendingPathComponent(removed))
         }
     }


PostHog/PostHogFileBackedQueue.swift
```diff
@@ -100,9 +100,14 @@ class PostHogFileBackedQueue {
...
-
+            guard let removed: String = _items.mutate({ items in
+                if items.isEmpty {
+                    return nil
+                }
+                return items.remove(at: 0) // We always remove from the top of the queue
+            }) else {
+                continue
+            }
             deleteSafely(queue.appendingPathComponent(removed))
```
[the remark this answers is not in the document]
jverkoey: Agreed, fixed.

jverkoey:
Style comments addressed :)

PostHog/PostHogFileBackedQueue.swift
```diff
@@ -100,10 +100,14 @@ class PostHogFileBackedQueue {
 
     private func deleteFiles(_ count: Int) {
         for _ in 0 ..< count {
-            if items.isEmpty { return }
-            let removed = items.remove(at: 0) // We always remove from the top of the queue
-
-            deleteSafely(queue.appendingPathComponent(removed))
+            if let removed: String = _items.mutate({ items in
```
marandaneto, line 103: L49 does something similar, maybe this patch has to be applied there as well.

PostHog/PostHogFileBackedQueue.swift
```diff
@@ -100,10 +100,14 @@ class PostHogFileBackedQueue {
...
+                if items.isEmpty {
+                    return nil
+                }
+                return items.remove(at: 0) // We always remove from the top of the queue
```
marandaneto, line 107: How did you reproduce this issue since the whole block is locked with the `isFlushingLock` and `isFlushing` flag? Can you pass the specific steps to reproduce the issue so we can check if the very same issue happens in more places as well?
I don't see how the `remove(at: 0)` would fail if the internal `items` list isn't modified anywhere else outside of this lock.

marandaneto: Checking the thread is easy (`Thread.isMainThread`), but independently of the thread, the add method should be run within the `dispatchQueue.async` block and then it's ok to lock with `isFlushingLock` and await its availability.
Again, another PR, fixing the crash here is more important now, but a follow-up with the improvement above would be cool.
https://github.com/PostHog/posthog-ios/blob/1bb1709031254f1e243f5a0de57fc9e969ec8c65/PostHog/PostHogQueue.swift#L212 is an edge case so I bet rarely happens

jverkoey: fwiw I only got one crash report with this call stack.

jverkoey:
huzzah! thank you!

PR #224, opened 2024-10-23
feat(autocapture): auto capture element interactions

ioannisj:

Autocapture feature for iOS (UIKit)

FE: https://github.com/PostHog/posthog/pull/25763, https://github.com/PostHog/posthog/pull/25996
Docs: https://github.com/PostHog/posthog.com/pull/9783

 Capture
  Elements

### Elements
- [x] UIButton
- [ ] UIMenu (and UIAction)
- [x] UIPageControl
- [x] UISearchBar
- [x] UISegmentedControl
- [x] UISlider
- [x] UISwitch
- [x] UIStepper
- [x] UITextField
- [x] UITextView
- [ ] UIAlertController  (alert)
- [ ] UIAlertController (action sheet)
- [ ] UIToolbar
- [ ] UINavigationBar
- [x] UIDatePicker
- [ ] UIColorPickerViewController
- [ ] UIFontPicker
- [ ] UIImagePickerController
- [x] UIPickerView
- [x] UIScrollView (UITableView, UICollectionView)

### Gestures
- [x] UITapGestureRecognizer
- [x] UISwipeGestureRecognizer
- [x] UIPanGestureRecognizer
- [x] UILongPressGestureRecognizer
- [x] UIPinchGestureRecognizer
- [x] UIRotationGestureRecognizer
- [x] UIScreenEdgePanGestureRecognizer

## :pencil: Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,16 +15,19 @@ jobs:
 
   release:
     name: Release
-    runs-on: macos-latest
+    runs-on: macos-14
     env:
       COCOAPODS_TRUNK_TOKEN: ${{ secrets.COCOAPODS_TRUNK_TOKEN }} # Using Manoel's token for now
 
     steps:
       - name: Git checkout
         uses: actions/checkout@v4
+      - uses: maxim-lobanov/setup-xcode@v1
+        with:
+          xcode-version: latest-stable
 
-      - name: Install Cocoapods
-        run: gem install cocoapods
+      - name: Bootstrap
+        run: make bootstrap
 
       - name: Release
         run: make releaseCocoaPods
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 ## Next
 
 - add autocapture support for UIKit ([#224](https://github.com/PostHog/posthog-ios/pull/224))
+
+## 3.14.2 - 2024-11-08
+
+- fix issue with resetting accent color in SwiftUI app ([#238](https://github.com/PostHog/posthog-ios/pull/238)) 
 - fix $feature_flag_called not captured after reloading flags ([#232](https://github.com/PostHog/posthog-ios/pull/232))
 
 ## 3.14.1 - 2024-11-05
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,6 @@
 
 build: buildSdk buildExamples
 
-# set -o pipefail && xcrun xcodebuild build -scheme PostHog -destination 'platform=visionOS Simulator,name=Apple Vision Pro' | xcpretty #visionOS
 buildSdk:
 	set -o pipefail && xcrun xcodebuild -downloadAllPlatforms
 	set -o pipefail && xcrun xcodebuild build -scheme PostHog -destination generic/platform=ios | xcpretty #ios
@@ -10,10 +9,12 @@ buildSdk:
 	set -o pipefail && xcrun xcodebuild build -scheme PostHog -destination generic/platform=macos | xcpretty #macOS
 	set -o pipefail && xcrun xcodebuild build -scheme PostHog -destination generic/platform=tvos | xcpretty #tvOS
 	set -o pipefail && xcrun xcodebuild build -scheme PostHog -destination generic/platform=watchos | xcpretty #watchOS
+	set -o pipefail && xcrun xcodebuild build -scheme PostHog -destination 'platform=visionOS Simulator,name=Apple Vision Pro' | xcpretty #visionOS
 
 buildExamples:
 	set -o pipefail && xcrun xcodebuild -downloadAllPlatforms
 	set -o pipefail && xcrun xcodebuild build -scheme PostHogExample -destination generic/platform=ios | xcpretty #ios
+	set -o pipefail && xcrun xcodebuild build -scheme PostHogExample -destination 'platform=visionOS Simulator,name=Apple Vision Pro' | xcpretty #visionOS
 	set -o pipefail && xcrun xcodebuild build -scheme PostHogObjCExample -destination generic/platform=ios | xcpretty #ObjC
 	set -o pipefail && xcrun xcodebuild build -scheme PostHogExampleMacOS -destination generic/platform=macos | xcpretty #macOS
 	set -o pipefail && xcrun xcodebuild build -scheme 'PostHogExampleWatchOS Watch App' -destination generic/platform=watchos | xcpretty #watchOS
@@ -53,10 +54,18 @@ api:
 # requires gem and brew
 # xcpretty needs 'export LANG=en_US.UTF-8'
 bootstrap:
+	gem install cocoapods
 	gem install xcpretty
 	brew install swiftlint
 	brew install swiftformat
 	brew install peripheryapp/periphery/periphery
 
+# download SDKs and runtimes
+# install any simulator(s) missing from runner image
+# release pod
 releaseCocoaPods:
+	# I think we can do without these next 2 steps but let's leave it for now
+	set -o pipefail && xcrun xcodebuild -downloadAllPlatforms 
+	# install Apple Vision Pro
+	xcrun simctl create "Apple Vision Pro" "Apple Vision Pro" "xros2.1"
 	pod trunk push PostHog.podspec --allow-warnings
--- a/Package.swift
+++ b/Package.swift
@@ -4,6 +4,7 @@ import PackageDescription
 let package = Package(
     name: "PostHog",
     platforms: [
+        // TODO: add  .visionOS(.v1), when SPM is >= 5.9
         .macOS(.v10_15), .iOS(.v13), .tvOS(.v13), .watchOS(.v6),
     ],
     products: [
--- a/PostHog.podspec
+++ b/PostHog.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name             = "PostHog"
-  s.version          = "3.14.1"
+  s.version          = "3.14.2"
   s.summary          = "The hassle-free way to add posthog to your iOS app."
 
   s.description      = <<-DESC
@@ -20,6 +20,7 @@ Pod::Spec.new do |s|
   s.tvos.deployment_target = '13.0'
   s.osx.deployment_target = "10.15"
   s.watchos.deployment_target = "6.0"
+  s.visionos.deployment_target = "1.0"
   s.swift_versions = "5.3"
 
   s.frameworks = 'Foundation'
--- a/PostHog/PostHogContext.swift
+++ b/PostHog/PostHogContext.swift
@@ -123,8 +123,10 @@ class PostHogContext {
         var properties: [String: Any] = [:]
 
         #if os(iOS) || os(tvOS)
-            properties["$screen_width"] = Float(UIScreen.main.bounds.width)
-            properties["$screen_height"] = Float(UIScreen.main.bounds.height)
+            if let screen = UIApplication.shared.windows.first?.screen {
+                properties["$screen_width"] = Float(screen.bounds.width)
+                properties["$screen_height"] = Float(screen.bounds.height)
+            }
         #elseif os(macOS)
             if let mainScreen = NSScreen.main {
                 let screenFrame = mainScreen.visibleFrame
--- a/PostHog/PostHogVersion.swift
+++ b/PostHog/PostHogVersion.swift
@@ -9,7 +9,7 @@ import Foundation
 
 // if you change this, make sure to also change it in the podspec and check if the script scripts/bump-version.sh still works
 // This property is internal only
-public var postHogVersion = "3.14.1"
+public var postHogVersion = "3.14.2"
 
 // This property is internal only
 public var postHogSdkName = "posthog-ios"


PostHog/Autocapture/AutocaptureEventProcessing.swift
```diff
@@ -0,0 +1,131 @@
...
+         - Parameters:
+            - source: The source of the event (e.g., gesture recognizer, action method, or notification).
+            - event: The autocapture event data, containing properties, screen name, and other metadata.
+
+         If the event has a `debounceInterval` greater than 0, the event is debounced.
+         This is useful for UIControls like `UISlider` that emit frequent value changes, ensuring only the last value is captured.
+         The debounce interval is defined per UIControl by the `ph_autocaptureDebounceInterval` property of `AutoCapturable`
+         */
+        func process(source: PostHogAutocaptureIntegration.EventData.EventSource, event: PostHogAutocaptureIntegration.EventData) {
+            assert(Thread.isMainThread, "Event captured off main thread")
```
marandaneto, line 42: I think this would crash the app and we don't want that.

PostHog/Autocapture/PostHogAutoCaptureConfig.swift
```diff
@@ -0,0 +1,41 @@
...
+//  PostHog
+//
+//  Created by Yiannis Josephides on 23/10/2024.
+//
+
+#if os(iOS) || targetEnvironment(macCatalyst)
+    import Foundation
+
+    @objc(PostHogAutocaptureConfig)
+    public class PostHogAutocaptureConfig: NSObject {
```
marandaneto, line 12: I think they should be disabled by default until its not experimental anymore and we know there are no side effects nor crashing peoples apps

PostHog/Autocapture/PostHogAutocaptureIntegration.swift
```diff
@@ -0,0 +1,468 @@
...
+        }
+    }
+
+    extension UIApplication {
+        @objc func ph_swizzled_uiapplication_sendAction(_ action: Selector, to target: Any?, from sender: Any?, for event: UIEvent?) -> Bool {
+            defer {
+                // Currently, the action methods pointing to a SwiftUI target are blocked.
+                let targetClass = String(describing: object_getClassName(target))
+                if targetClass.contains("SwiftUI") {
+                    print("PostHogSDK.shared.capture SwiftUI -> \(targetClass)")
```
marandaneto, line 109: `print` should not be part of the SDK, we can log though using `hedgeLog`

PostHog/Autocapture/PostHogAutocaptureIntegration.swift
```diff
@@ -0,0 +1,468 @@
...
+
+    #if !os(tvOS)
+        extension UIDatePicker {
+            override var ph_autocaptureEvents: UIControl.Event { .valueChanged }
+        }
+    #endif
+
+    private func sanitizeText(_ title: String) -> String {
+        title.replacingOccurrences(of: "[^a-zA-Z0-9.]+", with: "-", options: .regularExpression, range: nil)
+    }
```
marandaneto, line 449: Maybe we use this that already exists in the SDK?:

```
extension String {
    func mask() -> String {
        String(repeating: "*", count: count)
    }
}
```

PostHog/PostHogConfig.swift
```diff
@@ -55,6 +55,16 @@ import Foundation
...
     #endif
 
+    #if os(iOS) || targetEnvironment(macCatalyst)
+        /// Enable Autocapture for iOS
+        /// Experimental support
+        /// Default: false
+        @objc public var autocapture: Bool = false
+        /// Autocapture configuration
+        /// Experimental support
+        @objc public let autocaptureConfig: PostHogAutocaptureConfig = .init()
```
marandaneto, line 65:
```suggestion
        @objc public let autoCaptureConfig: PostHogAutocaptureConfig = .init()
```

would that be better?

PostHog/PostHogConfig.swift
```diff
@@ -55,6 +55,16 @@ import Foundation
         @objc public let sessionReplayConfig: PostHogSessionReplayConfig = .init()
     #endif
 
+    #if os(iOS) || targetEnvironment(macCatalyst)
+        /// Enable Autocapture for iOS
+        /// Experimental support
+        /// Default: false
+        @objc public var autocapture: Bool = false
```
marandaneto, line 62:
```suggestion
        @objc public var autoCapture: Bool = false
```

PostHog/Autocapture/PostHogAutoCaptureConfig.swift
marandaneto: nvm this depends on `autocapture` so ignore it.

PostHog/PostHogSDK.swift
```diff
@@ -610,6 +618,42 @@ let maxRetryDelay = 30.0
         ))
     }
 
+    func autocapture(
+        eventType: String,
+        elements: [[String: Any]],
+        elementsChain: String,
+        properties: [String: Any]
```
marandaneto, line 625: I think this one should be nullable? probably not mandatory

PostHog/UIViewController.swift
```diff
@@ -43,13 +43,15 @@
         }
 
         static func getViewControllerName(_ viewController: UIViewController) -> String? {
-            var title: String? = String(describing: viewController.classForCoder).replacingOccurrences(of: "ViewController", with: "")
-
-            if title?.isEmpty == true {
-                title = viewController.title ?? nil
+            let viewControllerClass = String(describing: type(of: viewController))
```
marandaneto, line 46: What did change here exactly? would the results be different before and after merging this? this could skew the data if it was previously one thing and after another thing, so we need to consider that.

PostHogExampleAutocapture/LICENSE/LICENSE.txt
```diff
@@ -0,0 +1,8 @@
+Copyright © 2021 Apple Inc.
```
marandaneto, line 1: why do we need this?

PostHog/Autocapture/AutocaptureEventProcessing.swift
```diff
@@ -0,0 +1,131 @@
...
+#if os(iOS) || targetEnvironment(macCatalyst)
+    import UIKit
+
+    protocol AutocaptureEventProcessing: AnyObject {
+        func process(source: PostHogAutocaptureIntegration.EventData.EventSource, event: PostHogAutocaptureIntegration.EventData)
+    }
+
+    class PostHogAutocaptureEventProcessor: AutocaptureEventProcessing {
+        private weak var postHogInstance: PostHogSDK?
+        private weak var postHogConfig: PostHogConfig?
```
marandaneto, line 17: I think config does not need to be nullable nor `weak`.

PostHog/Autocapture/AutocaptureEventProcessing.swift
```diff
@@ -0,0 +1,131 @@
...
+
+#if os(iOS) || targetEnvironment(macCatalyst)
+    import UIKit
+
+    protocol AutocaptureEventProcessing: AnyObject {
+        func process(source: PostHogAutocaptureIntegration.EventData.EventSource, event: PostHogAutocaptureIntegration.EventData)
+    }
+
+    class PostHogAutocaptureEventProcessor: AutocaptureEventProcessing {
+        private weak var postHogInstance: PostHogSDK?
```
marandaneto, line 16: can we just use the `PostHogSDK.shared` directly? I know it's not great for testability, but we don't need to handle nullable things, I have an idea to improve this.

PostHog/Autocapture/AutocaptureEventProcessing.swift
```diff
@@ -0,0 +1,131 @@
...
+
+        init(postHogInstance: PostHogSDK, configuration: PostHogConfig) {
+            self.postHogInstance = postHogInstance
+            self.postHogConfig = configuration
+            PostHogAutocaptureIntegration.addEventProcessor(self)
+        }
+
+        deinit {
+            PostHogAutocaptureIntegration.removeEventProcessor(self)
+        }
```
marandaneto, line 28: I'd like to understand a bit better how those event processors work, we can discuss this after the daily over a call tomorrow?

PostHog/Autocapture/AutocaptureEventProcessing.swift
marandaneto: Since user's builds are not in our control, as an SDK, crashing is not acceptable, in this case, I'd be strongly against `assert`s, `hedgeLog`, or any other alternative that would be better here because this will come back and bite us, defensive programming is very important in SDK dev.

PostHog/Autocapture/PostHogAutocaptureIntegration.swift
marandaneto: Mm, yeah, I'm not sure why that was done. I thought it was masking, but it's sanitizing special characters and spaces. I did not pay attention to it. What does the JS SDK do?

PostHog/PostHogConfig.swift
marandaneto: makes sense, i have not found any form of Auto Capture in the docs so lets keep as it is

PostHog/PostHogSDK.swift
marandaneto: `autocapture` is public in js-lite at least.

PostHog/Autocapture/PostHogAutocaptureIntegration.swift
marandaneto: I see, your call then, either like js-lite or js is is fine since both are working and nobody complained.

PostHog/PostHogSDK.swift
marandaneto: it does not need to be public, so your call here, I just checked the js-lite that's why I commented.

PostHog/Autocapture/AutocaptureEventProcessing.swift
marandaneto: All valid points, I'd focus on getting it working with the shared instance for now, but indeed the multiple instances is an issue (it is already since it does not work).
I'd say that all "integrations/autocapture" etc work with the shared instance and if you create more instances, you'd need to do everything (capture events, etc) all by hand since it's an edge case and most likely is used by SDKs/Libs, which should not capture events from the hosting app anyway.
Yes, you can only swizzle once, or, more times but the last swizzle wins and that leads to more complexity to get everything working, so I'd rather not do this unless we have to at some point.
The auto capture can be an integration, and the config has a list of integrations, for example you can see how this works on Android
https://github.com/PostHog/posthog-android/blob/351a86b7c7caf9c72aad684d3721fad2872cfa2c/posthog/src/main/java/com/posthog/PostHogConfig.kt#L156-L169

PostHog/Autocapture/PostHogAutocaptureIntegration.swift
```diff
@@ -0,0 +1,142 @@
...
+            )
+        }
+
+        /// Determines whether an autocapture event should be processed based on the event source and the current PostHog configuration.
+        ///
+        /// Note: Not sure if we should be checking feature flags yet here
+        ///
+        /// - Parameter source: source to be processed
+        private func shouldProcess(source: PostHogAutocaptureEventTracker.EventData.EventSource) -> Bool {
+            guard PostHogSDK.shared.isAutocaptureActive() else { return false }
```
marandaneto, line 130: is we make `func isEnabled()` public, we could use it directly here, since you already have the `config` instance

PostHog/Autocapture/PostHogAutocaptureIntegration.swift
```diff
@@ -0,0 +1,142 @@
...
+        func start() {
+            PostHogAutocaptureEventTracker.eventProcessor = self
+            hedgeLog("Autocapture integration started")
+        }
+
+        /**
+         Disables the autocapture integration by clearing the PostHogAutocaptureEventTracker routing
+         */
+        func stop() {
+            if PostHogAutocaptureEventTracker.eventProcessor != nil {
```
marandaneto, line 35: you probably want to clear `debounceTimers` here

PostHog/Autocapture/PostHogAutocaptureEventTracker.swift
```diff
@@ -0,0 +1,533 @@
...
+            let screenName: String?
+            let viewHierarchy: [ViewNode]
+            let targetClass: String
+            let accessibilityLabel: String?
+            let accessibilityIdentifier: String?
+            // values >0 means that this event will be debounced for `debounceInterval`
+            let debounceInterval: TimeInterval
+        }
+
+        static var eventProcessor: (any AutocaptureEventProcessing)? {
```
marandaneto, line 45: Could this be just a method instead of a static instance class?

marandaneto:
I wonder if at least some parts can be unit-tested?

PostHog/Autocapture/PostHogAutocaptureIntegration.swift
```diff
@@ -0,0 +1,142 @@
...
+            - source: The source of the event (e.g., gesture recognizer, action method, or notification).
+            - event: The autocapture event data, containing properties, screen name, and other metadata.
+
+         If the event has a `debounceInterval` greater than 0, the event is debounced.
+         This is useful for UIControls like `UISlider` that emit frequent value changes, ensuring only the last value is captured.
+         The debounce interval is defined per UIControl by the `ph_autocaptureDebounceInterval` property of `AutoCapturable`
+         */
+        func process(source: PostHogAutocaptureEventTracker.EventData.EventSource, event: PostHogAutocaptureEventTracker.EventData) {
+            if !Thread.isMainThread {
+                hedgeLog("Autocapture event processed off main thread. This should not happen.")
```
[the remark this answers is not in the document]
marandaneto: in this case, the check is unnecessary? I feel like if it should not happen and it's not allowed, bailout, otherwise, the check should not be there.

marandaneto:
> > I wonder if at least some parts can be unit-tested?
> 
> Yeah, I can add some on config, sanitization and potentially swizzling. For the rest, not sure it's going to be effective without diving into UI tests?

I'd not focus on UI tests now, just the bits that are possible via unit tests but still asserting/testing the feature.

PostHog/Autocapture/PostHogAutocaptureEventTracker.swift
marandaneto: this commit right? [`8b1406f` (#224)](https://github.com/PostHog/posthog-ios/pull/224/commits/8b1406fc7e2bb378bc9d5facaa469596406ad21c)
I meant something that doesn't need a weak ref and static ref
just a class method that processes what's needed.
having more weak refs and static vars would make the refactoring for multiple instances later on harder.

marandaneto: The 2nd point feels something closer to what I thought.
On pto now but feel free to get this ball rolling and dont get blocked, we can always improve later if needed :) so ✅

marandaneto:
> Working on the docs made me realize that this new configuration could be confusing to the users.
> 
> We already have `captureApplicationLifecycleEvents` and `captureScreenViews` which could both be considered an Autocapture feature.
> 
> Currently, the config looks like this:
> 
> ```swift
> config.autocapture = true // $autocapture support
> config.autocaptureConfig = true // configuration for autocapture element interactions
> config.captureApplicationLifecycleEvents = true // Application Opened etc
> config.captureScreenViews = true // $screen support
> ```
> 
> Ideally, all of these configs should be nested in `autocaptureConfig` and controlled via the `autocapture` option. I don't want to introduce a breaking change especially for an experimental feature, so I propose the following
> 
> ```swift
> config.captureApplicationLifecycleEvents = true // Application Open etc
> config.captureScreenViews = true // $screen support
> config.captureInteractions = true // $autocapture support
> ```
> 
> Once this is battle tested we can think about nesting or grouping the configurations.
> 
> What do you guys think? @marandaneto @pauldambra

While I agree with the confusion, JS is the same, see https://posthog.com/docs/product-analytics/autocapture
autocapture != capture_pageview, they are different things and the same applies here.
Maybe we should stop calling the app life cycle events, screen views, etc as autocapture and define autocapture as its documented in the [docs](https://posthog.com/docs/product-analytics/autocapture) as only auto capture control events such as clicks, taps, etc.
In this case, I just would fix the headers in the docs as "automatically captured" for those screen views, app lifecycle events, etc and "autocapture" for the auto capture feature.

> ```swift
> config.captureApplicationLifecycleEvents = true // Application Open etc
> config.captureScreenViews = true // $screen support
> config.autocapture = true // $autocapture support
> config.autocaptureConfig {
> ...
> }
> ```

how does that sound? it was probably my mistake to call such events "autocapture" as well because it was unclear to me that autocapture was only about the $autocapture events that are only captured for the control events.

marandaneto:
> However, in the intro section of the docs, under Autocapture, pageview and screen are included. I think this is what throws me off a bit for both.

@ioannisj it's a good point, I also noticed that when reading it again.
Double-check with the prod analytics team about that please, if page views and screens are not "autocapture", then we should fix the docs https://posthog.com/docs/product-analytics/autocapture or consider them as "autocapture" as well on mobile SDKs which is what I did before, but we need to be more clear on the docs about the different configurations.

PR #230, opened 2024-11-05
recording: fix RN iOS masking

marandaneto:
Closes https://github.com/PostHog/posthog-js-lite/issues/292

Running multiple tests with multiple views/accesibility labels etc

## :pencil: Checklist

### DIFF
--- a/PostHog/Replay/PostHogReplayIntegration.swift
+++ b/PostHog/Replay/PostHogReplayIntegration.swift
@@ -249,6 +249,12 @@
                 }
             }
 
+            // manually masked views through view modifier `PostHogMaskViewModifier`
+            if view.phIsManuallyMasked {
+                maskableWidgets.append(view.toAbsoluteRect(parent))
+                return
+            }
+
             // on RN, lots get converted to RCTRootContentView, RCTRootView, RCTView and sometimes its just the whole screen, we dont want to mask
             // in such cases
             if view.isNoCapture() || maskChildren {
--- a/PostHog/Replay/PostHogReplayIntegration.swift
+++ b/PostHog/Replay/PostHogReplayIntegration.swift
@@ -30,6 +30,9 @@
         private let swiftUIGenericTypes = ["_TtCOCV7SwiftUI11DisplayList11ViewUpdater8Platform13CGDrawingView",
                                            "_TtC7SwiftUIP33_A34643117F00277B93DEBAB70EC0697122_UIShapeHitTestingView"].compactMap { NSClassFromString($0) }
 
+        private let reactNativeTextView: AnyClass? = NSClassFromString("RCTTextView")
+        private let reactNativeImageView: AnyClass? = NSClassFromString("RCTImageView")
+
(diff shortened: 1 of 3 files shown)


PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -311,11 +339,11 @@
         }
 
         private func hasText(_ text: String?) -> Bool {
-            if let text, !text.isEmpty {
-                true
+            if let text = text, !text.isEmpty {
```
marandaneto, line 342: `make format` does that, i think we are using different lint and format versions @ioannisj ?

PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -232,9 +249,18 @@
...
+            // in such cases
+            if view.isNoCapture() || maskChildren {
+                let viewRect = view.toAbsoluteRect(parent)
+                let parentRect = parent.frame
+
+                // Check if the rectangles do not match
+                if !viewRect.equalTo(parentRect) {
+                    maskableWidgets.append(view.toAbsoluteRect(parent))
+                } else {
+                    maskChildren = true
```
marandaneto, line 262: if the parent has the ph-no-capture label but its the whole screen, we still want to mask its children

PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -232,9 +249,18 @@
                 }
             }
 
-            // manually masked views through view modifier `PostHogMaskViewModifier`
-            if view.phIsManuallyMasked {
```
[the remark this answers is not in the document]
marandaneto: i think this was a merge conflict, gonna fix it, good catch

PostHog/Replay/PostHogReplayIntegration.swift
marandaneto: `parent.frame` is the UIWindow, maybe parent is a bad name but its the view of the whole screen

PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -235,6 +252,21 @@
             // manually masked views through view modifier `PostHogMaskViewModifier`
             if view.phIsManuallyMasked {
                 maskableWidgets.append(view.toAbsoluteRect(parent))
+                return
```
marandaneto, line 255: @ioannisj I wonder if you missed that or if there was a reason to not return early since the view is maskLookingking at the children would not make sense if all childwerel be masked already

marandaneto: I renamed `parent` to `window` so its more obvious as well.
good point with `clipsToBounds`, I was not aware of this flag.

marandaneto: funny enough:
```
            if !view.clipsToBounds {
                return
            }
```

`clipsToBounds` is false for the window.

marandaneto: i will leave it without for now

Issue #211, opened 2024-10-11
$feature_flag_called should be triggered again for an already trigged flag if there are new values


marandaneto:
### Version

Latest

### Steps to Reproduce

Relates to https://github.com/PostHog/posthog-js-lite/pull/281
Read a flag
Reload flags
Read a flag
$feature_flag_called is captured only the first time is called, but not the 2nd which is fine when the result doesn't change, but if it does, it should be triggered.
Since we don't keep track of the results, we just clear the Set that check if flags were captured or not after the flags are reloaded.

The very same issue happens on Android

- [ ] Android
- [ ] iOS

### Expected Result

$feature_flag_called event is captured again with the new value

### Actual Result

$feature_flag_called event isn't captured again

PR #232, opened 2024-11-05
fix: feature_flag_called after reloadFeatureFlags

ioannisj:

Relates to: https://github.com/PostHog/posthog-ios/issues/211

Clear 'flagCallReported'' after reload so that $feature_flag_called can be captured

## :pencil: Checklist

### DIFF
--- a/PostHog/PostHogSDK.swift
+++ b/PostHog/PostHogSDK.swift
@@ -30,7 +30,7 @@ let maxRetryDelay = 30.0
     private let setupLock = NSLock()
     private let optOutLock = NSLock()
     private let groupsLock = NSLock()
-    private let featureFlagsLock = NSLock()
+    private let flagCallReportedLock = NSLock()
     private let personPropsLock = NSLock()
 
     private var queue: PostHogQueue?
@@ -346,7 +346,7 @@ let maxRetryDelay = 30.0
         // storage also removes all feature flags
         storage?.reset()
         config.storageManager?.reset()
-        featureFlagsLock.withLock {
+        flagCallReportedLock.withLock {
             flagCallReported.removeAll()
         }
         PostHogSessionManager.shared.endSession {
@@ -795,7 +795,7 @@ let maxRetryDelay = 30.0
             anonymousId: storageManager.getAnonymousId(),
             groups: groups ?? [:],
             callback: {
-                self.featureFlagsLock.withLock {
+                self.flagCallReportedLock.withLock {
                     self.flagCallReported.removeAll()
                 }
                 callback()
@@ -852,17 +852,21 @@ let maxRetryDelay = 30.0
     }
 
     private func reportFeatureFlagCalled(flagKey: String, flagValue: Any?) {
-        featureFlagsLock.withLock {
+        var shouldCapture = false
+
+        flagCallReportedLock.withLock {
             if !flagCallReported.contains(flagKey) {
-                let properties: [String: Any] = [
-                    "$feature_flag": flagKey,
-                    "$feature_flag_response": flagValue ?? "",
-                ]
-
                 flagCallReported.insert(flagKey)
-
-                capture("$feature_flag_called", properties: properties)
+                shouldCapture = true
             }
+        }
+
+        if shouldCapture {
+            let properties = [
+                "$feature_flag": flagKey,
+                "$feature_flag_response": flagValue ?? "",
+            ]
+            capture("$feature_flag_called", properties: properties)
         }
     }
 
@@ -936,7 +940,7 @@ let maxRetryDelay = 30.0
                 self.reachability?.stopNotifier()
                 reachability = nil
             #endif
-            featureFlagsLock.withLock {
+            flagCallReportedLock.withLock {
                 flagCallReported.removeAll()
             }
             context = nil
--- a/PostHogTests/PostHogSDKTest.swift
+++ b/PostHogTests/PostHogSDKTest.swift
@@ -852,9 +852,10 @@ class PostHogSDKTest: QuickSpec {
         }
 
         it("captures $feature_flag_called when getFeatureFlag is called") {
-            let sut = self.getSut(sendFeatureFlagEvent: true)
-            let group = DispatchGroup()
-            group.enter()
+            let sut = self.getSut(
+                sendFeatureFlagEvent: true,
+                flushAt: 1
+            )
 
             _ = sut.getFeatureFlag("some_key")
 


CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## Next
 
+- fix $feature_flag_called not captued after reloading flags ([#232](https://github.com/PostHog/posthog-ios/pull/232))
```
marandaneto, line 3:
```suggestion
- fix $feature_flag_called not captured after reloading flags ([#232](https://github.com/PostHog/posthog-ios/pull/232))
```

PostHog/PostHogFeatureFlags.swift
```diff
@@ -99,9 +99,9 @@ class PostHogFeatureFlags {
                 else {
                     hedgeLog("Error: Decide response missing correct featureFlags format")
 
-                    self.notifyAndRelease()
+                    callback([])
 
-                    return callback()
+                    return self.notifyAndRelease()
```
marandaneto, line 104: the callback can be executed after releasing the locks, or is there a reason to change this?

PostHogTests/PostHogSDKTest.swift
```diff
@@ -847,6 +847,98 @@ class PostHogSDKTest: QuickSpec {
...
+                preloadFeatureFlags: false,
+                sendFeatureFlagEvent: true,
+                flushAt: 2
+            )
+
+            // this should capture with value nil
+            _ = sut.getFeatureFlag(server.flagKey)
+
+            sut.reloadFeatureFlags {
+                DispatchQueue.main.async {
```
marandaneto, line 929: why do we use dispatch queue here?

PostHog/PostHogSDK.swift
```diff
@@ -791,7 +791,10 @@ let maxRetryDelay = 30.0
             distinctId: storageManager.getDistinctId(),
             anonymousId: storageManager.getAnonymousId(),
             groups: groups ?? [:],
-            callback: callback
+            callback: { updatedKeys in
+                self.flagCallReported.subtract(updatedKeys)
```
marandaneto, line 795: I think `flagCallReported` will need its own lock now since its mutated from different threads

PostHog/PostHogFeatureFlags.swift
```diff
@@ -127,25 +127,44 @@ class PostHogFeatureFlags {
...
-                }
 
-                self.notifyAndRelease()
+                    self.setCachedFeatureFlags(newFeatureFlags)
+                    self.setCachedFeatureFlagPayload(newFeatureFlagPayloads)
+
+                    // calculate a set of keys that were updated, added or removed
+                    //
+                    // Note: payload changes are not currently tracked, as this update mechanism
+                    //       is primarily used for capturing `$feature_flag_called` events tied to flag values.
```
marandaneto, line 151: The goal is to clear the dict with flags when reloading/resetting/closing the SDK, but here we are comparing whether the keys and values changed, which adds more complexity.
JS SDK does not do that either, so I don't think we should do that either, the simpler the better here.
I'd expect to see a read flag event if I reloaded my flags.

PostHogTests/PostHogSDKTest.swift
marandaneto: the event is sent async thats why the test will fail, `getBatchedEvents(server)` will await so test will succeed, if you need to read flags before calling this and during `reloadFeatureFlags`, you will need another approach, I think using ` DispatchQueue.main.async` here will make the test flaky because its not guaranteed that executes before the test ends

PostHog/PostHogSDK.swift
```diff
@@ -791,7 +791,10 @@ let maxRetryDelay = 30.0
             distinctId: storageManager.getDistinctId(),
             anonymousId: storageManager.getAnonymousId(),
             groups: groups ?? [:],
-            callback: callback
+            callback: {
+                self.flagCallReported.removeAll()
```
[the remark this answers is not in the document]
marandaneto: We dont need to compare anything, just clear the list if flags are reloaded

PostHogTests/PostHogSDKTest.swift
marandaneto: the call to `getBatchedEvents` is before you do `DispatchQueue.main.async` so in this case when you call `DispatchQueue.main.async` the async request was likely still happening, you'd need to call `getBatchedEvents` before, that's the method that awaits the async processes

marandaneto: or even better, `let requests = getDecideRequest(server)` should await all pendings requests to the decide api

PostHog/PostHogSDK.swift
```diff
@@ -30,6 +30,7 @@ let maxRetryDelay = 30.0
     private let setupLock = NSLock()
     private let optOutLock = NSLock()
     private let groupsLock = NSLock()
+    private let featureFlagsLock = NSLock()
```
marandaneto, line 33: I'd rename the lock to `flagCallReportedLock` or something like that since its not really related to the feature flags lock itself

PostHog/PostHogSDK.swift
```diff
@@ -844,15 +852,17 @@ let maxRetryDelay = 30.0
...
+                let properties: [String: Any] = [
+                    "$feature_flag": flagKey,
+                    "$feature_flag_response": flagValue ?? "",
+                ]
 
-            flagCallReported.insert(flagKey)
+                flagCallReported.insert(flagKey)
 
-            capture("$feature_flag_called", properties: properties)
+                capture("$feature_flag_called", properties: properties)
```
marandaneto, line 864: I'd move this outside of the lock, the only concurrent part is the mutable `flagCallReported`

PostHogTests/PostHogSDKTest.swift
```diff
@@ -847,6 +850,51 @@ class PostHogSDKTest: QuickSpec {
             sut.reset()
             sut.close()
         }
+
+        it("captures $feature_flag_called when getFeatureFlag is called") {
+            let sut = self.getSut(sendFeatureFlagEvent: true)
+            let group = DispatchGroup()
+            group.enter()
```
marandaneto, line 857: why do you need the `group` here if not used?

PostHogTests/PostHogSDKTest.swift
```diff
@@ -847,6 +850,51 @@ class PostHogSDKTest: QuickSpec {
             sut.reset()
             sut.close()
         }
+
+        it("captures $feature_flag_called when getFeatureFlag is called") {
+            let sut = self.getSut(sendFeatureFlagEvent: true)
```
marandaneto, line 855: should it be `flushAt: 1` here?

marandaneto:
> Hey just reverted changes. So we now just clear the set when flags are reloaded

nice, did another pass and left a few comments

marandaneto (approved):

PostHogTests/PostHogSDKTest.swift
```diff
@@ -10,18 +10,21 @@ import Nimble
 import Quick
 
 @testable import PostHog
+import XCTest
```
marandaneto, line 13: was that missing before? i think its not needed when using `QuickSpec`

PR #233, opened 2024-11-07
fix: flaky tests when adding groupds and reloading flags

marandaneto:
https://posthog.slack.com/archives/C03PB072FMJ/p1730970776839319?thread_ts=1730881816.333539&cid=C03PB072FMJ
_#skip-changelog_

## :pencil: Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,6 +46,7 @@ lint:
 	swiftformat . --lint --swiftversion 5.3 && swiftlint
 
 # requires gem and brew
+# xcpretty needs 'export LANG=en_US.UTF-8'
 bootstrap:
 	gem install xcpretty
 	brew install swiftlint
--- a/Makefile
+++ b/Makefile
@@ -31,6 +31,8 @@ swiftLint:
 swiftFormat:
 	swiftformat . --swiftversion 5.3
 
+# use -test-iterations 10 if you want to run the tests multiple times
+# use -only-testing:PostHogTests/PostHogQueueTest to run only a specific test
 testOniOSSimulator:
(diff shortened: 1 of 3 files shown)


marandaneto:
> Hey @marandaneto,
> 
> Honestly, I'm not a fan of this approach. Adding a flag in the code—whether as a property or a compiler flag—just to handle a specific unit test doesn’t seem ideal to me (since we are altering the behavior of something we are trying to test). If the code is working but the tests are failing, then we can think of a different way to test that unit or remove the test altogether. I'd vote for removing the test tbh

I'm strongly against removing the test otherwise if we change things we'd not know if we have broken something.
I agree that the approach isn't ideal but it's a quick fix at least, the other option would be to allow mocking many more things.
I also do that on [Android](https://github.com/PostHog/posthog-android/blob/d9052d0232d79211c73b2f5fd972e4e5efac2b27/posthog/src/main/java/com/posthog/PostHog.kt#L46) and it works well since its a simple boolean check, I'd not rely on this approach for more complicated stuff.
Another solution would be to add delays between the internal reload flags call and the manual calls, but that slows down the test pipeline.
If you don't agree with the approach but removing is not an option, can you come up with a quick solution? I don't think it's worth spending too much time refactoring things or mocking the world because of something very specific that affects only testing.

marandaneto:
> Btw if the underlying issue is that calling `group()` already reloads feature flags internally, let's test for that as well? So wait for the first decide call, then manually reload flags and wait for the second call. I don't think it's going to slow down the test that much

discussed offline and decided to LGTM

Issue #237, opened 2024-11-08
PostHog causes accent color in SwiftUI app to be reset


terabyte128:
### Version

3.14.1

### Steps to Reproduce

1. Initialize PostHog in a SwiftUI iOS app, like this:
```swift
let config = PostHogConfig(apiKey: POSTHOG_API_KEY, host: POSTHOG_HOST)
config.captureScreenViews = false
config.sessionReplay = true
config.sessionReplayConfig.screenshotMode = true
config.sessionReplayConfig.maskAllTextInputs = false
config.sessionReplayConfig.maskAllImages = false
PostHogSDK.shared.setup(config)
```
2. Run the app

### Expected Result

The accent color defined in the app's assets is respected.

### Actual Result

It is not.

I believe this is a regression of of #54. It seems that when the library was re-written in Swift, it began using the deprecated `UIScreen.main` symbol again. Applying the following diff appears to fix it:

```diff
diff --git a/PostHog/PostHogContext.swift b/PostHog/PostHogContext.swift
index 2042c58e8..54fd1d4b7 100644
--- a/PostHog/PostHogContext.swift
+++ b/PostHog/PostHogContext.swift
@@ -123,8 +123,10 @@ class PostHogContext {
         var properties: [String: Any] = [:]

         #if os(iOS) || os(tvOS)
-            properties["$screen_width"] = Float(UIScreen.main.bounds.width)
-            properties["$screen_height"] = Float(UIScreen.main.bounds.height)
+            if let screen = UIApplication.shared.windows.first?.screen {
+                properties["$screen_width"] = Float(screen.bounds.width)
+                properties["$screen_height"] = Float(screen.bounds.height)
+            }
         #elseif os(macOS)
             if let mainScreen = NSScreen.main {
                 let screenFrame = mainScreen.visibleFrame
```

PR #238, opened 2024-11-08
Fix SwiftUI screen size handling on iOS

terabyte128:
UIScreen.main is deprecated and calling it appears to reset the accent color of the UI.

See #237 

I verified that the change fixes the accent color issue. I further verified that calling `UIScreen.main` actually causes the issue in the first place, by inserting the following into my code and observing the accent color change:

```swift
let x = UIScreen.main.bounds.width
```

## :pencil: Checklist


PostHog/PostHogContext.swift
```diff
@@ -123,8 +123,10 @@ class PostHogContext {
         var properties: [String: Any] = [:]
 
         #if os(iOS) || os(tvOS)
-            properties["$screen_width"] = Float(UIScreen.main.bounds.width)
-            properties["$screen_height"] = Float(UIScreen.main.bounds.height)
+            if let screen = UIApplication.shared.windows.first?.screen {
```
dfed, line 126: This change means that all iOS/tvOS consumers receive a warning unless we're only calling into PostHog from the main thread. If a client's logging system runs off of `main`, accessing `UIApplication.shared` will cause warnings in Xcode. Not sure if that's a priority for y'all, but figured I'd raise.

marandaneto: We can check if the captured event is on the main thread and only do the thread jumping if needed to reduce the perf impact as well.
`UIApplication.shared.windows` is also deprecated so the change from A to B didn't make sense IMO.

marandaneto: `PostHogReplayIntegration.getCurrentWindow()` is likely the correct way right now, we should move `getCurrentWindow()` out of `PostHogReplayIntegration` and reuse across the SDK, I see more places using `UIApplication.shared` that is just trying to find the active `window`, eg `UIViewController.activeController` and `UIApplication.ph_currentWindow`

PR #247, opened 2024-11-11
feat: get current window screen size

ioannisj:

thread: https://github.com/PostHog/posthog-ios/pull/238?notification_referrer_id=NT_kwDOCu3_cbUxMzI2MDI1NDM1MDoxODMzNjk1ODU#discussion_r1835802465

- Switch and block to main thread if needed to grab screen size 
- Added a shared module extension to retrieve current window from UIApplication

#skip-changelog
## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Next
 
+- fix: accessing UI APIs off main thread to get screen size ([#247](https://github.com/PostHog/posthog-ios/pull/247))
+
 ## 3.15.0 - 2024-11-11
 
 - add autocapture support for UIKit ([#224](https://github.com/PostHog/posthog-ios/pull/224))


PostHog/UIViewController.swift
```diff
@@ -29,17 +29,9 @@
...
-                    let windowScene = scene as? UIWindowScene
-                    let sceneDelegate = windowScene?.delegate as? UIWindowSceneDelegate
-                    if let target = sceneDelegate, let window = target.window {
-                        return window?.rootViewController
-                    }
-                }
             }
-            return nil
+            // TODO: handle container controllers (see ph_topViewController)
+            return UIApplication.getCurrentWindow()?.rootViewController
```
marandaneto, line 34: This part isn't checking the `windowScene?.delegate as? UIWindowSceneDelegate` anymore, any reason to drop this check?

PostHog/Autocapture/PostHogAutocaptureEventTracker.swift
```diff
@@ -342,17 +342,8 @@
...
-            Array(UIApplication.shared.connectedScenes)
-                .compactMap { $0 as? UIWindowScene }
-                .flatMap(\.windows)
-                .first { $0.windowLevel != .statusBar }
-        }
-    }
-
     extension UIViewController {
-        class func ph_topViewController(base: UIViewController? = UIApplication.ph_currentWindow?.rootViewController) -> UIViewController? {
+        class func ph_topViewController(base: UIViewController? = UIApplication.getCurrentWindow()?.rootViewController) -> UIViewController? {
```
marandaneto, line 346: getCurrentWindow now isn't checking `windowLevel != .statusBar`, is that ok?

PostHog/Replay/UIApplicationTracker.swift
```diff
@@ -44,7 +44,7 @@
             guard event.type == .touches else {
                 return
             }
-            guard let window = PostHogReplayIntegration.getCurrentWindow() else {
+            guard let window = UIApplication.getCurrentWindow() else {
```
marandaneto, line 47: This likely needs `checkForegrounded: true` since its swizzled in the main thread during touches, previously it did and now it does not, any reason why not?
Touches and screenshots need to be from the very same screen otherwise click on screen A might be appearing on screen B and vice versa I believe (until we have proper support for multi windows)

PostHog/Utils/UIApplication+.swift
```diff
@@ -0,0 +1,31 @@
...
+//  UIApplication+.swift
+//  PostHog
+//
+//  Created by Yiannis Josephides on 11/11/2024.
+//
+
+import UIKit
+
+extension UIApplication {
+    static func getCurrentWindow(checkForegrounded: Bool = false) -> UIWindow? {
```
marandaneto: I think we could invert the default value here, the normal is to check the current window, which means active by default (and in the foreground).
Only events captured in the background thread would have to skip this check because there's no active window but we still want its metadata, but that hs the exception.

PostHog/UIViewController.swift
```diff
@@ -29,17 +29,9 @@
             // if a view is being dismissed, this will return nil
             if let root = viewIfLoaded?.window?.rootViewController {
                 return root
-            } else {
-                // preferred way to get active controller in ios 13+
-                for scene in UIApplication.shared.connectedScenes where scene.activationState == .foregroundActive {
-                    let windowScene = scene as? UIWindowScene
-                    let sceneDelegate = windowScene?.delegate as? UIWindowSceneDelegate
-                    if let target = sceneDelegate, let window = target.window {
```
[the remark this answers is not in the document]
marandaneto: I'm not 100% sure but I believe the reason to check the target `delegate` is that the delegate has the `rootViewController` but on the other approach not but I don't remember for sure, better to test this before causing a regression, otherwise all good dropping the check.

PostHog/PostHogContext.swift
```diff
@@ -18,6 +18,22 @@ class PostHogContext {
...
+            #elseif os(macOS)
+                return NSScreen.main?.visibleFrame.size
+            #else
+                return nil
+            #endif
+        }
+
+        return Thread.isMainThread
+            ? getWindowSize()
+            : DispatchQueue.main.sync { getWindowSize() }
```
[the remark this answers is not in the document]
marandaneto: there's no way to get this info from watchos then?

PostHog/Utils/UIApplication+.swift
```diff
@@ -0,0 +1,42 @@
...
+            if #available(iOS 15.0, tvOS 15.0, *) {
+                // attempt to retrieve directly from UIWindowScene
+                if let keyWindow = windowScenes.compactMap(\.keyWindow).first {
+                    return keyWindow
+                }
+            }
+
+            // fall bar to UIWindowSceneDelegate
+            for scene in UIApplication.shared.connectedScenes where scene.activationState == .foregroundActive {
+                let windowScene = scene as? UIWindowScene
```
marandaneto, line 29: This part repeats with L13-15, I guess we can assign that to a var and reuse it so we don't have to traverse the list and do the same checks again.
I also think this should be the last fallback, since the L37 is the same check as L22 but using an older API since `keyWindow` is only available >= iOS 15

PostHog/UIViewController.swift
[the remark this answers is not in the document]
marandaneto: > // if a view is being dismissed, this will return nil

there's this comment, I believe its because of that, and in this case, maybe we should use `filterForegrounded: false`? its a fallback anyway.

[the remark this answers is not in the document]
marandaneto: > I think we should consider removing the check on UIScene.activationState entirely and instead rely on UIApplication.shared.applicationState to determine whether to skip certain screen captures or not.

good point, I think we can do a follow up on that later so we don't drag this PR too much

PostHog/Utils/UIApplication+.swift
```diff
@@ -0,0 +1,45 @@
...
+                        return keyWindow
+                    }
+                }
+
+                // check scene.windows.isKeyWindow
+                for window in scene.windows {
+                    if window.isKeyWindow {
+                        return window
+                    }
+                }
```
marandaneto, line 33: can this be an else branch for `if #available`? otherwise its being executed twice for iOS >= 15, the former and this one when its not a keyWindow.

marandaneto (approved):

PostHog/PostHogContext.swift
```diff
@@ -11,13 +11,33 @@ import Foundation
...
+            #elseif os(watchOS)
+                return WKInterfaceDevice.current().screenBounds.size
+            #else
+                return nil
+            #endif
+        }
+
+        return Thread.isMainThread
+            ? getWindowSize()
+            : DispatchQueue.main.sync { getWindowSize() }
```
dfed, line 38: @ioannisj this `.sync` call creates a deadlock in my app. We wrap `PostHogSDK.shared.isFeatureEnabled` calls in a lock so that we can return stable values for a feature flag within one app lifecycle, like so:
```swift
	public func isEnabled(_ key: String) -> Bool {
		lock.sync {
			if let deliveredIsEnabled = unsafe_keyToPreviouslyDeliveredIsEnabled[key] {
				// Avoid potential bugs by always returning a stable value.
				// Do not allow a feature flag to change mid app lifecycle.
				return deliveredIsEnabled
			} else {
				let isEnabled = PostHogSDK.shared.isFeatureEnabled(key)
				unsafe_keyToPreviouslyDeliveredIsEnabled[key] = isEnabled
				return isEnabled
			}
		}
	}
```

On app launch, we call into this `isEnabled(…)` method from multiple queues, including main. If our background queue calls into this method just before main calls into this method, then the background queue gets the lock, the main queue waits on the lock, and then your code here waits forever for main to continue.

While this code exists in PostHog we will be unable to upgrade from our current version.

dfed: Ideally you wouldn't need to synchronously access this data just-in-time: a possible approach would be to intercept changes to the current window synchronously, capturing the window-specific values you need, and then storing those values in a locked container that background queues can access without getting a lock on all of `main`.

marandaneto: maybe a simpler approach/quick workaround would be:

```
if Thread.isMainThread {
            return getWindowSize()
        } else {
            var size: CGSize?
            let group = DispatchGroup()
            group.enter()
            DispatchQueue.main.async {
                size = getWindowSize()
                group.leave()
                
            }
            let timeout = DispatchTime.now() + .seconds(1)
            let _ = group.wait(timeout: timeout)
            return size
        }
```

or yes, listening to window changes and caching the values

dfed: That quick workaround would alleviate the deadlock but would cause a 1s hang on the main queue any time we would have deadlocked 😬

Certainly better than a deadlock, but also not something I can ship with.

Quickest fix imo is to cache the value behind a lock any time you're on `main` and read that cached value any time you're on a non-main queue. This would mean that you'd have to call into PostHog from main at least once before you'd get screen data attached to events, but that seems like a reasonable requirement imo.

marandaneto: If the SDK is init in a background thread, and the main thread is still holding a lock, the problem would persist though.

btw @dfed `PostHogSDK.shared.isFeatureEnabled` is already thread-safe, unless your app has to do many operations within that lock, but reading/writing a flag is thread-safe.
I understand your use case, but feels like the deadlock is caused by your own locks/logic, the main thread should be available since most features require main thread access, including our SDK and some of its features.
I agree that we should not contribute to a deadlock hence trying to figure out the best approach here but I'd not hold a lock in the main thread.

@ioannisj it's important to return the correct screen size info if we do caching, or better to not attach any data at all, I'd prioritize correct data over skewed data.

What's about this [notification](https://developer.apple.com/documentation/uikit/uiwindow/1621607-didbecomekeynotification)? Would that be reliable to observe and cache the current active window?

dfed: > PostHogSDK.shared.isFeatureEnabled is already thread-safe, unless your app has to do many operations within that lock, but reading/writing a flag is thread-safe.
I understand your use case, but feels like the deadlock is caused by your own locks/logic, the main thread should be available since most features require main thread access, including our SDK and some of its features.
I agree that we should not contribute to a deadlock hence trying to figure out the best approach here but I'd not hold a lock in the main thread.

You're absolutely right that it's my code, not yours, that is causing the deadlock. However, what I'm doing should be supported by PostHog – PostHog should **not** be synchronously calling out to a global resource like `main`. Being able to ensure that PostHog always returns stable data is not possible without a lock or single-threading my calls into PostHog. Holding a lock on `main` is fine if the work is quick, and this work _should_ be quick.

> would it be possible for you to test your app against this https://github.com/PostHog/posthog-ios/pull/252 when you have a chance? (possibly also verifying that you get the correct $screen_width, $screen_height on your events? 🙏)

done! Looking good to me!

Issue #205, opened 2024-10-02
Adds referrerURL automatically


marandaneto:
### Description

Relates to https://github.com/PostHog/posthog-android/issues/138

Few ideas for iOS https://github.com/PostHog/posthog-android/issues/138#issuecomment-2329046161
Also here https://github.com/ueman/dart-packages/blob/7a28d92a9f6834de04aa8ef49f5e36cbdcca524a/referrer/ios/Classes/ReferrerPlugin.swift#L29

Prop $referrer, and $referring_domain

- [x] Android
- [x] iOS
- [ ] RN
- [ ] Flutter

joaoneves-bloop:
Would be great to have this on React Native. Also UTM properties.

jeremiahseun:
@marandaneto Can I pick this up? I'm asking because I want to be sure it's still needed.

PR #249, opened 2024-11-12
fix: allow set person properties after identify

ioannisj:

Relates to: https://github.com/PostHog/posthog-flutter/issues/120#issuecomment-2469771444

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 - fix: allow changing person properties after identify ([#249](https://github.com/PostHog/posthog-ios/pull/249))
 
+## 3.15.1 - 2024-11-12
+
+- fix: accessing UI APIs off main thread to get screen size ([#247](https://github.com/PostHog/posthog-ios/pull/247))
+
 ## 3.15.0 - 2024-11-11
 
 - add autocapture support for UIKit ([#224](https://github.com/PostHog/posthog-ios/pull/224))
--- a/PostHog.podspec
+++ b/PostHog.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name             = "PostHog"
-  s.version          = "3.15.0"
+  s.version          = "3.15.1"
   s.summary          = "The hassle-free way to add posthog to your iOS app."
 
   s.description      = <<-DESC
--- a/PostHog.xcodeproj/project.pbxproj
+++ b/PostHog.xcodeproj/project.pbxproj
@@ -120,6 +120,7 @@
 		69F518382BB2BA0100F52C14 /* PostHogSwizzler.swift in Sources */ = {isa = PBXBuildFile; fileRef = 69F518372BB2BA0100F52C14 /* PostHogSwizzler.swift */; };
 		69F5183A2BB2BA8300F52C14 /* UIApplicationTracker.swift in Sources */ = {isa = PBXBuildFile; fileRef = 69F518392BB2BA8300F52C14 /* UIApplicationTracker.swift */; };
 		DA26419C2CC0499300CB427B /* PostHogAutocaptureEventTracker.swift in Sources */ = {isa = PBXBuildFile; fileRef = DA26419A2CC0499300CB427B /* PostHogAutocaptureEventTracker.swift */; };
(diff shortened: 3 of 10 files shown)


PostHog/PostHogSDK.swift
```diff
@@ -490,6 +493,11 @@ let maxRetryDelay = 30.0
             if shouldReloadFlagsForTesting {
                 reloadFeatureFlags()
             }
+        } else if hasUserPropertyChanges {
+            capture("$set",
+                    distinctId: distinctId,
+                    userProperties: userProperties,
+                    userPropertiesSetOnce: userPropertiesSetOnce)
```
marandaneto, line 500:
```suggestion
            capture("$set",
                    distinctId: distinctId,
                    userProperties: userProperties,
                    userPropertiesSetOnce: userPropertiesSetOnce)
                    
            // Note we don't reload flags on property changes as these get processed async
```

PostHog/PostHogSDK.swift
```diff
@@ -468,7 +468,10 @@ let maxRetryDelay = 30.0
 
         let isIdentified = storageManager.isIdentified()
 
-        if distinctId != oldDistinctId, !isIdentified {
+        let hasDistinctIdChanges = distinctId != oldDistinctId && !isIdentified
+        let hasUserPropertyChanges = !(userProperties?.isEmpty ?? true) || !(userPropertiesSetOnce?.isEmpty ?? true)
```
marandaneto, line 472: the benefit of keeping this check below is that its only executed if the first if is false, right now we do the check even if not needed, its a lil perf improvement, but your call

PostHog/PostHogSDK.swift
```diff
@@ -468,7 +468,10 @@ let maxRetryDelay = 30.0
 
         let isIdentified = storageManager.isIdentified()
 
-        if distinctId != oldDistinctId, !isIdentified {
+        let hasDistinctIdChanges = distinctId != oldDistinctId && !isIdentified
```
marandaneto, line 471: since we also check the `isIdentified` flag, the naming `hasDistinctIdChanges` is semantically wrong, maybe is hasDistinctIdChanges=false` just because `isIdentified=true`.
`hasDistinctIdChangesOrAnonymous` / `hasDistinctIdChangesOrNonIdentified` would be better for the reader, or just keep the check inside the `if`.

PostHog/PostHogSDK.swift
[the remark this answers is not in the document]
marandaneto: true, looking right now, that makes sense.
I'd write a new test since we'd be asserting different properties anyway

PostHog/PostHogSDK.swift
```diff
@@ -490,6 +490,13 @@ let maxRetryDelay = 30.0
             if shouldReloadFlagsForTesting {
                 reloadFeatureFlags()
             }
+        } else if !(userProperties?.isEmpty ?? true) || !(userPropertiesSetOnce?.isEmpty ?? true) {
```
marandaneto, line 493: I just noticed we need an extra check here.
[PostHog/posthog-android@`ef2e552` (#205)](https://github.com/PostHog/posthog-android/pull/205/commits/ef2e552e9318ba9c4a0ae17bdfe8789986871578)
See that the $set event should only happen if its for the very same distinctId

marandaneto: The JS SDK instead of not calling capture(distinctId, event=$set) which is what we do here, doesn't use the given distinctId and uses the cached distinctId, I find that odd.

identify(userA, {a=b})
identify(userB, {a=c})

in our case, it won't work since the user is identified as userA, and they are trying to update user props from the userB.
In JS, they will update the user props {a=c} for userA.

So I think my approach is more defensive to user errors.

PostHog/PostHogSDK.swift
```diff
@@ -468,30 +468,44 @@ let maxRetryDelay = 30.0
...
-            let sanitizedProperties = sanitizeProperties(properties)
-
-            queue.add(PostHogEvent(
-                event: "$identify",
-                distinctId: distinctId,
-                properties: sanitizedProperties
-            ))
+        // if identified -> distinctId == oldDistinctId
+        // if !identified -> distinctId != oldDistinctId
+        if (distinctId == oldDistinctId) == isIdentified {
```
marandaneto, line 473: I find this approach harder to read that we even need a comment, I read 3 times and I am still unsure if it does the right thing :D but your call here.

marandaneto (approved):

PR #250, opened 2024-11-14
fix: use global frame on wireframe views

### LINKED ISSUE #231: Wireframe Session Recording appears mangled
### Version

3.13.3

### Steps to Reproduce

1. Added PostHog to my project
2. Enabled session recording with the default method (wireframe) and disabled masking Images.
3. Started app and Recorded some sessions 
4. Went to PostHog to view them and they're mangled.

The app is a UIKit-based app, so there shouldn't be any issues with wireframe. Maybe one of my libraries is causing the issue if it does something weird to the view hierarchy?

### Expected Result

A nice wireframe layout of my app's UI

### Actual Result

Screenshot 2024-11-05 at 9 48 22 AM

ioannisj:

Closes #231 

This issue was a bit challenging to uncover, but the solution turned out to be quite straightforward. After digging into the frontend code, I realized we always use `position: fixed` when styling nodes. To ensure the wireframe renders correctly, we need to convert all view frames to global screen coordinates.

### Before
before

### After
after

Locally with some sample projects

## :pencil: Checklist

### DIFF
--- a/PostHog/Replay/PostHogReplayIntegration.swift
+++ b/PostHog/Replay/PostHogReplayIntegration.swift
@@ -146,14 +146,19 @@
             style.paddingLeft = Int(insets.left)
         }
 
-        private func createBasicWireframe(_ window: UIView) -> RRWireframe {
+        private func createBasicWireframe(_ view: UIView) -> RRWireframe {
             let wireframe = RRWireframe()
 
-            wireframe.id = window.hash
-            wireframe.posX = Int(window.frame.origin.x)
-            wireframe.posY = Int(window.frame.origin.y)
-            wireframe.width = Int(window.frame.size.width)
-            wireframe.height = Int(window.frame.size.height)
+            // since FE will render each node of the wireframe with position: fixed
+            // we need to convert bounds to global screen coordinates
+            // otherwise each view of depth > 1 will likely have an origin of 0,0 (which is the local origin)
+            let frame = view.toAbsoluteRect(view.window)
+
+            wireframe.id = view.hash
+            wireframe.posX = Int(frame.origin.x)
+            wireframe.posY = Int(frame.origin.y)
+            wireframe.width = Int(frame.size.width)
+            wireframe.height = Int(frame.size.height)
 
             return wireframe
         }
@@ -380,8 +385,8 @@
             }
 
             let wireframe = createBasicWireframe(view)
-
             let style = RRStyle()
+            var skipChildren = false
 
             if let textView = view as? UITextView {
                 wireframe.type = "text"
@@ -447,8 +452,14 @@
                 wireframe.inputType = "button"
                 wireframe.disabled = !button.isEnabled
 
+                // UIButton has a UILabel subview that displays the button's text
+                // We want to skip visiting children if button is sensitive
+                // Otherwise, we capture the button's title through its child UILabel
                 if let text = button.titleLabel?.text {
-                    wireframe.value = isButtonSensitive(button) ? text.mask() : text
+                    if isButtonSensitive(button) {
+                        skipChildren = true
+                        wireframe.value = text.mask()
+                    }
                 }
             }
 
@@ -496,7 +507,7 @@
 
             wireframe.style = style
 
-            if !view.subviews.isEmpty {
+            if !view.subviews.isEmpty, !skipChildren {
                 var childWireframes: [RRWireframe] = []
                 for subview in view.subviews {
                     if let child = toWireframe(subview) {
--- a/PostHog/Replay/PostHogReplayIntegration.swift
+++ b/PostHog/Replay/PostHogReplayIntegration.swift
@@ -506,7 +506,7 @@
             if !view.subviews.isEmpty {
                 var childWireframes: [RRWireframe] = []
                 for subview in view.subviews {
-                    if let child = toWireframe(subview, isParentSensitive: isSensitive) {
+                    if let child = toWireframe(subview, isParentSensitive: isSensitive || isParentSensitive) {
                         childWireframes.append(child)
                     }
                 }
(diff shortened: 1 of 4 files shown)


PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -146,14 +146,19 @@
             style.paddingLeft = Int(insets.left)
         }
 
-        private func createBasicWireframe(_ window: UIView) -> RRWireframe {
+        private func createBasicWireframe(_ view: UIView) -> RRWireframe {
```
marandaneto, line 149: the first time `toWireframe` gets called, we always pass the `window`, so maybe this method should have 2 params, the view and the window.
`view.window` returns nullable which is likely never the case, but if it does, the positioning will be broken again, passing the `view` and `window`, we know both are never nullable, similar to `findMaskableWidgets`

PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -447,8 +452,14 @@
                 wireframe.inputType = "button"
                 wireframe.disabled = !button.isEnabled
 
+                // UIButton has a UILabel subview that displays the button's text
+                // We want to skip visiting children if button is sensitive
+                // Otherwise, we capture the button's title through its child UILabel
```
marandaneto, line 457: but then we will check if the label is `isLabelSensitive` and it will mask anyway, right? confused...

marandaneto: maybe a better approach would be to do what we do in the `findMaskableWidgets` and `maskChildren` flag, we still traverse the children, but respect the parent's masking.
the point is that the children may have more wireframes to be laid out and we will lose the details, eg if the `font` is applied to the UILable and not the Button

PostHog/Utils/UIApplication+.swift
```diff
@@ -26,6 +26,7 @@
                 } else {
                     // check scene.windows.isKeyWindow
                     for window in scene.windows {
+                        // swiftlint:disable:next for_where
```
marandaneto, line 29: We use `where` in some other places already so let's just use it instead of silencing the linter.

PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -374,18 +379,19 @@
...
-        private func toWireframe(_ view: UIView) -> RRWireframe? {
+        private func toWireframe(_ view: UIView, isParentSensitive: Bool = false) -> RRWireframe? {
             if !view.isVisible() {
                 return nil
             }
 
             let wireframe = createBasicWireframe(view)
-
             let style = RRStyle()
+            var isSensitive = false
```
marandaneto, line 389: I think this should be a local var within each if block otherwise a component that extends 2+ interfaces can swap the values, not sure if this will happen but on Android it'd, eg a Button is a `Button` and also a `TextView`, so if `isSensitive=true` and `isSensitive=false` in the 2nd case, you ignored the 1st result which is isSensitive=true

PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -499,7 +506,7 @@
             if !view.subviews.isEmpty {
                 var childWireframes: [RRWireframe] = []
                 for subview in view.subviews {
-                    if let child = toWireframe(subview) {
+                    if let child = toWireframe(subview, isParentSensitive: isSensitive) {
```
marandaneto, line 509: I think this should be passed as a reference, and cleared out after all subviews, see `findMaskableWidgets`.
Right now if:
Button -> sensitive -> isSensitive=true
Label -> not sensitive -> isSensitive=false
Something else -> not sensitive -> isSensitive=false but should have respected isSensitive=true from the Button parent

marandaneto: I can't tell if not running the code, I know I tested `findMaskableWidgets` passing a reference + cleaning up the instance after looping the children, and I know that works.

PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -423,39 +429,40 @@
...
                     }
                 }
             }
 
             if let imageView = view as? UIImageView {
+                isSensitive = isImageViewSensitive(imageView)
                 wireframe.type = "image"
                 if let image = imageView.image {
-                    if !isImageViewSensitive(imageView) {
+                    if !isSensitive, !isParentSensitive {
```
marandaneto, line 447: Not sure this is correct as `and`? since all others are `or`.

marandaneto: the other thing is that `isParentSensitive` is mostly if not all, about masking text, but in this case, the image will be or not masked based on that config too.

marandaneto: `findMaskableWidgets` only passes the `maskChildren` flag if the view is `view.isNoCapture()`, I don't pass `maskChildren` if it's sensitive in text or image.
Every time I find a view sensitive in text or image, I return early, So I don't run into any of those problems.
I know you chose to implement it in a slightly different way, but take another look in `findMaskableWidgets` and learn from it.

marandaneto: I pointed to the method implementation, plus gave a few ideas, but the current implementation is different.

marandaneto: discussed offline, since this is out of scope, we will follow up on another PR, and get this bug fix merged first.

marandaneto (approved):

PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -448,6 +453,8 @@
                 wireframe.disabled = !button.isEnabled
 
                 if let text = button.titleLabel?.text {
+                    // NOTE: this will create a ghosting effect since text will also be captured in child UILabel
```
marandaneto, line 456: will create an issue

marandaneto: https://github.com/PostHog/posthog-ios/issues/255

PR #252, opened 2024-11-15
feat: observe key window changes and cache screen size

ioannisj:

discussion: https://github.com/PostHog/posthog-ios/pull/247#discussion_r1842345311

React to the following notifications and cache current screen size: 

**macOS:** 
- NSWindow.didChangeScreenNotification
- NSWindow.didBecomeKeyNotification

**watchOS:** 
- WKApplication.didBecomeActiveNotification
- or "UIApplicationDidBecomeActiveNotification"

**iOS:** 
- UIDevice.orientationDidChangeNotification
- UIWindow.didBecomeKeyNotification

**tvOS:**
- UIWindow.didBecomeKeyNotification

Spent some time writing a unit test to detect deadlocks, but dropped it since it would hang the test execution in case of a deadlock (old code).
Tested on sample apps for WatchOS, iOS and macOS 

## :pencil: Checklist

### DIFF
--- a/PostHog/PostHogContext.swift
+++ b/PostHog/PostHogContext.swift
@@ -100,12 +100,22 @@ class PostHogContext {
         }
     #else
         init() {
-            registerNotifications()
+            if #available(watchOS 7.0, *) {
+                registerNotifications()
+            } else {
+                onShouldUpdateScreenSize()
+            }
         }
     #endif
 
     deinit {
-        unregisterNotifications()
+        #if !os(watchOS)
+            unregisterNotifications()
+        #else
+            if #available(watchOS 7.0, *) {
+                unregisterNotifications()
+            }
+        #endif
     }
 
     private lazy var theSdkInfo: [String: Any] = {
@@ -185,11 +195,6 @@ class PostHogContext {
                                                        selector: #selector(onShouldUpdateScreenSize),
                                                        name: WKApplication.didBecomeActiveNotification,
                                                        object: nil)
-            } else {
-                NotificationCenter.default.addObserver(self,
-                                                       selector: #selector(onShouldUpdateScreenSize),
-                                                       name: .init("UIApplicationDidBecomeActiveNotification"),
-                                                       object: nil)
             }
         #endif
     }
@@ -219,10 +224,6 @@ class PostHogContext {
             if #available(watchOS 7.0, *) {
                 NotificationCenter.default.removeObserver(self,
                                                           name: WKApplication.didBecomeActiveNotification,
-                                                          object: nil)
-            } else {
-                NotificationCenter.default.removeObserver(self,
-                                                          name: .init("UIApplicationDidBecomeActiveNotification"),
                                                           object: nil)
             }
         #endif
--- a/PostHog/PostHogSDK.swift
+++ b/PostHog/PostHogSDK.swift
@@ -1049,10 +1049,6 @@ let maxRetryDelay = 30.0
                 NotificationCenter.default.removeObserver(self,
                                                           name: WKApplication.didBecomeActiveNotification,
                                                           object: nil)
-            } else {
-                NotificationCenter.default.removeObserver(self,
-                                                          name: .init("UIApplicationDidBecomeActiveNotification"),
-                                                          object: nil)
             }
         #endif
     }


PostHog/PostHogContext.swift
```diff
@@ -161,4 +153,94 @@ class PostHogContext {
...
+                                                   object: nil)
+            NotificationCenter.default.addObserver(self,
+                                                   selector: #selector(cacheScreenSize),
+                                                   name: NSWindow.didChangeScreenNotification,
+                                                   object: nil)
+        #elseif os(watchOS)
+            if #available(watchOS 7.0, *) {
+                NotificationCenter.default.addObserver(self,
+                                                       selector: #selector(cacheScreenSize),
+                                                       name: WKApplication.didBecomeActiveNotification,
```
marandaneto, line 182: I don't think we have those in the `PostHogSDK#registerNotifications`
So likely missing those events in there as well, mind copying the missing ones?

PostHog/PostHogContext.swift
```diff
@@ -161,4 +153,94 @@ class PostHogContext {
...
+            #elseif os(watchOS)
+                return WKInterfaceDevice.current().screenBounds.size
+            #else
+                return nil
+            #endif
+        }()
+    }
+
+    @objc private func cacheScreenSize() {
+        // orientation change need a nudge on next run-loop
```
marandaneto, line 241: the notifications are not always delivered in the main thread already?

marandaneto: yeah that would be good, also if that's not the case, check `Thread.isMainThread` and only do `DispatchQueue.main.async` if needed, I think this is also good so the value is updated as soon as possible

PostHog/PostHogContext.swift
```diff
@@ -16,28 +16,13 @@ import Foundation
 #endif
 
 class PostHogContext {
+    @ReadWriteLock
```
marandaneto, line 19: I have been using `NSLock` for almost everything but the `PostHogFileBackedQueue` (came from the older version)
I intend to remove `ReadWriteLock` later since `NSLock` gives the ability of atomicity for multiple operations.
No need to change it now but keep that in mind.

PostHog/PostHogContext.swift
```diff
@@ -161,4 +153,124 @@ class PostHogContext {
...
+            if #available(watchOS 7.0, *) {
+                NotificationCenter.default.addObserver(self,
+                                                       selector: #selector(onShouldUpdateScreenSize),
+                                                       name: WKApplication.didBecomeActiveNotification,
+                                                       object: nil)
+            } else {
+                NotificationCenter.default.addObserver(self,
+                                                       selector: #selector(onShouldUpdateScreenSize),
+                                                       name: .init("UIApplicationDidBecomeActiveNotification"),
+                                                       object: nil)
```
marandaneto, line 192: is that available for watchos? https://developer.apple.com/documentation/uikit/uiapplicationdidbecomeactivenotification says its not

PostHog/PostHogContext.swift
```diff
@@ -161,4 +153,124 @@ class PostHogContext {
...
+        #endif
+    }
+
+    /// Retrieves the current screen size of the application window based on platform
+    private func getScreenSize() -> CGSize? {
+        #if os(iOS) || os(tvOS)
+            return UIApplication.getCurrentWindow(filterForegrounded: false)?.bounds.size
+        #elseif os(macOS)
+            // NSScreen.frame represents the full screen rectangle and includes any space occupied by menu, dock or camera bezel
+            return NSApplication.shared.windows.first { $0.isKeyWindow }?.screen?.frame.size
```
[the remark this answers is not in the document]
marandaneto: cool, for macos session replay we can double-check this because if it's wrong, the screen bound will be off since all positions are absolute

PostHog/PostHogContext.swift
```diff
@@ -161,4 +153,124 @@ class PostHogContext {
...
+    #if os(iOS)
+        // Special treatment for `orientationDidChangeNotification` since the notification seems to be _sometimes_ called early, before screen bounds are flipped
+        @objc private func onOrientationDidChange() {
+            updateScreenSize {
+                self.getScreenSize().map { size in
+                    // manually set width and height based on device orientation. (Needed for fast orientation changes)
+                    if UIDevice.current.orientation.isLandscape {
+                        CGSize(width: max(size.width, size.height), height: min(size.height, size.width))
+                    } else {
+                        CGSize(width: min(size.width, size.height), height: max(size.height, size.width))
```
marandaneto, line 254: why do we need the `max` and `min`?

marandaneto:
@ioannisj left a few questions but otherwise LGTM

PostHog/PostHogContext.swift
marandaneto: Surprised that would trigger but it's not documented and in the code is `UIKIT_EXTERN NSNotificationName const UIApplicationDidBecomeActiveNotification API_UNAVAILABLE(watchos) NS_SWIFT_NONISOLATED;` (API_UNAVAILABLE(watchos).
your call here, I'd just double check it's really fired because I'd be super surprised.

marandaneto (approved):

PR #256, opened 2024-11-19
fix: handle zero touch coordinates

### LINKED ISSUE #254: Filter out 0,0 touches
### Description

https://github.com/PostHog/posthog-ios/blob/052971b7ad6dd1d0a154922ad0a129c00e6efeb6/PostHog/Replay/UIApplicationTracker.swift#L68-L69

we've seen cases where x and y are zeroed, this is likely invalid and should be filtered out.

ioannisj:
Closes #254 

The likely cause of zeroed-out coordinates is that we were accessing UITouch asynchronously. I was able to replicate locally by adding a conditional breakpoint checking for zero location. Every touch event that hit the breakpoint had its associated `touch.view`, `touch.window` and `touch.gestureRecognizers` erased. 

I think the system may be releasing or erasing this info by the time we handle it, so I changed to collect the necessary info on main thread, and then process async. I kept a failsafe check before capturing in place just in case. 

Locally with sample projects

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Next
 
+- fix: reading screen size could sometimes lead to a deadlock ([#252](https://github.com/PostHog/posthog-ios/pull/252))
 - fix: avoid zero touch locations ([#256](https://github.com/PostHog/posthog-ios/pull/256))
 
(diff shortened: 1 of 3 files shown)


marandaneto (approved):
nice finding.

Issue #159, opened 2024-08-13
Network bodies for network logs


marandaneto:
### Description

This should go somewhere here https://github.com/PostHog/posthog-ios/blob/b0c6c1847228ce1fb78438be21cfe39cddc55254/PostHog/Replay/URLSessionInterceptor.swift#L108-L117

marandaneto:
Depends on https://github.com/PostHog/posthog-ios/issues/204

PR #260, opened 2024-11-21
feat: refactor network logs instrumentation

### LINKED ISSUE #204: Refactor network logs instrumentation for session replay
### Description

Relates to https://github.com/PostHog/posthog-js-lite/issues/276
https://github.com/PostHog/posthog-ios/pull/203 is a quick fix but instrumentation should be reworked so we instrument the datatask methods without a completion handler but we are still able to collect the right duration.
There are a few ways (and more complex) of doing this, such as instrumenting the delegate methods, wrapping the completion handlers, instrumenting the URLProtocol for complete control, etc.
My idea is to borrow the logic if possible from https://github.com/kasketis/netfox.
They do this in all possible ways and its a well known lib

ioannisj:

Closes #204 
Also some work on #159 

This is an initial implementation of using a custom URLProtocol to intercept network traffic. 

TODO: 
- Work on supporting custom URLSession and URLSessionConfiguration
- `PostHogHTTPProtocol` currently uses it's own instance for URLSession to carry out the network. This should not be the case
- Need to test how this behaves with various HTTP verbs (`POST`, `PUT` etc)
- Either introduce captureBody, captureHeaders in config or remove body and header capture from this PR and address these separately

Sample project with async/await and traditional URLSession

## :pencil: Checklist

### DIFF
--- a/PostHog/PostHogSwizzler.swift
+++ b/PostHog/PostHogSwizzler.swift
@@ -12,3 +12,10 @@ func swizzle(forClass: AnyClass, original: Selector, new: Selector) {
     guard let swizzledMethod = class_getInstanceMethod(forClass, new) else { return }
     method_exchangeImplementations(originalMethod, swizzledMethod)
 }
+
+func swizzleClassMethod(forClass: AnyClass, original: Selector, new: Selector) {
+    guard let c = object_getClass(forClass) else { return }
+    guard let originalMethod = class_getClassMethod(c, original) else { return }
+    guard let swizzledMethod = class_getClassMethod(c, new) else { return }
+    method_exchangeImplementations(originalMethod, swizzledMethod)
+}
--- a/PostHog/Replay/NetworkRequestSample.swift
+++ b/PostHog/Replay/NetworkRequestSample.swift
@@ -1,0 +1,257 @@
+//
+//  NetworkRequestSample.swift
+//  PostHog
+//
+//  Created by Yiannis Josephides on 11/11/2024.
+//
+
+#if os(iOS)
+    import Foundation
+
+    private let jsonRegex: String = "^application/.*json"
+    private let xmlRegex: String = "^application/.*xml"
+
+    class NetworkRequestSample: Identifiable {
+        enum ContentType: String {
+            case json
+            case xml
+            case html
+            case image
+            case other
+
+            init(contentType: String) {
+                switch contentType {
+                case _ where contentType.matches(jsonRegex):
+                    self = .json
+                case _ where contentType.matches(xmlRegex) || contentType == "text/xml":
+                    self = .xml
+                case "text/html":
+                    self = .html
+                case _ where contentType.hasPrefix("image/"):
+                    self = .image
+                default:
+                    self = .other
+                }
+            }
+        }
+
+        lazy var id = UUID().uuidString
+
+        var timestamp = getCurrentTimeMilliseconds()
+        var timeOrigin = getMonotonicTimeInMilliseconds()
+
+        var requestStartTime: UInt64?
+        var requestURL: URL?
+        var requestMethod: String?
+        var requestHeaders: [String: Any]?
+        var requestContentType: ContentType?
+        var requestContentTypeRaw: String?
+        var requestBodyStr: String?
+        var requestBodyLength: Int?
+
+        var responseError: String?
+        var responseData: NSMutableData?
+        var responseStatus: Int?
+        var responseContentType: ContentType?
+        var responseContentTypeRaw: String?
+        var responseStartTime: UInt64?
+        var responseEndTime: UInt64?
+        var responseHeaders: [String: Any]?
+        var responseBodyStr: String?
+        var responseBodyLength: Int?
+
+        var durationMs: UInt64?
+
+        var isProcessed: Bool = false
+
+        // called when a request starts loading
+        func start(request: URLRequest) {
+            requestStartTime = getMonotonicTimeInMilliseconds()
+            requestURL = request.url?.absoluteURL
+            requestMethod = request.httpMethod
+            requestHeaders = request.normalizedHeaderFields ?? [:]
+
+            // grab content-type. Keys normalized with .lowercase()
+            if let contentType = requestHeaders?["content-type"] as? String {
+                let contentType = contentType.components(separatedBy: ";")[0]
+                requestContentTypeRaw = contentType
+                requestContentType = ContentType(contentType: contentType)
+            }
+
+            // grab request body
+            if let requestData = request.httpBody ?? request.httpBodyStream?.consume() {
+                if let responseContentType, responseContentType == .image {
+                    // don't record response body for image types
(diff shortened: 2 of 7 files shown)


PostHog/PostHogSwizzler.swift
```diff
@@ -12,3 +12,10 @@ func swizzle(forClass: AnyClass, original: Selector, new: Selector) {
     guard let swizzledMethod = class_getInstanceMethod(forClass, new) else { return }
     method_exchangeImplementations(originalMethod, swizzledMethod)
 }
+
+func swizzleClassMethod(forClass: AnyClass, original: Selector, new: Selector) {
```
marandaneto, line 16: maybe rename the one above to make more clear its about instance method?

PostHog/Replay/NetworkRequestSample.swift
```diff
@@ -0,0 +1,257 @@
...
+                requestContentTypeRaw = contentType
+                requestContentType = ContentType(contentType: contentType)
+            }
+
+            // grab request body
+            if let requestData = request.httpBody ?? request.httpBodyStream?.consume() {
+                if let responseContentType, responseContentType == .image {
+                    // don't record response body for image types
+                    let bodyStr = requestData.base64EncodedString(options: .endLineWithLineFeed)
+                    requestBodyLength = bodyStr.count
```
marandaneto, line 86: body length in bytes is != than base64 encoded char count.
we can probably read `Content-Length` here or just count bytes from `requestData`

PostHog/Replay/NetworkRequestSample.swift
```diff
@@ -0,0 +1,257 @@
...
+
+            // grab content-type. Keys normalized with .lowercase()
+            if let contentType = requestHeaders?["content-type"] as? String {
+                let contentType = contentType.components(separatedBy: ";")[0]
+                requestContentTypeRaw = contentType
+                requestContentType = ContentType(contentType: contentType)
+            }
+
+            // grab request body
+            if let requestData = request.httpBody ?? request.httpBodyStream?.consume() {
```
marandaneto, line 82: since this has a performance impact + PII risk, we should add a new flag `captureBody` as part of session replay config

PostHog/Replay/NetworkRequestSample.swift
```diff
@@ -0,0 +1,257 @@
...
+            if let responseData = responseData as? Data {
+                if let responseContentType, responseContentType == .image {
+                    // don't record response body for image types
+                    let bodyStr = responseData.base64EncodedString(options: .endLineWithLineFeed)
+                    responseBodyLength = bodyStr.count
+                } else if let utfString = String(data: responseData, encoding: String.Encoding.utf8) {
+                    responseBodyStr = utfString
+                    responseBodyLength = utfString.count
+                }
+            }
```
marandaneto, line 124: not sure if 100% the same but we can extract from `start` and reuse it here? if `responseData` and `requestData` are of the same type

PostHog/Replay/NetworkRequestSample.swift
```diff
@@ -0,0 +1,257 @@
...
+            responseData?.append(data)
+        }
+
+        // sample was queued upstream - for debug purposes
+        func markProcessed() {
+            isProcessed = true
+        }
+    }
+
+    private func getMonotonicTimeInMilliseconds() -> UInt64 {
```
marandaneto, line 142: `getMonotonicTimeInMilliseconds` already exists, can we reuse it? See URLSession extension

PostHog/Replay/PostHogDefaultHTTPProtocol.swift
```diff
@@ -0,0 +1,281 @@
...
+        private var sessionDataTask: URLSessionDataTask?
+        private var currentSample = NetworkRequestSample()
+        private var response: URLResponse?
+
+        override init(request: URLRequest, cachedResponse: CachedURLResponse?, client: URLProtocolClient?) {
+            super.init(request: request, cachedResponse: cachedResponse, client: client)
+
+            if session == nil {
+                // ⚠️ - This is currently always using a fresh session with the `default` configuration
+                //    - Need to figure out a way to use custom sessions and configurations here
```
marandaneto, line 38: That's one of the reasons I gave up on URLProtocol back then since it'd require figuring out more stuff rather than inspiring from the current DD approach.
We need to be sure that React native works as well, I'd need to check if they use a custom or not.

PostHog/Replay/PostHogDefaultHTTPProtocol.swift
```diff
@@ -0,0 +1,281 @@
...
+        override public class func canInit(with request: URLRequest) -> Bool {
+            canHandle(request: request)
+        }
+
+        override class func canInit(with task: URLSessionTask) -> Bool {
+            canHandle(task: task)
+        }
+
+        private static func canHandle(task: URLSessionTask) -> Bool {
+            if #available(iOS 13.0, macOS 10.15, *) {
```
marandaneto, line 52: we dont need this I think
`        .macOS(.v10_15), .iOS(.v13), .tvOS(.v13), .watchOS(.v6),`
so the SDK is always running on >= than that

PostHog/Replay/NetworkRequestSample.swift
```diff
@@ -0,0 +1,257 @@
...
+
+            durationMs = relative(completedTime, to: requestStartTime)
+
+            if let responseData = responseData as? Data {
+                if let responseContentType, responseContentType == .image {
+                    // don't record response body for image types
+                    let bodyStr = responseData.base64EncodedString(options: .endLineWithLineFeed)
+                    responseBodyLength = bodyStr.count
+                } else if let utfString = String(data: responseData, encoding: String.Encoding.utf8) {
+                    responseBodyStr = utfString
```
marandaneto, line 121: also this https://posthog.com/docs/session-replay/network-recording#posthog-truncated-the-request-or-response-body

marandaneto: and this https://posthog.com/docs/session-replay/network-recording#a-network-request-says-it-was-redacted

marandaneto: and this https://posthog.com/docs/session-replay/network-recording#how-to-register-a-callback-to-inspect-and-redact-each-network-request
ok enough :D

PostHog/Replay/PostHogDefaultHTTPProtocol.swift
```diff
@@ -0,0 +1,281 @@
...
+
+            let data: [String: Any] = [
+                "type": 6,
+                "data": pluginData,
+                "timestamp": currentSample.timestamp,
+            ]
+
+            snapshotsData.append(data)
+
+            PostHogSDK.shared.capture("$snapshot", properties: [
```
marandaneto, line 123: we need to check `isCaptureNetworkEnabled` before processing a request and before capturing so we are sure we don't capture when there's no active session

PR #261, opened 2024-11-22
fix: detect and mask out photo library and user photos

ioannisj:

Best effort to detect and mask any UIImage which belongs to the user's photo library

We also try to detect system image pickers (and other remote views) and mask those out as well

Docs: https://github.com/PostHog/posthog.com/pull/9968

**Before**
CleanShot 2024-11-22 at 10 21 23

**After**
CleanShot 2024-11-22 at 10 19 52

Locally in UIKit sample project with an image picker view

## :pencil: Checklist

### DIFF
--- a/PostHog/Replay/PostHogReplayIntegration.swift
+++ b/PostHog/Replay/PostHogReplayIntegration.swift
@@ -287,7 +287,7 @@
             }
 
             // detect any views that don't belong to the current process (likely system views)
-            if config.sessionReplayConfig.maskSandboxedViews,
+            if config.sessionReplayConfig.maskAllSandboxedViews,
                let systemSandboxedView,
                view.isKind(of: systemSandboxedView)
             {


.swiftlint.yml
```diff
@@ -15,7 +15,10 @@ disabled_rules:
   - trailing_comma
   - opening_brace
 
-line_length: 160
+line_length: 
+  warning: 160
+  ignores_comments: true
```
marandaneto, line 20: nice

PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -429,14 +467,21 @@
...
+            guard let image = view.image else { return false }
+
+            // sensitive, regardless
+            if view.isNoCapture() {
+                return true
             }
-            return (config.sessionReplayConfig.maskAllImages && !isAsset) || view.isNoCapture()
+
+            if config.sessionReplayConfig.maskAllImages {
+                // asset images are probable not sensitive
```
marandaneto, line 479:
```suggestion
                // asset images are probably not sensitive
```

PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -429,14 +467,21 @@
...
-            if let image = view.image {
-                isAsset = isAssetsImage(image)
-            } else {
-                // if there's no image, there's nothing to mask
-                return false
+            // if there's no image, there's nothing to mask
+            guard let image = view.image else { return false }
+
+            // sensitive, regardless
+            if view.isNoCapture() {
```
marandaneto, line 474: this should be the case already in L353 right? we are doing 2 checks for the very same view.

PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -380,6 +404,20 @@
             image.imageAsset?.value(forKey: "_containingBundle") != nil
         }
 
+        // Photo library images have a UUID identifier as _assetName (e.g 64EF5A48-2E96-4AB2-A79B-AAB7E9116E3D)
+        // SF symbol and bundle images have the actual symbol name as _assetName (e.g chevron.backward)
```
marandaneto, line 408: I wonder if its the same for all OS versions, and if it works for SwiftUI and UIKit?

marandaneto: Well there's a similar approach in isAssetsImage so that's fine I guess

PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -380,6 +404,20 @@
...
+        private func isPhotoLibraryImage(_ image: UIImage) -> Bool {
+            guard let assetName = image.imageAsset?.value(forKey: "_assetName") as? String else {
+                return false
+            }
+
+            if assetName.isEmpty { return false }
+            if image.isSymbolImage { return false }
+            if isAssetsImage(image) { return false }
```
marandaneto, line 416: I think this method call should be an `or` operation with `isPhotoLibraryImage` since they are checking different things

PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -283,6 +286,27 @@
                 }
             }
 
+            // if view in photo library picker controller, always mask the whole view
+            if let pickerViewController = view.nearestViewController as? UIImagePickerController {
+                maskableWidgets.append(pickerViewController.view.toAbsoluteRect(window))
+                return
+            }
```
marandaneto, line 293: I think this should be masked by default, but we should offer a way out.
I do know a user that wanted to see the picture because the app was all about that.
Maybe a new flag config `maskNativeImagePicker` or something like that?

PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -283,6 +286,27 @@
...
+                }
+            }
+
+            // detect any views that don't belong to the current process (likely system views)
+            if let systemRemoteView {
+                if view.isKind(of: systemRemoteView) {
+                    maskableWidgets.append(view.toAbsoluteRect(window))
+                    return
+                }
+            }
```
marandaneto, line 308: do you have any examples of what a system view here is?

PostHog/Replay/PostHogReplayIntegration.swift
marandaneto: true that, maybe `view.isNoCapture()` should be the very first thing to be done.
they are on android https://github.com/PostHog/posthog-android/blob/00df23213f574344871c2e16cca3dc16e3aa2a41/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt#L542-L549
so we don't need to repeat them.

PostHog/Replay/PostHogReplayIntegration.swift
marandaneto: nvm, `isAssetsImage` is still checked next to `maskAllImages`, made the comment before reaching that part

PostHog/Replay/PostHogReplayIntegration.swift
marandaneto: if we can distinguish the album itself from the photo that the user has chosen or taken, I agree, if it's not possible, then it'd say it should be an opt-in feature (masked by default) otherwise we are just blocking this type of usage with no way out.

PostHog/Replay/PostHogReplayIntegration.swift
marandaneto: most likely depends on https://github.com/PostHog/posthog-ios/pull/261#discussion_r1853700510

PostHog/Replay/PostHogReplayIntegration.swift
marandaneto: eg: a photo app would need to see what's happening otherwise session replay isn't useful for this type of app, even if it's sensitive, they can always add this under their terms & conditions and make this "possible".

PostHog/Replay/PostHogSessionReplayConfig.swift
```diff
@@ -18,6 +18,17 @@
         /// Default: true
         @objc public var maskAllImages: Bool = true
 
+        /// Enable masking of all sandboxed system views
+        /// These may include UIImagePickerController, PHPickerViewController and CNContactPickerViewController
+        /// Experimental support
+        /// Default: true
+        @objc public var maskSandboxedViews: Bool = true
```
marandaneto, line 25:
```suggestion
        @objc public var maskAllSandboxedViews: Bool = true
```

just so it keeps the pattern from others.
we may introduce more granular approaches in the future such as a list of types

PostHog/Replay/PostHogSessionReplayConfig.swift
```diff
@@ -18,6 +18,17 @@
...
+
+        /// Enable masking of images that likely originated from user's photo library
+        /// Experimental support (UIKit only)
+        /// Default: true
+        @objc public var maskPhotoLibraryImages: Bool = true
```
marandaneto, line 30: how would we mask this for swiftui? if that by enacling `maskAllSandboxedViews `? if so, I'd add it to the description.

PR #265, opened 2024-11-27
feat(replay): skip snapshot during view controller transitions

ioannisj:

There are cases where masked views peek through during view controller transitions. This is because of two things: 
- the way we mask snapshot views. We look for **visible** views in the view hierarchy that are sensitive and then draw a masking rectangle on top of them
- the way view transitions work on iOS: During a transition, the original view is hidden, and a `snapshot()` view is added in a transition context and animated between start and end position (or animated interactively). 

A fix for this is to avoid capturing a snapshot between these transitions. This leads to slightly less appealing recordings, but with less chances of unintentionally revealing PII 

|before|after|
|-|-|
|Before|After|

Local sample

## :pencil: Checklist

### DIFF
--- a/PostHog/Replay/PostHogReplayIntegration.swift
+++ b/PostHog/Replay/PostHogReplayIntegration.swift
@@ -99,14 +99,22 @@
             }
         }
 
+        private func isNotFlutter() -> Bool {
+            postHogSdkName != "posthog-flutter"
+        }
+
         func start() {
             stopTimer()
-            DispatchQueue.main.async {
-                self.timer = Timer.scheduledTimer(withTimeInterval: self.config.sessionReplayConfig.debouncerDelay, repeats: true, block: { _ in
-                    self.snapshot()
-                })
+
+            // flutter captures snapshots, so we don't need to capture them here
+            if isNotFlutter() {
+                DispatchQueue.main.async {
+                    self.timer = Timer.scheduledTimer(withTimeInterval: self.config.sessionReplayConfig.debouncerDelay, repeats: true, block: { _ in
+                        self.snapshot()
+                    })
+                }
+                ViewLayoutTracker.swizzleLayoutSubviews()
             }
-            ViewLayoutTracker.swizzleLayoutSubviews()
 
             UIApplicationTracker.swizzleSendEvent()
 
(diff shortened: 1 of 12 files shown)


PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -386,6 +391,22 @@
...
+        }
+
+        // Recursively check if a view controller (or any of its children) is currently animating a transition
+        // During a transition, a UIViewControllerTransitionCoordinator is assigned to a UIViewController which controls the overall transition
+        private func isAnimatingTransition(_ viewController: UIViewController) -> Bool {
+            if let isAnimated = viewController.transitionCoordinator?.isAnimated {
+                return isAnimated
+            }
+            return viewController.children.contains(where: isAnimatingTransition)
+        }
```
marandaneto, line 408: I have two questions here: Since this loops all its children, can it be slow? Should we introduce a maximum depth?
If the view is deeply nested, it gives me the feeling that we can run into an issue similar to https://github.com/PostHog/posthog-ios/pull/259

The other one is, if there's a constant animated screen, either just a button or really anything, does that mean that we will never capture a screenshot?

PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -367,6 +367,11 @@
                 return nil
             }
 
+            // this will bail on interactive animations (e.g edge slide)
+            if isAnimatingTransition(window) {
+                return nil
+            }
```
marandaneto, line 373: could be just an `or` operation with the condition above.

marandaneto: If there's a transition happening, we will skip the screenshot and that is fine, do we capture a screenshot when the transition is over or do we have to wait for a few user interaction that triggers an UI update?
My question is if we are skipping a screenshot somehow, if that is the case, we could "delay" this screenshot if possible

PostHog/Replay/PostHogReplayIntegration.swift
marandaneto: cool, thanks for clarifying.
your call if we should check just the `rootViewController`, that can be a good start, and if we see its not enough, we can loop all its children on a next improvement, if only the `rootViewController` is enough for most cases.

marandaneto (approved):
NICE

PR #194, opened 2024-09-12
fix: missing import for url session extensions

marandaneto:
## :bulb: Motivation and Context
follow up https://github.com/PostHog/posthog-ios/pull/189

## :pencil: Checklist

- [x] I reviewed the submitted code.
- [ ] I added tests to verify the changes.
- [ ] I updated the docs if needed.
- [x] No breaking change or entry added to the changelog.

PR #269, opened 2024-12-03
fix: screen events and autocapture should ignore keyboard window

ioannisj:
Original thread: https://posthog.slack.com/archives/C0113360FFV/p1733238766585229
Ticket: https://posthoghelp.zendesk.com/agent/tickets/21072 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/24648)

### Problem

It seems that we were unintentionally capturing view controllers belonging to the keyboard window when sending `$screen` events. While this might not be a significant issue under normal circumstances, for applications that are message- or form-heavy, this behavior could result in inflated autocapture event volumes. 	

### Fix

- Updated logic to ignore any UIView or UIViewController that belongs to the keyboard window
- Added a similar logic got `$autocapture` events + some minor improvements

Sample project

## :pencil: Checklist

### DIFF
--- a/PostHog/Autocapture/PostHogAutocaptureEventTracker.swift
+++ b/PostHog/Autocapture/PostHogAutocaptureEventTracker.swift
@@ -205,6 +205,23 @@
             // first, call original method
             ph_swizzled_setContentOffset_Setter(contentOffset)
 
+            guard shouldTrack(self) else {
+                return
+            }
+
+            // ignore all keyboard events
+            if let window, window.isKeyboardWindow {
+                return
+            }
+
+            // nothing changed
+            guard
+                self.contentOffset.x.distance(to: contentOffset.x) > 0 ||
+                self.contentOffset.y.distance(to: contentOffset.y) > 0
+            else {
+                return
+            }
+
             // block scrolls on UIPickerTableView. (captured via a forwarding delegate implementation)
             if String(describing: type(of: self)) == "UIPickerTableView" {
                 return
@@ -476,6 +493,7 @@
         if view.isHidden { return false }
         if !view.isUserInteractionEnabled { return false }
         if view.isNoCapture() { return false }
+        if view.window?.isKeyboardWindow == true { return false }
 
         if let textField = view as? UITextField, textField.isSensitiveText() {
             return false
--- a/PostHog/Autocapture/PostHogAutocaptureEventTracker.swift
+++ b/PostHog/Autocapture/PostHogAutocaptureEventTracker.swift
@@ -201,9 +201,9 @@
     }
 
     extension UIScrollView {
-        @objc func ph_swizzled_setContentOffset_Setter(_ contentOffset: CGPoint) {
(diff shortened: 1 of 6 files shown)


PostHog/Autocapture/PostHogAutocaptureEventTracker.swift
```diff
@@ -205,6 +205,23 @@
...
+
+            // ignore all keyboard events
+            if let window, window.isKeyboardWindow {
+                return
+            }
+
+            // nothing changed
+            guard
+                self.contentOffset.x.distance(to: contentOffset.x) > 0 ||
+                self.contentOffset.y.distance(to: contentOffset.y) > 0
```
marandaneto, line 220: is this a workaround for detecting if the keyboard is open or closed?

PostHog/Utils/UIWindow+.swift
```diff
@@ -0,0 +1,16 @@
+//
+//  UIWindow+.swift
+//  PostHog
+//
+//  Created by Yiannis Josephides on 03/12/2024.
+//
+
+#if os(iOS) || os(tvOS)
+    import UIKit
```
marandaneto, line 9: do we need to import `Foundation`? i recall having an issue once if not doing this despite compiling

PostHog/UIViewController.swift
```diff
@@ -58,11 +58,17 @@
...
+            // these may include: UIInputWindowController, _UICursorAccessoryViewController, UICompatibilityInputViewController,UIKeyboardHiddenViewController_Autofill and others
+            if let window = viewIfLoaded?.window, window.isKeyboardWindow {
+                return
+            }
+
             captureScreenView(viewIfLoaded?.window)
             // it looks like we're calling ourselves, but we're actually
             // calling the original implementation of viewDidAppear since it's been swizzled.
-            viewDidApperOverride(animated: animated)
+            viewDidAppearOverride(animated: animated)
```
marandaneto, line 71: I wonder if we should not capture `captureScreenView` here but still forward the call to `viewDidAppearOverride`.
Feels like the right thing to do otherwise we are breaking an OS/apps behaviour

PostHog/Autocapture/PostHogAutocaptureEventTracker.swift
marandaneto: please describe with a comment in the code since it was not obvious to me only with the "nothing changed"

PostHog/UIViewController.swift
marandaneto: we can also do `let window = viewIfLoaded?.window` so we dont do the `?.` twice in the same object when calling `captureScreenView`

PostHog/Utils/UIWindow+.swift
marandaneto: found the issue https://github.com/PostHog/posthog-ios/pull/194/files specifically URLSessionExtension
If `UIKit` knows about String, Bool and type(of...), should be fine, your call here.

marandaneto (approved):

PR #271, opened 2024-12-05
feat: add support for ph-label in autocapture elements

ioannisj:

Private slack thread: https://posthog.slack.com/archives/C03PB072FMJ/p1733219779409259
Docs: https://github.com/PostHog/posthog.com/pull/10085

Autocapture elements can now be given a PostHog label which can then be used in analytics to filter for a specific element (using Tag Name and/or CSS Selector)

Autocapture elements now get `class` and `id` attributes: 
- `class`: the superclass of the element if not a common type like `UIView`, `UIControl`, `NSObject` or `UIResponder`
- `id`: the value of the custom label defined through `postHogLabel` (UIKit) or `.postHogLabel(_ label: String?)` (SwiftUI)

## SwiftUI
- I think most of the interactions in a SwiftUI app are already autocaptured (since most of interactive SwiftUI controls map to a UIKit counterpart). List includes collection views, scroll views, table views, text fields, text views, switches etc
- An important element we are missing is support for `UIButton` (SwiftUI renders buttons as images)

## Minor
- Some minor improvements/corrections
- New simpler implementation of `ForwardingPickerViewDelegate`

## Notes

The implementation uses `$elements_chain` (constructed by the client) since `$elements` will be deprecated soon https://github.com/PostHog/posthog/blob/f9be4fa28668e242d56bab9651cf9c79b40d9249/plugin-server/src/worker/ingestion/process-event.ts#L119

## :pencil: Checklist

### DIFF
--- a/PostHog/Autocapture/PostHogAutocaptureEventTracker.swift
+++ b/PostHog/Autocapture/PostHogAutocaptureEventTracker.swift
@@ -207,9 +207,23 @@
     }
 
     extension UIScrollView {
-        @objc func ph_swizzled_setContentOffset_Setter(_ contentOffset: CGPoint) {
+        @objc func ph_swizzled_setContentOffset_Setter(_ newContentOffset: CGPoint) {
             // first, call original method
-            ph_swizzled_setContentOffset_Setter(contentOffset)
+            ph_swizzled_setContentOffset_Setter(newContentOffset)
+
+            guard shouldTrack(self) else {
+                return
+            }
+
+            // ignore all keyboard events
+            if let window, window.isKeyboardWindow {
+                return
+            }
+
+            // scrollview did not scroll (contentOffset didn't change)
+            guard contentOffset != newContentOffset else {
+                return
+            }
 
             // block scrolls on UIPickerTableView. (captured via a forwarding delegate implementation)
             if String(describing: type(of: self)) == "UIPickerTableView" {
@@ -510,6 +524,7 @@
         if view.isHidden { return false }
         if !view.isUserInteractionEnabled { return false }
         if view.isNoCapture() { return false }
+        if view.window?.isKeyboardWindow == true { return false }
 
         if let textField = view as? UITextField, textField.isSensitiveText() {
             return false
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 - feat: ability to add a custom label to autocapture elements ([#271](https://github.com/PostHog/posthog-ios/pull/271))
 
+## 3.16.2 - 2024-12-05
+
+- fix: ignore autocapture events from keyboard window ([#269](https://github.com/PostHog/posthog-ios/pull/269))
+
 ## 3.16.1 - 2024-12-04
 
 - fix: screen flicker when capturing a screenshot when a sensitive text field is on screen ([#270](https://github.com/PostHog/posthog-ios/pull/270))
--- a/PostHog.podspec
+++ b/PostHog.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name             = "PostHog"
-  s.version          = "3.16.1"
+  s.version          = "3.16.2"
   s.summary          = "The hassle-free way to add posthog to your iOS app."
 
   s.description      = <<-DESC
--- a/PostHog.xcodeproj/project.pbxproj
+++ b/PostHog.xcodeproj/project.pbxproj
@@ -120,6 +120,7 @@
 		69F5181A2BAC81FC00F52C14 /* UITextInputTraits+Util.swift in Sources */ = {isa = PBXBuildFile; fileRef = 69F518192BAC81FC00F52C14 /* UITextInputTraits+Util.swift */; };
 		69F518382BB2BA0100F52C14 /* PostHogSwizzler.swift in Sources */ = {isa = PBXBuildFile; fileRef = 69F518372BB2BA0100F52C14 /* PostHogSwizzler.swift */; };
 		69F5183A2BB2BA8300F52C14 /* UIApplicationTracker.swift in Sources */ = {isa = PBXBuildFile; fileRef = 69F518392BB2BA8300F52C14 /* UIApplicationTracker.swift */; };
+		DA0CA6F12CFF6B6300AF9500 /* UIWindow+.swift in Sources */ = {isa = PBXBuildFile; fileRef = DA0CA6F02CFF6B6300AF9500 /* UIWindow+.swift */; };
 		DA26419C2CC0499300CB427B /* PostHogAutocaptureEventTracker.swift in Sources */ = {isa = PBXBuildFile; fileRef = DA26419A2CC0499300CB427B /* PostHogAutocaptureEventTracker.swift */; };
 		DA4932FD2D0102950092C213 /* AssociatedKeys.swift in Sources */ = {isa = PBXBuildFile; fileRef = DA4932FC2D0102910092C213 /* AssociatedKeys.swift */; };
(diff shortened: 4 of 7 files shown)


PostHog/Autocapture/ForwardingPickerViewDelegate.swift
```diff
@@ -8,953 +6,66 @@
...
-
-        func pickerView(_ pickerView: UIPickerView, rowHeightForComponent component: Int) -> CGFloat {
-            actualDelegate?.pickerView?(pickerView, rowHeightForComponent: component) ?? .zero
-        }
-    }
-
-    /// Combination 6: `didSelectRow`, `widthForComponent`
-    private class ForwardingPickerViewDelegate6: NSObject, UIPickerViewDelegate {
+    final class ForwardingPickerViewDelegate: NSObject, UIPickerViewDelegate, UIPickerViewDataSource {
+        // this needs to be week since `actualDelegate` will hold a strong reference to `ForwardingPickerViewDelegate`
```
marandaneto, line 12:
```suggestion
        // this needs to be weak since `actualDelegate` will hold a strong reference to `ForwardingPickerViewDelegate`
```

PostHog/Autocapture/PostHogAutocaptureEventTracker.swift
```diff
@@ -492,7 +524,7 @@
         return true
     }
 
-    // TODO: Filter out or obfuscsate strings that look like sensitive data
+    // TODO: Filter out or obfuscate strings that look like sensitive data
```
marandaneto, line 527: an alternative for now is to use the `propertiesSanitizer`, you can mention this in the docs, but its a manoel (manual) job :D

PostHog/Autocapture/SwiftUI/View+PostHogLabel.swift
```diff
@@ -0,0 +1,141 @@
...
+        private let label: String?
+
+        init(label: String?) {
+            self.label = label
+            super.init(frame: .zero)
+        }
+
+        @available(*, unavailable)
+        required init?(coder _: NSCoder) {
+            fatalError("init(coder:) has not been implemented")
```
marandaneto, line 67: Its ok to call hedgeHog but not crash the app here, or, never crash the app in any circumcircle

PostHog/Autocapture/SwiftUI/View+PostHogLabel.swift
```diff
@@ -0,0 +1,141 @@
...
+        }
+
+        override func layoutSubviews() {
+            super.didMoveToWindow()
+
+            superview?.postHogNoCapture = true
+            postHogNoCapture = true
```
marandaneto, line 74: I am a bit confused, why are we doing this here if this modifier is about the view's id?

PostHog/Autocapture/UIView+PostHogLabel.swift
```diff
@@ -0,0 +1,29 @@
...
+
+         By setting a custom label, you can easily identify and filter interactions with this specific element in your analytics data.
+
+         ### Usage
+         ```swift
+         let myView = UIView()
+         myView.postHogLabel = "customLabel"
+         ```
+         */
+        var postHogLabel: String? {
```
marandaneto, line 23: I think a follow up PR would be to read the `postHogLabel` in session recording and if its `ph-no-capture`, we mask it, or adding a new label for it, so we have another option instead of accessibility labels, create an issue?

PostHog/PostHogSDK.swift
```diff
@@ -678,7 +678,6 @@ let maxRetryDelay = 30.0
 
     func autocapture(
         eventType: String,
-        elements: [[String: Any]],
```
marandaneto, line 681: We should not break public APIs, or wasn't it public?

PostHog/Utils/DictUtils.swift
```diff
@@ -7,7 +7,7 @@
 
 import Foundation
 
-public func sanitizeDicionary(_ dict: [String: Any]?) -> [String: Any]? {
+public func sanitizeDictionary(_ dict: [String: Any]?) -> [String: Any]? {
```
marandaneto, line 10: Good catch

PostHog/Autocapture/PostHogAutocaptureEventTracker.swift
```diff
@@ -363,9 +369,35 @@
...
+            while let superclass = previousType.superclass() {
+                // NSObject or UIResponder are base types for almost all UIKit and not really helpful here
+                if superclass == NSObject.self || superclass == UIResponder.self {
+                    return baseTypes.joined(separator: " ")
+                }
+                previousType = superclass
+                baseTypes.append(typeName(of: superclass))
+            }
+
+            return baseTypes.joined(separator: " ")
```
marandaneto, line 400: Current example:

`<UITextField id="usernameTextField" class="UIControl UIView">I can be labeled</UITextField>`

I'm not convinced the `class` with nested hierarchy is useful. I just can't think of a useful example, do we use that at PostHog or can you think of an insight that would require the hierarchy tree? Maybe just the parent's class is fine if it's not the same type eg:

 `<UITextField id="usernameTextField" class="UIControl">I can be labeled</UITextField>`

But having the duplicated "UIView" and friends just seems wasting bytes over the network, happy to be wrong here, just need to think if its useful

PostHog/Autocapture/SwiftUI/View+PostHogLabel.swift
```diff
@@ -0,0 +1,141 @@
...
+//  View+PostHogLabel.swift
+//  PostHog
+//
+//  Created by Yiannis Josephides on 04/12/2024.
+//
+
+#if os(iOS) || targetEnvironment(macCatalyst)
+    import SwiftUI
+
+    public extension View {
```
marandaneto, line 11: can we combine modifiers? eg I want to add a label to a button but I don't want it to be captured in session recording

.swiftlint.yml
```diff
@@ -30,6 +30,8 @@ file_length:
 identifier_name:
   excluded:
     - id
+    - c
+    - r
```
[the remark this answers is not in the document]
marandaneto: Tbh I am not a big fan of singe char names because it's not obvious when you read them.
If you work with more languages makes it even harder.
https://developer.apple.com/documentation/swift/array/reduce(_:_:)
they use `x, y`, but IMO the best would be `initial` `nextPartial` or something like that, so its obvious from a first glance

PostHog/Autocapture/PostHogAutocaptureEventTracker.swift
marandaneto: yeah just superclass would be fine, I think the web makes sense to be more chatty because its everything a `div` so they need more info, it's different for us.
Maybe add a maxDepth for super classes as well so we don't have super long nested base classes, and gain some perf improvements in there by traversing the list a bit less

PostHog/Autocapture/SwiftUI/View+PostHogLabel.swift
```diff
@@ -0,0 +1,141 @@
...
+                    ...
+                }
+                .postHogLabel("loginButton")
+              }
+         }
+         ```
+
+         - Parameter label: A custom label that uniquely identifies the element for analytics purposes.
+         */
+        func postHogLabel(_ label: String?) -> some View {
```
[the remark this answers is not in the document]
marandaneto: oh yeah we discussed that before, my bad!
I just wonder if in that case, if the label is null, then we don't even need to create a `PostHogLabelTaggerViewModifier`, so a bit of a perf improvement by creating/wrapping less objects

marandaneto (approved):

PR #273, opened 2024-12-12
feat: change screenshot image format from Jpeg to WebP

ioannisj:

Motivation is to reduce the size of our replay snapshots by switching to a more compressed image format. 
From empirical testing on our sample apps, WebP is around ~40% more compressed than JPEG. 

**Why WebP?**
- WebP is natively supported on Android 
- WebP is supported by all modern browsers. [~97%](https://caniuse.com/webp) of browsers support WebP
- We want to maintain the same image format across all SDKs so we can better monitor the impact of the switch. 

**Alternative formats**

- AVIF was also a good alternative since it's the successor of WebP. 
   - Browser [support](https://caniuse.com/avif) for it is good, but less than WebP. 
   - We may consider adding support for it in the future but for now we'll stick with WebP for reasons outlined above. 

**WebP on iOS:**

Unfortunately, there's no native support for _encoding_ WebP images in iOS. (iOS added support for decoding in iOS 14). 

To add support we vendored in [libwebp](https://github.com/webmproject/libwebp/) which is the standard lib for WebP encoding. 

We had to change our package definitions both for Cocopods and SPM, which differ a bit because: 

- SPM does not support mix-language packages, 
    - We had to create a separate private module (libwebp) and import it in our main target.
    - SPM manages all bridging code between Swift and C 
    - SPM builds use a conditional `import libwebp` 
- Pods, on the other hand, don't support nested frameworks (`use_frameworks!` is standard), or at least I couldn't get it to work. 
    - So for Cocopods we had to package as a mix-language framework. 
    - This means that we needed to add some module maps and umbrella headers to get C-interop to work
    
> Note: There are some changes in project files as well since I needed to add some bridging code.   

> Note: this is a big PR mainly due to vendored code. Once I remove decoding/animated webp bits it could be more manageable but still a bit of a pain to review. 🙏

Checklist: 

- [x] Make sure SPM and Cocopods packages are functional
   - [x] `swift package describe`
   - [x] `pod spec lint PostHog.podspec --allow-warnings`
- [x] Make sure that all sample apps compile
- [x] Make sure we maintain permissive licensing info from libwebp
- [x] Cleanup decoding related vendored code
- [x] Cleanup animated webp related vendored code

@marandaneto I'll need your help testing this out, especially verifying that we didn't break package builds for people. 

## :pencil: Checklist

### DIFF
--- a/PostHog.xcodeproj/project.pbxproj
+++ b/PostHog.xcodeproj/project.pbxproj
@@ -1112,7 +1112,6 @@
 			path = SwiftUI;
 			sourceTree = "<group>";
 		};
-		DAD76A222D006BF7003E1A43 /* SwiftUI */ = {
 		DAAC94DA2D0B0455006A9B6E /* sharpyuv */ = {
 			isa = PBXGroup;
 			children = (


PostHog/Replay/UIImage+Util.swift
```diff
@@ -11,14 +11,29 @@
...
+            toWebPBase64(compressionQuality) ?? toJpegBase64(compressionQuality)
+        }
 
-            if let base64 = base64 {
-                return "data:image/jpeg;base64,\(base64)"
+        private func toWebPBase64(_ compressionQuality: CGFloat) -> String? {
+            webpData(
+                compressionQuality: compressionQuality,
+                options: [
+                    .alphaQuality(0), // lowest (smallest size)
```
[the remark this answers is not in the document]
marandaneto: i think we should match the config on android/ios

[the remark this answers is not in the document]
marandaneto: The webp config has;

/// Lossless encoding (0=lossy(default), 1=lossless).
        case lossless(Bool)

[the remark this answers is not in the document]
marandaneto: we also use 30% quality during compression

PostHog/Replay/UIImage+Util.swift
```diff
@@ -11,14 +11,29 @@
...
     extension UIImage {
         func toBase64(_ compressionQuality: CGFloat = 0.3) -> String? {
-            let jpegData = jpegData(compressionQuality: compressionQuality)
-            let base64 = jpegData?.base64EncodedString()
+            toWebPBase64(compressionQuality) ?? toJpegBase64(compressionQuality)
+        }
 
-            if let base64 = base64 {
-                return "data:image/jpeg;base64,\(base64)"
+        private func toWebPBase64(_ compressionQuality: CGFloat) -> String? {
```
marandaneto, line 17: RN and Flutter will need to use this so it has to be a public method, unless we do the change via `imageToBase64` method

PostHog/Utils/UIImage+WebP.swift
```diff
@@ -0,0 +1,277 @@
...
+         - Returns: A data object containing the WebP data, or nil if there’s a problem generating the data.
+         */
+        func webpData(compressionQuality: CGFloat, options: [WebPOption] = []) -> Data? {
+            // Early exit if image is missing
+            guard let cgImage = cgImage else {
+                return nil
+            }
+
+            // clamp compressionQuality to the valid range [0.0, 1.0]
+            let compressionQuality = min(max(compressionQuality, 0.0), 1.0)
```
marandaneto, line 35: we call this method with 30% hardcoded so no need to validate?

PostHog/Utils/UIImage+WebP.swift
```diff
@@ -0,0 +1,277 @@
...
+
+            // validate dimensions
+            let width = Int(cgImage.width)
+            let height = Int(cgImage.height)
+
+            guard width > 0, width <= WEBP_MAX_DIMENSION, height > 0, height <= WEBP_MAX_DIMENSION else {
+                return nil
+            }
```
marandaneto, line 43: where does `WEBP_MAX_DIMENSION` come from?

vendor/libwebp/src/dsp/dec.c
```diff
@@ -0,0 +1,894 @@
...
+  }
+
+#if defined(WEBP_HAVE_NEON)
+  if (WEBP_NEON_OMIT_C_CODE ||
+      (VP8GetCPUInfo != NULL && VP8GetCPUInfo(kNEON))) {
+    VP8DspInitNEON();
+  }
+#endif
+
+  assert(VP8TransformWHT != NULL);
```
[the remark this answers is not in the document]
marandaneto: If this can be enabled with an extra config that will crash people's apps in production, yes, I'd remove, lo,g and return early

PostHog/Replay/UIImage+Util.swift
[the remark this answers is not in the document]
marandaneto: This is the only setting I do everything else should be the default, so as long as we do 30% image quality, lossy, they should match, I don't have control over any other setting.

PostHog/Replay/UIImage+Util.swift
marandaneto: all good then

PostHog/Utils/UIImage+WebP.swift
marandaneto: TIL webp has a max dimension

marandaneto:
> @marandaneto can you try `make buildExamples` on your machine? Compiles okay on mine but fails in CI for some reason

works for me but I see this error in CI:

❌  /Users/runner/work/posthog-ios/posthog-ios/vendor/libwebp/include/types.h:20:10: include of non-modular header inside framework module 'PostHog.types': '/Applications/Xcode_15.2.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/15.0.0/include/inttypes.h' [-Werror,-Wnon-modular-include-in-framework-module]

#include <inttypes.h>

Check the CI `runs-on: macos-13` and `xcode-version: latest-stable`
Maybe its running on an older/newer versions where `inttypes.h` isn't available? (clang version etc), we need to be sure that webp works on iOS >= 13 btw

PostHog/Replay/UIImage+Util.swift
```diff
@@ -11,14 +11,19 @@
 
     extension UIImage {
         func toBase64(_ compressionQuality: CGFloat = 0.3) -> String? {
-            let jpegData = jpegData(compressionQuality: compressionQuality)
-            let base64 = jpegData?.base64EncodedString()
+            toWebPBase64(compressionQuality) ?? toJpegBase64(compressionQuality)
```
marandaneto, line 14: will `toWebPBase64` return nil here if it fails or still returns the string `data:image/webp;base64,` without the encoded image? otherwise the fall back will never work

marandaneto: ah ok, missed the webpData bits here.

marandaneto (approved):
once we make CI happy, we can merge this

Issue #227, opened 2024-11-01
Session Id didn't rotate after 23 hours


ioannisj:
### Description

A user reported that a session recording lasted for around 24 hours. After taking a look at the recording, we do indeed capture a few snapshots at the very end of the recording (timestamps look correct), right before the session is rotated. 

This is weird cause we clear the session_id on every capture [here](https://github.com/PostHog/posthog-ios/blob/4524c42d2b70a9e7ebffb913e7a2ab60ce00ecd1/PostHog/PostHogSDK.swift#L549) and rotate it after about 30 mins when the app is foregrounded [here](https://github.com/PostHog/posthog-ios/blob/4524c42d2b70a9e7ebffb913e7a2ab60ce00ecd1/PostHog/PostHogSDK.swift#L1046). 

Also, session id/window id is managed only in memory, so it's very unlikely that the OS suspended the app but didn't terminate it for 23 hours. 

From: https://posthoghelp.zendesk.com/agent/tickets/19974

marandaneto:
> @marandaneto any ideas on what could be going on here? Especially last part of my comment above - leads me to think that these could be events in the queue with the wrong timestamp somehow

I think the problem is that the `resetSessionIfExpired` isn't considering the timestamp of the snapshot event which is only generated when `PostHogEvent` instance is created.
In this case, if its possible to know that the session has expired and the session has more than 24h, the event snapshot event should be even dropped and not captured without a session_id.
For example, JS has the function `checkAndGetSessionAndWindowId` and other ones that do a lot of checking, I don't think we need to do everything the same but we are probably missing something.
Strange is that the snapshot had a session_id right, and that I don't know how its possible without further investigation.

marandaneto:
Another point is that we capture a `"$snapshot"` event in a background thread (eg PostHogReplayIntegration.dispatchQueue), but the event timestamp is only created when its executed, I think the event.timestamp should be passed to the capture method along with the `"$snapshot"` event so we compare the time that the event actually happened instead of when the event was captured (when the queue wasn't paused anymore or something).

marandaneto:
https://github.com/PostHog/posthog-ios/pull/228
I think we should pass `timestamp` here as described [here](https://github.com/PostHog/posthog-ios/issues/227#issuecomment-2453495539)

marandaneto:
https://github.com/PostHog/posthog-js/blob/a6655acf6910ecea144758b89358311797604941/src/sessionid.ts#L15
[...]

PR #276, opened 2024-12-19
feat: manual start/stop session recording

### LINKED ISSUE #262: Manually start and stop session recordings
### Description

Ticket: https://posthoghelp.zendesk.com/agent/tickets/20809

Add APIs to allow for manual start/stop of session recordings

- [x] [iOS](https://github.com/PostHog/posthog-ios/issues/262)
- [x] [Android](https://github.com/PostHog/posthog-android/issues/210)
- [ ] [RN](https://github.com/PostHog/posthog-js/issues/2169)
- [ ] [Flutter](https://github.com/PostHog/posthog-flutter/issues/124)

https://github.com/PostHog/posthog-js/blob/1b57ea9f6b7736f3e3fce0fd512b467ddb35c5d9/src/posthog-core.ts#L1837

This option is an alternative for:
- [Sampling](https://github.com/PostHog/posthog-android/issues/188)
- [Minimal duration](https://github.com/PostHog/posthog-android/issues/189)
- [Screen by screen basis](https://github.com/PostHog/posthog-ios/issues/160)

ioannisj:

Closes #262, #227 

Docs PR: https://github.com/PostHog/posthog.com/pull/10235

Added the following:
- `startSessionRecording(resumeCurrent: Bool)`
- `stopSessionRecording()`

### Session Management

Related discussion: https://posthog.slack.com/archives/C03PB072FMJ/p1733146678787689?thread_ts=1732810962.491699&cid=C03PB072FMJ

Based on the discussion above, I refactored how we manager and rotate replay sessions. Quite a big change but the main points of the change are: 

- Internal `getSessionId` changes form and is now responsible on running validation rules and rotates or clears the session if needed before returning
- `getSessionId` can be called with `readOnly` to skip validation and just return current session id. In both cases, a new id will be created if it doesn't exist but only when the app is foregrounded 
- Validation rules are: 
    - Rotates session after *30 minutes* of inactivity
    - Clears session after *30 minutes* of inactivity (when app is backgrounded at the time of calling this method)
    - Enforces a maximum session duration of *24 hours*
    - All validations can be run against a given timestamp, otherwise `.now` is used. For replay, we capture the timestamp of the event as soon as possible, so we need to run session validation checks on that timestamp to account for delayed captures
- `$snapshot` events add their own `$session_id` property which is acquired as early as possible and `capture()` method will respect that $session_id 
- `$snapshot` events always call `capture()` with a given timestamp, acquired as early as possible
- Added a `reason` on hedgeLog messages related to session id changes for debugging purposes

## TODO: 
- [x] Skip $snapshot events if $session_id is missing from event properties

- Unit tests
- Added a section in PostHogExample for manual testing
- Objc project interface

## :pencil: Checklist

### DIFF
--- a/.github/workflows/build-examples.yml
+++ b/.github/workflows/build-examples.yml
@@ -8,7 +8,7 @@ on:
       - "**/*.md"
 jobs:
   build:
-    runs-on: macos-13
+    runs-on: macos-14
     steps:
       - uses: actions/checkout@v4
       - uses: maxim-lobanov/setup-xcode@v1
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,7 +8,7 @@ on:
       - "**/*.md"
 jobs:
   build:
-    runs-on: macos-13
+    runs-on: macos-14
     steps:
       - uses: actions/checkout@v4
       - uses: maxim-lobanov/setup-xcode@v1


PostHog/Replay/PostHogSessionReplayConfig.swift
```diff
@@ -50,6 +50,18 @@
         /// Defaults to 1s
         @objc public var debouncerDelay: TimeInterval = 1.0
 
+        /// Whether to start session replay automatically or manually during the SDK setup
+        /// Options are:
+        /// - .automatic: Start session replay automatically during the SDK setup
+        /// - .manual: Start/Stop session replay manually by calling `startSessionReplay()` and `stopSessionReplay()`
+        /// Default is .automatic
+        @objc public var startMode: PostHogSessionReplayStartMode = .automatic
```
marandaneto, line 58: The JS SDK does not have this since it is automatic by default, but if `sessionReplay` is disabled, they can just call `startSessionRecording`
Any reason to be different?

PostHog/Replay/ViewLayoutTracker.swift
```diff
@@ -29,8 +29,8 @@
                 return
             }
             swizzle(forClass: UIView.self,
-                    original: #selector(UIView.layoutSubviewsOverride),
-                    new: #selector(UIView.layoutSubviews))
+                    original: #selector(UIView.layoutSubviews),
+                    new: #selector(UIView.layoutSubviewsOverride))
```
marandaneto, line 33: That feel wrong? because the selectors are inverted after swizzle

PostHogTests/PostHogContextTest.swift
```diff
@@ -50,10 +50,6 @@ class PostHogContextTest: QuickSpec {
 
             let context = sut.dynamicContext()
 
-            #if os(iOS) || os(tvOS)
-                expect(context["$screen_width"] as? Float) != nil
-                expect(context["$screen_height"] as? Float) != nil
-            #endif
```
marandaneto, line 56: why did we remove this?

PostHogTests/PostHogAutocaptureEventTrackerSpec.swift
```diff
@@ -19,7 +19,6 @@
                     let view = UIView()
                     let eventData = view.eventData!
 
-                    expect(eventData.targetClass).to(equal("UIView"))
```
marandaneto, line 22: Any reason to touch autocapture files in this PR? feels unrelated

PostHog/PostHogSessionManager.swift
```diff
@@ -8,115 +8,209 @@
 import Foundation
 
 // only for internal use
+// Do we need to expose this as public API? Could be internal static instead?
```
marandaneto, line 11: this is used by RN/Flutter, so yes since its accessed via a package dependency

PostHog/PostHogStorage.swift
```diff
@@ -15,7 +15,12 @@ import Foundation
  */
 func applicationSupportDirectoryURL() -> URL {
     let url = FileManager.default.urls(for: .applicationSupportDirectory, in: .userDomainMask).first!
-    return url.appendingPathComponent(Bundle.main.bundleIdentifier!)
+    #if canImport(XCTest) // only visible to test targets
+        return url.appendingPathComponent(Bundle.main.bundleIdentifier ?? "com.posthog.test")
+    #else
+        // TODO: Should we be using a fallback temp directory instead of force unwrapping here?
```
marandaneto, line 21: If for some reason this can be nil, I'd fall back to in-memory storage but we don't have this yet.
We cannot cache everything in the applicationSupportDirectory that is visible for all the apps, this has to be sandboxed for only this app.

PostHog/Replay/PostHogSessionReplayConfig.swift
marandaneto: > Could be a bit awkward though switching the config.sessionReplay behind the scenes for the user?

we won't do this, the user chooses to `sessionReplay: true` for everything automatically, or keep it `false` and start/stop manually, this is how the Web works.

PostHogTests/PostHogContextTest.swift
marandaneto: if this is `nil`, the tests would always fail and CI is happy right now, so I don't follow?

PostHogTests/PostHogSessionManagerTest.swift
```diff
@@ -0,0 +1,370 @@
+//
+//  PostHogSessionManagerTest.swift
+//  PostHog
+//
+//  Created by Yiannis Josephides on 16/12/2024.
+//
+
+import Foundation
+import Testing
```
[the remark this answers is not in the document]
marandaneto: any reason to switch from Nimble, and Quick that works across all tooling versions?

PostHogTests/PostHogContextTest.swift
marandaneto: True, I think I had issues with simulators back then and have gone with a purely swift test, but it seems like the time to do it if many tests depend on the simulator infrastructure, create an issue?

PostHogTests/PostHogSessionManagerTest.swift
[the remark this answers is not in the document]
marandaneto: if this is failing CI and it's not an easy fix, I'd rather keep the test infrastructure that works across different tooling versions.
No preference either as long as everything works.

.github/workflows/test.yml
```diff
@@ -8,11 +8,11 @@ on:
...
 jobs:
   test:
-    runs-on: macos-13
+    runs-on: macos-14
     steps:
       - uses: actions/checkout@v4
       - uses: maxim-lobanov/setup-xcode@v1
         with:
-          xcode-version: latest-stable
+          xcode-version: '16'
```
marandaneto, line 16: does not work with 16.2?

PostHog/PostHogSDK.swift
```diff
@@ -365,25 +372,14 @@ let maxRetryDelay = 30.0
...
         // reload flags as anon user
         if shouldReloadFlagsForTesting {
             reloadFeatureFlags()
         }
     }
 
-    private func resetViews() {
-        #if os(iOS)
-            if config.sessionReplay, featureFlags?.isSessionReplayFlagActive() ?? false {
-                replayIntegration?.resetViews()
```
marandaneto, line 382: how the views are reset if we don't call `replayIntegration?.resetViews()` anymore?

PostHog/PostHogSessionManager.swift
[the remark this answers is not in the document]
marandaneto: yes, https://github.com/search?q=repo%3APostHog%2Fposthog-react-native-session-replay+PostHogSessionManager&type=code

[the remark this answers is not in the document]
marandaneto: `setSessionId` has to exist and respect the RN session id since the id is generated and rotated in there, so if there's any code that would rotate everything here, should not do so for RN.
So the new conditions has to respect the `isNotReactNative`, does it already?

PostHog/PostHogSDK.swift
```diff
@@ -1013,25 +997,91 @@ let maxRetryDelay = 30.0
...
+
+        /**
+         Starts session recording.
+         This method will have no effect if PostHog is not enabled or if session replay integration is not available.
+         - Parameter resumeCurrent:
+            Whether to resume recording of current session (true) or start a new session (false).
+         */
+        @objc(startSessionRecordingWithResumeCurrent:)
+        public func startSessionRecording(resumeCurrent: Bool) {
+            if !isEnabled() {
```
marandaneto, line 1029: Should we check `isOptOutState` as well?
If we allow it as it is now and snapshots are dropped because `isOptOutState==true`, and later the user opts in, a lot of snapshots will be missing and the session will be broken, so maybe its better to not even start the session in this case?

PostHog/Replay/ViewLayoutTracker.swift
marandaneto: did you check this? I still think something is strange here since everything was working before (otherwise session replay wouldn't work at all), and I don't see how it's working after this change, unless the bug was only to remove the swizzling?

PostHog/PostHogSDK.swift
marandaneto: It's probably a use case I did not think of before, maybe makes more sense to raise a new issue with the problem, so we can address it separately, maybe the same problem happens on the other mobile platforms.

PostHog/PostHogSessionManager.swift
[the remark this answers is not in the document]
marandaneto: Every time the session rotates automatically, it cannot do it for RN.
For the methods that have to be called manually, it's ok, but you need to be sure that the SDK isn't calling those methods itself.

marandaneto (approved):

marandaneto (approved):

PR #279, opened 2024-12-23
fix: add UITextEffectsWindow type in screen ignore list

ioannisj:

follow up for: https://github.com/PostHog/posthog-ios/pull/269
ticket: https://posthoghelp.zendesk.com/agent/tickets/21072 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/24648)

We missed `UITextEffectsWindow` which is another window presented when keyboard appears (for SwiftUI at least). 

I also listed all the UIViewController types in code so we know what we believe we are ignoring. If any $screen event ever pops up with on of those UIViewController types, then it means it was presented on another UIWindow type. 

Note: We don't ignore any other system UIViewController / UIWindow (e.g UIActivityViewController) but: 
1. I doubt these will ever prove problematic in terms of billing (a messaging app was flooded with keyboard screen events)
2. They may be even useful for some customers

Local sample project

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Next
 
+- fix: ignore additional keyboard windows for $screen event ([#279](https://github.com/PostHog/posthog-ios/pull/279))
+
 ## 3.17.1 - 2024-12-18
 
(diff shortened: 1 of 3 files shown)


pauldambra (approved):
👍

Issue #274, opened 2024-12-18
Requested URL is incorrect when PostHog host includes a path


bcheidemann:
### Version

3.17.0

### Steps to Reproduce

See https://github.com/PostHog/posthog-flutter/issues/128

1. Set `com.posthog.posthog.POSTHOG_HOST` to `http://localhost:3000/some/path`
2. Run a dummy server on port `3000` which logs incoming requests
3. Trigger a batch of events to be sent
4. Observe that the requested URL is `http://localhost:3000/batch`

### Expected Result

The requested URL is `http://localhost:3000/some/path/batch`

### Actual Result

The requested URL is `http://localhost:3000/batch`

marandaneto:
Before closing this, I'd need to check if the same happens on Android and RN, good catch.

marandaneto:
> Before closing this, I'd need to check if the same happens on Android and RN, good catch.

Android just appends the endpoint so the path would be `http://localhost:3000/some/path/batch` correctly

PR #283, opened 2025-01-21
feat: load remote config and feature flags

ioannisj:

Building the foundation for mobile surveys. This is one part of many on the feature branch `feat/mobile-surveys`. 

This adds remote config loading and caching on SDK
- Added remote config loading and caching on startup
- Refactored a bit feature flags loading to take into account remote config value `hasFeatureFlags`. When this value is false, we can skip loading feature flags cause the team doesn't have any flags enabled. 
- On new remote config load, if `hasFeatureFlags` is true we make sure that flags are reloaded

#skip-changelog

Modified and added some unit tests

## :pencil: Checklist


PostHog/PostHogApi.swift
```diff
@@ -35,6 +35,26 @@ class PostHogApi {
...
+            URL(string: "https://us-assets.i.posthog.com")
+        case "https://eu.i.posthog.com":
+            URL(string: "https://eu-assets.i.posthog.com")
+        default:
+            config.host
+        } else {
+            return nil
+        }
+
+        let url = baseUrl.appendingPathComponent("/array/\(config.apiKey)/config")
```
marandaneto, line 50: I think we should fix https://github.com/PostHog/posthog-ios/issues/274 otherwise its one more place that this will fail

PostHog/PostHogApi.swift
```diff
@@ -217,4 +237,45 @@ class PostHogApi {
...
+
+        let task = URLSession(configuration: config).dataTask(with: request) { data, response, error in
+            if let error {
+                hedgeLog("Error calling the remote config API: \(error.localizedDescription)")
+                return completion(nil, error)
+            }
+
+            let httpResponse = response as! HTTPURLResponse
+
+            if !(200 ... 299 ~= httpResponse.statusCode) {
```
marandaneto, line 259: we can probably double check with ben that is the expected errors here and if there's any special case.

PostHog/PostHogRemoteConfig.swift
```diff
@@ -82,7 +167,46 @@ class PostHogFeatureFlags {
...
+    }
+
+    private func loadFeatureFlagsInternal(
+        distinctId: String,
+        anonymousId: String,
+        groups: [String: String],
+        callback: @escaping () -> Void
+    ) {
+        let hasFeatureFlags = remoteConfigLock.withLock {
+            self.remoteConfig?["hasFeatureFlags"] as? Bool
```
marandaneto, line 201: we can make a const since we are repeating this key

marandaneto (approved):

PR #284, opened 2025-01-23
feat: add survey domain models

ioannisj:

[Mobile Surveys] - Part 2

Add survey domain models and parse logic. 
I've included all properties for now. Will probably strip down or comment out at the end when we know which of these properties are not be used. 

see: https://github.com/PostHog/posthog-js/blob/a5591088db330e4c0d235b63a55a9b6fc75d84b0/src/posthog-surveys-types.ts#L147

#skip-changelog

Added unit tests

## :pencil: Checklist


PostHog/Models/Survey.swift
```diff
@@ -0,0 +1,346 @@
...
+enum SurveyQuestion: Decodable {
+    case basic(BasicSurveyQuestion)
+    case link(LinkSurveyQuestion)
+    case rating(RatingSurveyQuestion)
+    case multipleChoice(MultipleSurveyQuestion)
+    case singleChoice(MultipleSurveyQuestion)
+
+    init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        let type = try container.decode(SurveyQuestionType.self, forKey: .type)
```
marandaneto, line 78: when the code throws we need to be sure to swallow exception and log out and not crash peoples apps

PostHog/PostHogApi.swift
```diff
@@ -279,3 +279,27 @@ class PostHogApi {
...
 }
+
+extension PostHogApi {
+    static var jsonDecoder: JSONDecoder = {
+        let decoder = JSONDecoder()
+
+        let dateFormatter = DateFormatter()
+        dateFormatter.locale = Locale(identifier: "en_US_POSIX")
+        dateFormatter.timeZone = TimeZone(abbreviation: "UTC")
+        dateFormatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'"
```
marandaneto, line 290: can you reuse the DateUtils file?

PostHog/PostHogApi.swift
```diff
@@ -279,3 +279,27 @@ class PostHogApi {
...
+            let container = try decoder.singleValueContainer()
+            let dateString = try container.decode(String.self)
+            guard let date = dateFormatter.date(from: dateString) else {
+                throw DecodingError.dataCorruptedError(
+                    in: container, debugDescription: "Invalid date format"
+                )
+            }
+            return date
+        }
+        decoder.keyDecodingStrategy = .convertFromSnakeCase
```
marandaneto, line 302: Does the API return in the snake case? we need to be sure this is consistent, if there are mixed strategies, this can fail.

PostHogTests/PostHogSurveysTest.swift
```diff
@@ -0,0 +1,201 @@
...
+            #expect(sut.currentIteration == nil)
+            #expect(sut.currentIterationStartDate == nil)
+        }
+
+        @Test("Survey question types decode correctly")
+        func surveyQuestionTypesDecodeCorrectly() throws {
+            let data = try loadFixture("fixture_survey_question_types")
+            let sut = try PostHogApi.jsonDecoder.decode(Survey.self, from: data)
+
+            // Basic question
```
marandaneto, line 71: each question type should be its own test, so if it fails, we know each type failed

PostHogTests/Resources/fixture_survey_basic.json
```diff
@@ -0,0 +1,43 @@
+{
+    "id": "01947134-8a35-0000-549a-193b86fa2e44",
```
marandaneto, line 2: are you generating the fixtures via the UI, using the API type? (copy-pasta)

PostHog/PostHogApi.swift
marandaneto: what `convertFromSnakeCase` do i know, what I meant is that we cannot guarantee that all fields we are going to use are following this convention (since FE and BE have legacy code in different naming strategies, maybe its not an issue for surveys though), but as long as you check and all fields are correctly serializable, then all fine.

marandaneto (approved):
nice work!

PR #287, opened 2025-01-26
fix: identify mac catalyst

ioannisj:

see: https://posthog.com/questions/i-os-app-also-available-for-mac-os-is-not-showing-any-mac-os-users-in-product-analytics

Identify as `macOS` when running on Mac Catalyst or compatibility mode (Designed for iPad/iPhone). 

Here are the keys for the two cases: 

### Mac Catalyst

```
"$device_manufacturer": "Apple"
"$device_model": "Mac15,6"
"$device_name: "macbook-pro.local"
"$device_type": "Desktop"
"$is_emulator": "false"
"$is_ios_running_on_mac": "false"
"$is_mac_catalyst_app": "true"
"$os_name: "macOS"
"$os_version": "15.1.1"
```
  
### iOS App running on Mac (Designed for iPad/iPhone)

```
"$device_manufacturer": "Apple"
"$device_model": "Mac15,6"
"$device_name": "macbook-pro.local"
"$device_type": "Desktop"
"$is_emulator": "false"
"$is_ios_running_on_mac": "true"
"$is_mac_catalyst_app": "false"
"$os_name": "macOS"
"$os_version": "15.1.1"
```

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 ## Next
 
 - fix: identify macOS when running Mac Catalyst or iOS on Mac ([#287](https://github.com/PostHog/posthog-ios/pull/287))
+
+## 3.19.2 - 2025-01-30
+
+- fix: XCFramework builds failing ([#288](https://github.com/PostHog/posthog-ios/pull/288))
 - chore: Session Replay GA ([#286](https://github.com/PostHog/posthog-ios/pull/286))
 
 ## 3.19.1 - 2025-01-13
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,13 @@ buildSdk:
 	set -o pipefail && xcrun xcodebuild clean build -scheme PostHog -destination generic/platform=watchos | xcpretty #watchOS
 	set -o pipefail && xcrun xcodebuild clean build -scheme PostHog -destination 'platform=visionOS Simulator,name=Apple Vision Pro' | xcpretty #visionOS
 
-buildExamples:
+buildExamples: \
+	buildExamplesPlatforms \
+	buildExampleXCFramework \
+	buildExamplePodsDynamic \
+	buildExamplePodsStatic \
+
+buildExamplesPlatforms:
 	set -o pipefail && xcrun xcodebuild -downloadAllPlatforms
 	set -o pipefail && xcrun xcodebuild clean build -scheme PostHogExample -destination generic/platform=ios | xcpretty #ios
 	set -o pipefail && xcrun xcodebuild clean build -scheme PostHogExample -destination 'platform=visionOS Simulator,name=Apple Vision Pro' | xcpretty #visionOS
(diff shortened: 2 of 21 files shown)


PostHog/PostHogContext.swift
```diff
@@ -56,28 +56,62 @@ class PostHogContext {
...
+
+            if isMacCatalystApp || isiOSAppOnMac {
+                let underlyingOS = device.systemName
+                let underlyingOSVersion = device.systemVersion
+                let macOSVersion = processInfo.operatingSystemVersionString
+                // e.g macOS Version 15.1.1 (Build 24B91) / iPadOS 18.1
+                properties["$os_version"] = "macOS \(macOSVersion) / \(underlyingOS) \(underlyingOSVersion)"
+                properties["$os_name"] = isMacCatalystApp
+                    ? "macOS (Mac Catalyst)"
+                    : "macOS (iOS running on Mac)"
```
marandaneto, line 87: The fix is great but we need an alternative solution here.
People use the os_name and os_version with matching values in feature flags, hogql, etc.
Maybe we need a new property to tell if it's a mac catalyst or not then? like a boolean one.

marandaneto: I think whether it's catalyst or not is more about the app itself and not the OS, so maybe this info should be with one of the `$app_X` properties. the underlying OS is still the same.

marandaneto: I think `$os_` should be about the running OS on the physical device, if that's wrong, we have to fix it.
I think `$device_` should be about the physical device, if that's wrong, we have to fix it.
Adding a new app flag makes sense.

PostHog/PostHogContext.swift
```diff
@@ -53,30 +53,73 @@ class PostHogContext {
...
+            let processInfo = ProcessInfo.processInfo
+
+            if isMacCatalystApp || isiOSAppOnMac {
+                let underlyingOS = device.systemName
+                let underlyingOSVersion = device.systemVersion
+                let macOSVersion = processInfo.operatingSystemVersionString
+
+                if isMacCatalystApp {
+                    let processInfo = ProcessInfo.processInfo
+                    properties["$os_name"] = "macOS \(processInfo.operatingSystemVersionString)" // eg Version 14.2.1 (Build 23C71)
```
marandaneto, line 86: this should be os_name and not version

PostHog/PostHogContext.swift
```diff
@@ -53,30 +53,73 @@ class PostHogContext {
...
+                } else {
+                    properties["$os_name"] = device.systemName
+                    properties["$os_version"] = device.systemVersion
+                }
+                // device.userInterfaceIdiom reports .pad here, so we use a static value instead
+                // - For an app deployable on iPad, the idiom type is always .pad (instead of .mac)
+                // 
+                // Source: https://developer.apple.com/documentation/apple-silicon/adapting-ios-code-to-run-in-the-macos-environment#Handle-unknown-device-types-gracefully
+                properties["$device_type"] = "Desktop"
+                properties["$device_name"] = processInfo.hostName
```
marandaneto, line 98: what does hostName return here? this is about the device name (Model), and not the user's name or device's nickname

PostHog/PostHogContext.swift
marandaneto: > For iOS running on Mac, $os_ points to iPadOS

@ioannisj not sure if this is correct? its a mac, so the OS is the mac os, will people think they are running the app on an ipadoS?

marandaneto: So again just to confirm, the `os` and `device` props should be related to the physical device.
if the app is in compat. mode, this info should go either with `app` props or new props.
if you want to add the info of the emulated OS in compat mode, this should go under new props, right now `device` and `os` props are meant for the physical device.

PostHog/PostHogContext.swift
marandaneto: this should be the info of the physical device.
So if we're using `Host.current().localizedName` because `device.model` isn't possible, then lets keep it

PostHog/PostHogContext.swift
marandaneto: mm i feel like the naming should be deterministic since it can be used for filtering.
having the version as part of it feels wrong.
we have os_version for that.
I'd remove from both.

pauldambra:
fly-by that I appreciate the care and attention on this PR... some decisions are pretty hard to change and these posthog provided props are definitely that 💖

marandaneto (approved):

octohub:
@pauldambra @ioannisj Thank you so much for the work on this. Excited to leverage it! How can I track when this will be in a published version of the SDK?

marandaneto:
> @pauldambra @ioannisj Thank you so much for the work on this. Excited to leverage it! How can I track when this will be in a published version of the SDK?

https://github.com/PostHog/posthog-ios/releases/tag/3.19.3 :)

PR #288, opened 2025-01-28
fix: xcframework builds

ioannisj:

see: https://posthog.slack.com/archives/C06GCP04DT7/p1737633958220259

The way we were importing `phlibwebp` caused it to be exposed in the .swiftinterface of the PostHog library. This unintended exposure occurred because `phlibwebp` was included as a public dependency (currently default) in code, which made its symbols visible outside of the library. In turn, this became a build issue when building an xcframework since usually frameworks are archived using the `BUILD_LIBRARY_FOR_DISTRIBUTION=YES` for module stability. 

To resolve this, the import of `phlibwebp` was made internal which ensures it does not leak into the public interface of the library. 

#skip-changelog

- Confirmed successful builds using a new xcframework sample project. 
- xcframework flow (archive -> create xcframework -> embed in client app) is now tested as part of `make buildExamples`

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Next
 
 - fix: XCFramework builds failing ([#288](https://github.com/PostHog/posthog-ios/pull/288))
+- chore: Session Replay GA ([#286](https://github.com/PostHog/posthog-ios/pull/286))
 
 ## 3.19.1 - 2025-01-13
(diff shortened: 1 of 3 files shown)


marandaneto:
Probably deserves a changelog entry?

marandaneto (approved):

PR #291, opened 2025-02-03
Add visionOS support

harlanhaskins:
This switches to `canImport(UIKit)`/`canImport(AppKit)`/`canImport(WatchKit)` instead of OS-specific checks in the cases where the only requirement is importing the specific framework and using those symbols.

This also adds an entry into the device type matching for Apple Vision Pro and updates the demo project and tests to also run on visionOS.

We're currently unable to use PostHog on visionOS because of compilation issues, and we would like to be able to.

Added build support to the Demo project for Apple Vision Pro and ran in the simulator. All tests continue to pass. Note: I had to update Nimble to the latest version for the test to be able to build successfully.

## :pencil: Checklist


marandaneto:
@harlanhaskins thanks for the PR.

@ioannisj will let you review this.
One thing to pay attention to is when to import `ios` or `uikit`, session replay is `ios` only so we need to be sure we're not leaking any method, view modifier, etc for platforms that have uikit but not session replay yet, otherwise it seems legit.

harlanhaskins:
Ah, you’re completely right, I did neglect to think of the special position watchOS is in. Honestly I think it’s not worth switching to canImport for “consistency” if we have to break the rule in so many places to explicitly exclude watchOS.

I’ll go through this again and manually add `|| os(visionOS)` where it makes sense instead of blanket using canImport. There are more iOS APIs exposed to visionOS than tvOS so I don’t think it’ll be a find&replace operation.

marandaneto:
> Ah, you’re completely right, I did neglect to think of the special position watchOS is in. Honestly I think it’s not worth switching to canImport for “consistency” if we have to break the rule in so many places to explicitly exclude watchOS.
> 
> I’ll go through this again and manually add `|| os(visionOS)` where it makes sense instead of blanket using canImport. There are more iOS APIs exposed to visionOS than tvOS so I don’t think it’ll be a find&replace operation.

@harlanhaskins would you be able to finish this up? thanks for the follow up.

harlanhaskins:
@marandaneto My apologies; Admittedly I completely dropped this. I'll pick it back up and resubmit 🫡

PR #294, opened 2025-02-07
fix: safely convert floats to int

### LINKED ISSUE #293: Crash on iPad using sessionReplay = true
### Version

3.19.1

### Steps to Reproduce

1. App with tab bar and switui home
2. Init PostHog with config.sessionReplay = true
3. Got a crash at app launch

### Expected Result

No crash at all

### Actual Result

Crash at
PostHogReplayIntegration.swift
Line: 635
Reason: Thread 1: Swift runtime failure: Double value cannot be converted to Int because it is either infinite or NaN

Image

ioannisj:

Closes #293 

```
let nanNumber = NSNumber(value: CGFloat.nan) // NaN
let frNumber = NSNumber(value: CGFloat(4.5)) // 4.1
let gfmNumber = NSNumber(value: CGFloat.greatestFiniteMagnitude) // 1.797693134862316E+308

nanNumber.intValue // 0
gfmNumber.intValue // Int.max (9,223,372,036,854,775,807)
frNumber.intValue // 4
```

## :pencil: Checklist


PostHog/Replay/Float+.swift
```diff
@@ -0,0 +1,18 @@
+//
+//  Float+.swift
```
marandaneto, line 2: all other files are called `Name+Util.swift, should we keep the pattern?

marandaneto:
the unit tests could be part of the tests instead of the PR description :)

PR #299, opened 2025-02-14
feat: support multiple PostHog instances part I

ioannisj:

First part of supporting multiple instances 

Todo: 
- [x] Now storage is per api key, with some old storage migration logic so we don't miss any events. Old events will be migrated to the first API used
- [x] Maintain one instance per api key. So, calling .with() with the same API key should return the same PostHogSDK instance
- [x] Replay and element autocapture integrations installed only on first created instance
- [ ] `captureApplicationLifecycleEvents` (autocapture) only on first created instance
- [ ] `captureScreenViews` only on first created instance

#skip-changelog

## :pencil: Checklist


PostHog/Autocapture/PostHogAutocaptureIntegration.swift
```diff
@@ -11,11 +11,35 @@
...
+        private static var integrationInstalled = false
+
+        private weak var postHogInstance: PostHogSDK?
         private var debounceTimers: [Int: Timer] = [:]
 
-        init(_ config: PostHogConfig) {
-            postHogConfig = config
+        init?(_ posthog: PostHogSDK) {
+            Self.integrationInstalledLock.lock()
+            defer { Self.integrationInstalledLock.unlock() }
```
marandaneto, line 22: I usually do `lock.withLock {...}`
Is there any difference?

PostHog/Autocapture/PostHogAutocaptureIntegration.swift
```diff
@@ -11,11 +11,35 @@
     private let elementsChainDelimiter = ";"
 
     class PostHogAutocaptureIntegration: AutocaptureEventProcessing {
-        private let postHogConfig: PostHogConfig
+        private static var integrationInstalledLock = NSLock()
+        private static var integrationInstalled = false
+
+        private weak var postHogInstance: PostHogSDK?
```
marandaneto, line 17: does it have to be `weak` if we control when we pass the instance and we set it to null?

PostHog/Autocapture/PostHogAutocaptureIntegration.swift
```diff
@@ -11,11 +11,35 @@
...
+
+        func uninstall(_ posthog: PostHogSDK) {
+            Self.integrationInstalledLock.lock()
+            defer { Self.integrationInstalledLock.unlock() }
+
+            // uninstall only for integration instance
+            if postHogInstance === posthog {
+                postHogInstance = nil
+                Self.integrationInstalled = false
+            }
```
marandaneto, line 42: good point, i think i missed that on android

PostHog/PostHogSDK.swift
```diff
@@ -44,7 +44,8 @@ let maxRetryDelay = 30.0
     private var flagCallReported = Set<String>()
     private var featureFlags: PostHogFeatureFlags?
     private var context: PostHogContext?
-    private static var apiKeys = Set<String>()
+    private static let instancesLock = NSLock()
+    private static var instances: [String: Weak<PostHogSDK>] = [:]
```
marandaneto, line 48: i have not done that on android, wondering if this is a better (or not) approach? can we chat about this?

PostHog/PostHogStorage.swift
```diff
@@ -15,16 +15,54 @@ import Foundation
...
+  */
+func appGroupContainerUrl(config: PostHogConfig) -> URL? {
+    guard let appGroupIdentifier = config.appGroupIdentifier else { return nil }
+
     #if canImport(XCTest) // only visible to test targets
-        return url.appendingPathComponent(Bundle.main.bundleIdentifier ?? "com.posthog.test")
+        let bundleIdentifier = "com.posthog.test"
     #else
-        return url.appendingPathComponent(Bundle.main.bundleIdentifier!)
+        let bundleIdentifier = Bundle.main.bundleIdentifier!
```
marandaneto, line 47: can we create a func to return the `Bundle.main.bundleIdentifier`? since this code now is dupe here and above

PostHog/PostHogStorage.swift
```diff
@@ -122,39 +160,51 @@ class PostHogStorage {
     /**
      There are cases where applications using posthog-ios want to share analytics data between host app and
      an app extension, Widget or App Clip. If there's a defined `appGroupIdentifier` in configuration,
-     we want to use a shared container for storing data so that extensions correcly identify a user (and batch process events)
+     we want to use a shared container for storing data so that extensions correctly identify a user (and batch process events)
      */
+    private static func getBaseAppFolderUrl(from configuration: PostHogConfig) -> URL {
```
marandaneto, line 165: i will review this later today

PostHog/PostHogSDK.swift
marandaneto: discussed offline, we don't need this.

PostHog/PostHogStorage.swift
```diff
@@ -122,39 +160,51 @@ class PostHogStorage {
...
+                    // Migrate the item and its contents if it exists
+                    try migrateItem(at: legacyFileUrl, to: newFileUrl, fileManager: fileManager)
+
+                    // Remove the legacy item after successful migration
+                    if fileManager.fileExists(atPath: legacyFileUrl.path) {
+                        try fileManager.removeItem(at: legacyFileUrl)
+                    }
+                }
+            } catch {
+                hedgeLog("Error during storage migration: \(error)")
```
marandaneto, line 203: I think the migration should be its own method, its strange to be within the `getAppFolderUrl`

marandaneto, line 203: Lets add the path as part of the error as well, so in case things go wrong, and people share the logging, we might spot something, unless the `error` is clear about the paths

PostHog/Autocapture/PostHogAutocaptureIntegration.swift
marandaneto: nope, but I don't know how the swift reference counting works deeply, so as long as this instance isn't null randomly, all good

PostHog/PostHogSDK.swift
```diff
@@ -1297,4 +1297,14 @@ let maxRetryDelay = 30.0
     #endif
 }
 
+#if TESTING
```
marandaneto, line 1300: i like this trick

PostHog/PostHogStorage.swift
```diff
@@ -46,7 +83,8 @@ class PostHogStorage {
     init(_ config: PostHogConfig) {
         appFolderUrl = Self.getAppFolderUrl(from: config)
 
-        createDirectoryAtURLIfNeeded(url: appFolderUrl)
+        // migrate legacy storage if needed
+        Self.migrateLegacyStorage(from: config, to: appFolderUrl)
```
marandaneto, line 87: should this be done after the SDK is init?
just wondering if `config` is needed but the values set are not fully started yet because this is executed in the ctor before the SDK is fully functional

PostHog/PostHogStorage.swift
```diff
@@ -15,16 +15,53 @@ import Foundation
...
+    let bundleIdentifier = getBundleIdentifier()
+
+    let url = FileManager.default
+        .containerURL(forSecurityApplicationGroupIdentifier: appGroupIdentifier)?
+        .appendingPathComponent("Library/Application Support/")
+        .appendingPathComponent(bundleIdentifier)
+
+    if let url {
+        createDirectoryAtURLIfNeeded(url: url)
+        return directoryExists(url) ? url : nil
```
marandaneto, line 48: can this throw?

PostHog/PostHogStorage.swift
```diff
@@ -122,39 +160,52 @@ class PostHogStorage {
...
-                return directoryExists(url) ? url : nil
-            }
+    private static func getBaseAppFolderUrl(from configuration: PostHogConfig) -> URL {
+        appGroupContainerUrl(config: configuration) ?? applicationSupportDirectoryURL()
+    }
+
+    private static func migrateItem(at sourceUrl: URL, to destinationUrl: URL, fileManager: FileManager) throws {
+        guard fileManager.fileExists(atPath: sourceUrl.path) else { return }
+        // Copy file or directory over (if it doesn't exist)
+        if !fileManager.fileExists(atPath: destinationUrl.path) {
```
marandaneto, line 172: can this throw?

PostHog/PostHogStorage.swift
```diff
@@ -122,39 +160,52 @@ class PostHogStorage {
...
+                do {
+                    // Migrate the item and its contents if it exists
+                    try migrateItem(at: legacyFileUrl, to: newFileUrl, fileManager: fileManager)
+
+                    // Remove the legacy item after successful migration
+                    if fileManager.fileExists(atPath: legacyFileUrl.path) {
+                        try fileManager.removeItem(at: legacyFileUrl)
+                    }
+                } catch {
+                    hedgeLog("Error during storage migration for file \(storageKey.rawValue) at path \(legacyFileUrl.path): \(error)")
```
marandaneto, line 196: should we delete the file if it failed anyways?

marandaneto (approved):
left a few comments but otherwise LGTM

PR #305, opened 2025-02-19
fix: react native builds as static lib

### LINKED ISSUE #292: Build fails for React Native when compiled as a static library
### Description

While linking as a static or dynamic framework works in RN sample project. When linked as a static library (`use_frameworks!` not used in Podfile) then we get some build errors. 

see: https://github.com/PostHog/posthog-react-native-session-replay/actions/runs/13153636208/job/36705772315?pr=22

ioannisj:

Closes #292

This should fix the error we are seeing when trying to [build](https://github.com/PostHog/posthog-react-native-session-replay/actions/runs/13153636208/job/36705772315?pr=22) react native replay plugin as a static lib. 

Looks like when Pods are compiled as a **static lib**, everything is flattened and compiled together. We have a "utils.h" file that collided with a similarly named file from one of rn dependencies (utils.h, types.h are common names, so I suspect this was breaking in other cases as well). 

I'm not extremely knowledgable on how to properly deal with C headers, but the current fix is to basically append "ph_" in front of every header to make it more unique. 

Also reworked our Pods example to compile as dynamic framework, static framework and static lib. 

I'll follow up with another fix on navigation masking. Once a new patch version is out, I expect retrying the [failing job](https://github.com/PostHog/posthog-react-native-session-replay/actions/runs/13153636208/job/36705772315?pr=22) to ✅

#skip-changelog

Tested locally by: 
- checkout https://github.com/PostHog/posthog-react-native-session-replay
- Add `pod "PostHog", :podspec => "../../../posthog-ios/PostHog.podspec"` to example Podfile
- Modify local PostHog.podspec to ` s.source  = { :git => "" }` (this will point to local git, but changes need to be committed first)
- Build example iOS project 

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,19 @@
 ## Next
 
+- fix: recordings not always properly masked during screen transitions ([#306](https://github.com/PostHog/posthog-ios/pull/306))
+
+## 3.19.6 - 2025-02-18
+
+- fix: crash on autocapture when a segmented control has not selection ([#304](https://github.com/PostHog/posthog-ios/pull/304))
+
+## 3.19.5 - 2025-02-11
+
+- fix: flutter session recordings not working ([#297](https://github.com/PostHog/posthog-ios/pull/297))
+
+## 3.19.4 - 2025-02-07
+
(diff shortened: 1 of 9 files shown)


marandaneto:
> Also reworked our Pods example to compile as dynamic framework, static framework and static lib.

love this!

PR #308, opened 2025-02-23
here we go

dmarticus:
with https://github.com/PostHog/posthog/pull/28564, we now start to respond different with the /decide APIs if users have gone over their quota limit.

### DIFF
--- a/PostHog/PostHogFeatureFlags.swift
+++ b/PostHog/PostHogFeatureFlags.swift
@@ -94,6 +94,20 @@ class PostHogFeatureFlags {
                    groups: groups)
         { data, _ in
             self.dispatchQueue.async {
+                // Check for quota limitation first
+                if let quotaLimited = data?["quotaLimited"] as? [String],
+                   quotaLimited.contains("feature_flags") {
+                    hedgeLog("Warning: Feature flags quota limit reached - clearing all feature flags and payloads")
+                    self.featureFlagsLock.withLock {
+                        // Clear both feature flags and payloads
+                        self.setCachedFeatureFlags([:])
+                        self.setCachedFeatureFlagPayload([:])
+                    }
+                    
+                    self.notifyAndRelease()
+                    return callback()
+                }
+
                 guard let featureFlags = data?["featureFlags"] as? [String: Any],
                       let featureFlagPayloads = data?["featureFlagPayloads"] as? [String: Any]
                 else {
--- a/PostHogTests/TestUtils/MockPostHogServer.swift
+++ b/PostHogTests/TestUtils/MockPostHogServer.swift
@@ -41,6 +41,7 @@ class MockPostHogServer {
     public var returnReplayWithMultiVariant = false
     public var replayVariantName = "myBooleanRecordingFlag"
     public var replayVariantValue: Any = true
+    public var quotaLimitFeatureFlags: Bool = false
 
     init() {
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## Next
 
+## 3.20.0 - 2025-02-24
```
marandaneto, line 3:
```suggestion
## 3.20.0 - 2025-02-25
```

PostHogTests/TestUtils/MockPostHogServer.swift
```diff
@@ -53,6 +54,10 @@ class MockPostHogServer {
                 self.replayVariantName: self.replayVariantValue,
             ]
 
+            if self.quotaLimitFeatureFlags {
```
marandaneto, line 57: @dmarticus can we add a unit test for this as well?
that would go here https://github.com/PostHog/posthog-ios/blob/main/PostHogTests/PostHogFeatureFlagsTest.swift

dmarticus: done!

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## Next
 
+## 3.20.0 - 2025-02-25
```
marandaneto, line 3:
```suggestion
## 3.20.0 - 2025-02-26
```

marandaneto:
> > let's figure out [session recording](https://github.com/PostHog/posthog-android/pull/228#discussion_r1966811362) before merging this btw
> 
> I think as it currently stands, session recording will default to [enabled](https://github.com/PostHog/posthog-ios/blob/db977410f409887c4f12fa56e65f1c3cf3f831f8/PostHog/PostHogFeatureFlags.swift#L54) if flag is missing. Android has a similar implementation. Not sure I'm reading this correctly, @marandaneto am I missing something?

i think we have to patch this part https://github.com/PostHog/posthog-ios/blob/db977410f409887c4f12fa56e65f1c3cf3f831f8/PostHog/PostHogFeatureFlags.swift#L59
if `linkedFlag` is provided but does not exist within the `featureFlags`, `recordingActive` should be false, can you patch this?

PostHog/PostHogFeatureFlags.swift
```diff
@@ -55,10 +55,8 @@ class PostHogFeatureFlags {
         var recordingActive = true
 
         // check for boolean flags
-        if let linkedFlag = sessionRecording["linkedFlag"] as? String,
-           let value = featureFlags[linkedFlag] as? Bool
-        {
-            recordingActive = value
+        if let linkedFlag = sessionRecording["linkedFlag"] as? String {
+            recordingActive = featureFlags[linkedFlag] as? Bool ?? false
```
marandaneto, line 59: @ioannisj we should do something similar for the next check too:
L64 most likely

marandaneto: `featureFlags[linkedFlag]` can also be a string, and that means its enabled, this is the case for multivariant, linked with "any"

marandaneto: See PR [PostHog/posthog-android@`2929c4d` (#230)](https://github.com/PostHog/posthog-android/pull/230/commits/2929c4de1b808ef026d1b1e77dd401d6617564ca)

marandaneto (approved):

PR #309, opened 2025-02-24
feat: support multiple PostHog instances part II

ioannisj:

Second part of supporting multiple instances 

Todo: 
- [x] Now storage is per api key, with some old storage migration logic so we don't miss any events. Old events will be migrated to the first API used
- [x] Maintain one instance per api key. So, calling .with() with the same API key should return the same PostHogSDK instance
- [x] Replay and element autocapture integrations installed only on first created instance
- [x] `captureApplicationLifecycleEvents` (autocapture) only on first created instance
- [x] `captureScreenViews` only on first created instance
- [x] Decouple URLSessionInterceptor from PostHogSDK.shared
- [x] Decouple UIApplication.sendEvent from PostHogSDK.shared
- [x] Decouple SwiftUI `postHogScreenView` from PostHogSDK.shared
- [x] Decouple URLSession extensions (replay network logs) from PostHogSDK.shared

#skip-changelog

## :pencil: Checklist

### DIFF
--- a/PostHog/PostHogIntegration.swift
+++ b/PostHog/PostHogIntegration.swift
@@ -49,32 +49,3 @@ protocol PostHogIntegration {
      */
     func stop()
 }
-
-extension PostHogConfig {
-    /// Array of integrations to be installed based on current configuration
-    var integrations: [PostHogIntegration] {
-        var integrations: [PostHogIntegration] = []
-
-        if captureScreenViews {
-            integrations.append(PostHogScreenViewIntegration())
-        }
-
-        if captureApplicationLifecycleEvents {
-            integrations.append(PostHogAppLifeCycleIntegration())
-        }
-
-        #if os(iOS)
-            if sessionReplay {
-                integrations.append(PostHogReplayIntegration())
-            }
-        #endif
-
-        #if os(iOS) || targetEnvironment(macCatalyst)
-            if captureElementInteractions {
-                integrations.append(PostHogAutocaptureIntegration())
-            }
-        #endif
-
-        return integrations
-    }
-}
--- a/PostHog/PostHogConfig.swift
+++ b/PostHog/PostHogConfig.swift
@@ -81,4 +81,31 @@ import Foundation
         self.apiKey = apiKey
         self.host = URL(string: host) ?? URL(string: PostHogConfig.defaultHost)!
     }
+
+    /// Returns an array of integrations to be installed based on current configuration
+    func getIntegrations() -> [PostHogIntegration] {
+        var integrations: [PostHogIntegration] = []
+
+        if captureScreenViews {
+            integrations.append(PostHogScreenViewIntegration())
+        }
+
+        if captureApplicationLifecycleEvents {
+            integrations.append(PostHogAppLifeCycleIntegration())
+        }
+
+        #if os(iOS)
+            if sessionReplay {
+                integrations.append(PostHogReplayIntegration())
+            }
+        #endif
+
+        #if os(iOS) || targetEnvironment(macCatalyst)
+            if captureElementInteractions {
+                integrations.append(PostHogAutocaptureIntegration())
+            }
+        #endif
+
+        return integrations
+    }
 }
--- a/PostHog/PostHogSDK.swift
+++ b/PostHog/PostHogSDK.swift
@@ -1078,7 +1078,7 @@ let maxRetryDelay = 30.0
     #endif
 
     private func installIntegrations() {
-        let integrations = config.integrations
+        let integrations = config.getIntegrations()
         var installed: [PostHogIntegration] = []
 
         for integration in integrations {


PostHog/App Life Cycle/PostHogAppLifeCycleIntegration.swift
```diff
@@ -0,0 +1,194 @@
...
+        }
+
+        guard !wasInstalled else { return nil }
+
+        postHog = posthog
+
+        // needs to happen on next runloop cause we are still in SDK init sequence here
+        DispatchQueue.main.async {
+            self.captureAppInstallOrUpdated()
+        }
```
marandaneto, line 49: I think to avoid this we can do this:
https://github.com/PostHog/posthog-android/blob/41ac8e9dbf989348823052c8c0f14c9c0d1fa2fd/posthog/src/main/java/com/posthog/PostHog.kt#L125-L131
we create the instances, and "install" them once the SDK is operational at the end of the setup method

PostHog/App Life Cycle/PostHogAppLifeCycleIntegration.swift
```diff
@@ -0,0 +1,194 @@
...
+     Stop capturing app lifecycle events
+     */
+    func stop() {
+        didFinishLaunchingToken = nil
+        didBecomeActiveToken = nil
+        didEnterBackgroundToken = nil
+    }
+
+    private func captureAppInstallOrUpdated() {
+        guard let postHog, !didCaptureAppInstalled else { return }
```
marandaneto, line 88:
```suggestion
        guard let postHog, !captureAppInstallOrUpdated else { return }
```

PostHog/App Life Cycle/PostHogAppLifeCycleIntegration.swift
```diff
@@ -0,0 +1,194 @@
...
+        }
+
+        postHog.capture(event, properties: props)
+    }
+
+    private func captureAppOpened() {
+        guard let postHog else { return }
+
+        if !postHog.config.captureApplicationLifecycleEvents {
+            return
```
marandaneto, line 150: we should log `hedgeLog` here

PostHog/App Life Cycle/PostHogAppLifeCycleIntegration.swift
```diff
@@ -0,0 +1,194 @@
...
+            appFromBackground = true
+        }
+
+        postHog.capture("Application Opened", properties: props)
+    }
+
+    private func captureAppBackgrounded() {
+        guard let postHog else { return }
+
+        if !postHog.config.captureApplicationLifecycleEvents {
```
marandaneto, line 178: same https://github.com/PostHog/posthog-ios/pull/309/files#r1967783078

PostHog/App Life Cycle/PostHogAppLifeCycleIntegration.swift
marandaneto: so we know if people tell its not capturing after reading the logs

PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -158,10 +198,84 @@
...
+            }
+
+            // capture necessary touch information on the main thread before performing any asynchronous operations
+            // - this ensures that UITouch associated objects like UIView, UIWindow, or [UIGestureRecognizer] are still valid.
+            // - these objects may be released or erased by the system if accessed asynchronously, resulting in invalid/zeroed-out touch coordinates
+            let touchInfo = touches.map {
+                (phase: $0.phase, location: $0.location(in: window))
+            }
+
+            PostHogReplayIntegration.dispatchQueue.async { [touchInfo, weak postHog = postHog] in
```
marandaneto, line 230: can `dispatchQueue` not be public/static now since its within the same class?

marandaneto:
> @ioannisj for `URLSessionExtension` you can just add an optional param to the methods, with a posthog instance, and use the shared instance if the param is null/not given if the extension can be used in objc, we need a new overload, though

same for `postHogScreenView` extension

PostHog/App Life Cycle/PostHogAppLifeCycleIntegration.swift
marandaneto: its just a flag rename suggestion

PostHog/PostHogIntegration.swift
```diff
@@ -0,0 +1,80 @@
...
+     * Note: This is typically called automatically during uninstallation
+     * but may be called manually to temporarily suspend the integration
+     * while maintaining its installation status (e.g manual start/stop for session recording)
+     */
+    func stop()
+}
+
+extension PostHogConfig {
+    /// Array of integrations to be installed based on current configuration
+    var integrations: [PostHogIntegration] {
```
marandaneto, line 55: should this just be part of the config object instead of an extension?

marandaneto (approved):
left a question otherwise LGTM

PR #312, opened 2025-02-27
fix: view masking when using clipShape view modifier

### LINKED ISSUE #296: Session Replay Mask text forces parent element to mask everything
### Version

3.19.4

### Steps to Reproduce

1. Mask a Text() element using `.postHogMask()`
2. The text element is inside a parent element which acts like a list item View
3. if the parent element has the `.clipShare` modifier, the whole element gets masked instead of just the text

### Expected Result

The inner Text view should be masked

Image

### Actual Result

The whole list item gets Masked

Image

ioannisj:

Closes #296 

`.clipShape` and `.clipped` SwiftUI view modifiers seem to be injecting a `SwiftUI._UIInheritedView` to the view hierarchy that has the size of the view (the parent view when masking in this case) 

When we manually tag views with `postHogNoCapture` in `postHogMask()` modifier, we were also tagging this wrapper view as well, which was then being masked by session replay. 

Current fix is to ignore `SwiftUI._UIInheritedView` types in postHogMask(). Future improvements to session replay could include testing if a view has visual impact before deciding to mask or visit children

| Before  | After   |  
|------|-----|
| |  |

Local sample project

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Next
 
 - fix: SwiftUI view masking when using clipShape view modifier ([#312](https://github.com/PostHog/posthog-ios/pull/312))
+- fix: reported crash on PostHogSessionManager ([#311](https://github.com/PostHog/posthog-ios/pull/311))
 
 ## 3.19.8 - 2025-02-26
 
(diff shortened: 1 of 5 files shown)


PostHog/SwiftUI/PostHogTagViewModifier.swift
```diff
@@ -175,7 +184,15 @@
 
         return commonAncestor
             .allDescendants(between: anchorView, and: taggerView)
-            .filter { !$0.postHogView } // exclude injected views
+            .lazy
```
marandaneto, line 187: what does lazy here do? we did not need before

marandaneto (approved):

PostHog/SwiftUI/PostHogTagViewModifier.swift
marandaneto: Oh nice, TIL, your call here.
The && is more readable but less functional so you can decide :)

PR #316, opened 2025-03-12
fix: storage not working on tvOS

### LINKED ISSUE #315: Apple TV support is broken
### Version

3.2.0

### Steps to Reproduce

Despite the documentation at https://posthog.com/docs/libraries/ios stating that Apple TV is supported (from tvOS 13 onward), on actually testing this I found out that it is not in fact supported:

1. Enable debug logging (config.debug = true)
2. Launch the Apple TV app with PostHog integrated
3. Observe error messages on the debug console such as: [PostHog] Error creating storage directory: Error Domain=NSCocoaErrorDomain Code=513 "You don’t have permission to save the file “phc_klFZ2lyYCUw7AoMcLfZ0UOxJJPoeKKmOCLARjbYDU3Y” in the folder “REDACTED”." UserInfo={NSFilePath=/var/mobile/Containers/Data/Application/3B250B5C-54D3-4DC6-9C0B-0B21DC24327E/Library/Application Support/REDACTED/phc_klFZ2lyYCUw7AoMcLfZ0UOxJJPoeKKmOCLARjbYDU3Y, NSURL=file:///var/mobile/Containers/Data/Application/3B250B5C-54D3-4DC6-9C0B-0B21DC24327E/Library/Application%20Support/REDACTED/phc_klFZ2lyYCUw7AoMcLfZ0UOxJJPoeKKmOCLARjbYDU3Y, NSUnderlyingError=0x3021dbe10 {Error Domain=NSPOSIXErrorDomain Code=1 "Operation not permitted"}}

### Expected Result

I have an app that supports iOS, macOS and Apple TV from a common codebase. The PostHog integration is handled via a common manager class with minimal changes required between the 3 platforms. For iOS and macOS, everything works as expected and I can see the events on the PostHog console. However, for Apple TV, I observe zero events with the OS set to Apple TV - which is expected as the PostHog SDK cannot write to the storage folder on Apple TV.

### Actual Result

As mentioned above, error messages on the debug console such as: [PostHog] Error creating storage directory: Error Domain=NSCocoaErrorDomain Code=513 "You don’t have permission to save the file “phc_klFZ2lyYCUw7AoMcLfZ0UOxJJPoeKKmOCLARjbYDU3Y” in the folder “REDACTED”." UserInfo={NSFilePath=/var/mobile/Containers/Data/Application/3B250B5C-54D3-4DC6-9C0B-0B21DC24327E/Library/Application Support/REDACTED/phc_klFZ2lyYCUw7AoMcLfZ0UOxJJPoeKKmOCLARjbYDU3Y, NSURL=file:///var/mobile/Containers/Data/Application/3B250B5C-54D3-4DC6-9C0B-0B21DC24327E/Library/Application%20Support/REDACTED/phc_klFZ2lyYCUw7AoMcLfZ0UOxJJPoeKKmOCLARjbYDU3Y, NSUnderlyingError=0x3021dbe10 {Error Domain=NSPOSIXErrorDomain Code=1 "Operation not permitted"}}

ioannisj:

Closes #315 

Local storage support on tvOS is [limited](https://developer.apple.com/library/archive/documentation/General/Conceptual/AppleTV_PG/), so we cannot use `Application Support` directory for storage. 
Unfortunately, the permission error is only visible when running on a physical device and not on simulators. 

This fix does the following: 
- Use `Library/Caches` for file queues
- Use lower defaults on tvOS for some SDK configs like flushAt to lower the chances of missing events

> tvOS only, other platforms should not be affected

Hasn't been fully tested since I don't have access to a physical device, but the user who reported this verified that switching to `Library/Caches` fixed the issue. 

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Next
 
 - fix: disk storage not working on tvOS ([#316](https://github.com/PostHog/posthog-ios/pull/316))
+- fix: wrong is_identified fallback value ([#317](https://github.com/PostHog/posthog-ios/pull/317))
 
 ## 3.20.0 - 2025-03-04
 
--- a/PostHog/PostHogStorageManager.swift
+++ b/PostHog/PostHogStorageManager.swift
@@ -95,7 +95,7 @@ public class PostHogStorageManager {
     public func isIdentified() -> Bool {
         identifiedLock.withLock {
             if isIdentifiedValue == nil {
-                isIdentifiedValue = storage.getBool(forKey: .isIdentified) ?? (getDistinctId() != getDistinctId())
+                isIdentifiedValue = storage.getBool(forKey: .isIdentified) ?? (getDistinctId() != getAnonymousId())
             }
         }
         return isIdentifiedValue ?? false
--- a/PostHogTests/PostHogStorageManagerTest.swift
+++ b/PostHogTests/PostHogStorageManagerTest.swift
@@ -11,9 +11,11 @@ import Nimble
 import Quick
 
 class PostHogStorageManagerTest: QuickSpec {
-    func getSut() -> PostHogStorageManager {
-        let config = PostHogConfig(apiKey: "123")
-        return PostHogStorageManager(config)
+    func getSut(_ config: PostHogConfig? = nil) -> PostHogStorageManager {
+        let theConfig = config ?? PostHogConfig(apiKey: "123")
+        let storage = PostHogStorage(theConfig)
+        storage.reset()
+        return PostHogStorageManager(theConfig)
     }
 
     override func spec() {
@@ -46,15 +48,29 @@ class PostHogStorageManagerTest: QuickSpec {
             sut.reset(true)
         }
 
-        it("Can can accept id customization via config") {
+        it("Can accept anon id customization via config") {
             let config = PostHogConfig(apiKey: "123")
             let fixedUuid = UUID.v7()
             config.getAnonymousId = { _ in fixedUuid }
-            let sut = PostHogStorageManager(config)
+            let sut = self.getSut(config)
             let anonymousId = sut.getAnonymousId()
             expect(anonymousId) == fixedUuid.uuidString
 
             sut.reset(true)
+        }
+
+        it("Uses the correct fallback value for isIdentified") {
+            let anonymousIdToSet = UUID.v7()
+            let distinctIdToSet = UUID.v7().uuidString
+
+            let config = PostHogConfig(apiKey: "123")
+            config.getAnonymousId = { _ in anonymousIdToSet }
+
+            let sut = self.getSut(config)
+            sut.setDistinctId(distinctIdToSet)
+
+            // Don't call setIdentified(true), isIdentified should be derived from different anon and distinct ids
+            expect(sut.isIdentified()) == true
         }
     }
 }


PostHog/PostHogConfig.swift
```diff
@@ -7,6 +7,20 @@
 import Foundation
 
 @objc(PostHogConfig) public class PostHogConfig: NSObject {
+    enum Defaults {
+        #if os(tvOS)
+            static let flushAt: Int = 5
+            static let maxQueueSize: Int = 100
+            static let maxBatchSize: Int = 5
+            static let flushIntervalSeconds: TimeInterval = 5
```
marandaneto, line 15: i think these 2 is fine as it is
the bigger the batch the better
the one is is a trade off of battery and resources, but we don't need more than 30s I think

PostHog/PostHogStorage.swift
```diff
@@ -38,10 +50,18 @@ func appGroupContainerUrl(config: PostHogConfig) -> URL? {
...
+        let url = FileManager.default
+            .containerURL(forSecurityApplicationGroupIdentifier: appGroupIdentifier)?
+            .appendingPathComponent("Library/Caches/")
+            .appendingPathComponent(bundleIdentifier)
+    #else
+        let url = FileManager.default
+            .containerURL(forSecurityApplicationGroupIdentifier: appGroupIdentifier)?
+            .appendingPathComponent("Library/Application Support/")
+            .appendingPathComponent(bundleIdentifier)
+    #endif
```
marandaneto, line 64: this can be a method, and we just pass the `appendingPathComponent` as param, so DRY

PostHog/PostHogStorage.swift
```diff
@@ -75,16 +95,30 @@ class PostHogStorage {
         case sessionReplay = "posthog.sessionReplay"
         case isIdentified = "posthog.isIdentified"
         case personProcessingEnabled = "posthog.enabledPersonProcessing"
+
+        var useFileStorage: Bool {
```
marandaneto, line 99: Not sure i follow why we need this

PostHog/PostHogConfig.swift
```diff
@@ -7,6 +7,20 @@
 import Foundation
 
 @objc(PostHogConfig) public class PostHogConfig: NSObject {
+    enum Defaults {
+        #if os(tvOS)
+            static let flushAt: Int = 5
+            static let maxQueueSize: Int = 100
+            static let maxBatchSize: Int = 50
+            static let flushIntervalSeconds: TimeInterval = 30
```
marandaneto, line 15: we are repeating both, can we just let it off the if/else branch

PostHog/PostHogStorage.swift
```diff
@@ -10,11 +10,24 @@ import Foundation
...
  */
 func applicationSupportDirectoryURL() -> URL {
-    let url = FileManager.default.urls(for: .applicationSupportDirectory, in: .userDomainMask).first!
+    #if os(tvOS)
+        // tvOS restricts access to Application Support directory on physical devices
+        // Use Library/Caches directory which may have less frequent eviction behavior than temp (which is purged when the app quits)
+        let url = FileManager.default.urls(for: .cachesDirectory, in: .userDomainMask).first!
+    #else
+        let url = FileManager.default.urls(for: .applicationSupportDirectory, in: .userDomainMask).first!
+    #endif
```
marandaneto, line 29: we can do similar to `librarySubPath` here
define the param with the if condition, so we don't repeat the whole `FileManager` line

PostHog/PostHogStorage.swift
marandaneto: I think its common that the OS clears up caching whenever is needed, but I don't think its common that it clears the cache every time you exit the app, is it?
if that's not the case, I'd just use the cache always, so we don't have to have this extra logic around `useFileStorage`

marandaneto: I also understand that using the `UserDefaults` would help persisting the distinct id whenever this happens, if it happens, so your call here

PostHog/PostHogStorage.swift
```diff
@@ -75,16 +95,30 @@ class PostHogStorage {
         case sessionReplay = "posthog.sessionReplay"
         case isIdentified = "posthog.isIdentified"
         case personProcessingEnabled = "posthog.enabledPersonProcessing"
+
+        var useFileStorage: Bool {
+            #if os(tvOS)
+                [.replayQeueue, .oldQeueue, .queue].contains(self)
```
marandaneto, line 102: i think more things should be here then because they can be heavy for user defaults (limit 500kb) eg flags, flags payloads, registerProperties, groups, we just cannot predict how much data would be because it really depends on how many flags and payloads they have ,etc

marandaneto (approved):
left a few suggestions and one point to discuss otherwise LGTM

PR #325, opened 2025-04-02
feat: add support for surveys on iOS

### LINKED ISSUE #168: Posthog Survey on iOS
### Problem Statement

Relates to https://github.com/PostHog/posthog-js-lite/issues/110

https://posthog.com/questions/record-survey-feedback

### Solution Brainstorm

_No response_

ioannisj:

Closes #168 
Docs: https://github.com/PostHog/posthog.com/pull/11155
Related: https://github.com/PostHog/posthog/issues/25085

Adds support for native iOS Surveys (Experimental)

| Feature                          					 	| Support                            				  |
| -----------------------------------| ---------------------------------- |
| **Questions**                     			   |						                                   |
| All question types                			   | ✅ 						                               |
| Multi-question surveys            		   | ✅						                               |
| Confirmation message             		   | ✅ 						                               |
| Feedback button presentation        | ❌                                					      |
| **Customization / Appearance**    |                                 						      |
| Set colors in PostHog dashboard   | ✅                                 				      |
| Shuffle questions                 			   | ❌                                  					  |
| PostHog branding   			               | ❌ 						                               |
| Delay popup after screen view       | ❌					                                   |
| Custom presentation 				      | ❌	    				                                   |
| Position config                   			  | ❌ Always bottom sheet		              |
| **Display conditions**            		  |                                    						 |
| Feature flag & property targeting  | ✅         					                          |
| Screen targeting                     		  | ❌                                						 |
| Device type targeting	                  | ✅                                                        |
| Survey wait period       		              | ❌                                                        |
| Event triggers                   				 | ✅						                                  |
| Action triggers                   			 | ❌						                                  |

- Unit tests
- Running surveys locally

## :pencil: Checklist


PostHog/Models/Survey.swift
```diff
@@ -0,0 +1,396 @@
...
+//
+//  Created by Yiannis Josephides on 20/01/2025.
+//
+
+import Foundation
+
+/// Represents the main survey object containing metadata, questions, conditions, and appearance settings.
+/// see: posthog-js/posthog-surveys-types.ts
+struct Survey: Decodable, Identifiable {
+    /// The unique identifier for the survey
```
marandaneto, line 13: 2 things here.
I think we should break into different files for most of the models.
Also, consider adding a PostHog prefix because of name clashing.

PostHog/PostHogRemoteConfig.swift
```diff
@@ -1,26 +1,40 @@
...
     private var loadingFeatureFlags = false
     private var sessionReplayFlagActive = false
-
     private var featureFlags: [String: Any]?
+
+    private var remoteConfigLock = NSLock()
+    private let loadingRemoteConfigLock = NSLock()
+    private var loadingRemoteConfig = false
+    private var remoteConfig: [String: Any]?
+    private var hasFeatureFlags: Bool?
```
marandaneto, line 27: do we need a global state for that? i think we can just check the return from the API and reload flags if there's any

PostHog/PostHogSDK.swift
```diff
@@ -599,6 +593,14 @@ let maxRetryDelay = 30.0
         )
 
         targetQueue?.add(posthogEvent)
+
+        #if os(iOS)
+            let surveysIntegration = installedIntegrations.compactMap {
+                $0 as? PostHogSurveyIntegration
+            }.first
+
+            surveysIntegration?.onEvent(event: posthogEvent.event)
```
marandaneto, line 602: i think its easier to just cache the instance as we do with others.

PostHog/PostHogRemoteConfig.swift
```diff
@@ -31,9 +45,85 @@ class PostHogFeatureFlags {
...
+        }
+    }
+
+    private func preloadFeatureFlags() {
+        featureFlagsLock.withLock {
+            // load disk cached config to memory
+            _ = getCachedFeatureFlags()
+        }
+
+        if config.preloadFeatureFlags {
```
marandaneto, line 75: we should add a config for the remote config API as well and fallback to the decide API in case the remote config is disabled

lucasheriques (approved):
i think everything looks mostly good! for the logic parts, we had already talked about them on past PRs so no further comments from me.

great job getting this to the finish line!!! ❤️

prob address Manoel's comments before merging. I checked mostly the displaying logic and events. Can't comment much on the code style

PostHog/PostHogRemoteConfig.swift
marandaneto: If someone calls `reloadFeatureFlags` manually, the intention is to fetch the API because something might have changed. In this case, we should not check what's cached in memory.

PostHog/PostHogSDK.swift
marandaneto: i think this can be a problem with multiple sdk instances though, unless you allow `surveysIntegration` to be installed for all instances, otherwise you might be calling the wrong one

marandaneto: yeah i guess so, worth adding a comment on the posthog config (surveys: Boolean) though.

PostHog/PostHogRemoteConfig.swift
marandaneto: same on RN, we just need to add a comment that disabling this would disable surveys.
since this is an experimental feature (remote config), we need a way to disable it and fallback to decide API for everything else

marandaneto (approved):

PostHog/PostHogRemoteConfig.swift
```diff
@@ -120,42 +227,27 @@ class PostHogFeatureFlags {
...
-                        // keeps the value from config.sessionReplay since having sessionRecording
-                        // means its enabled on the project settings, but its only enabled
-                        // when local replay integration is enabled/active
-                        if let endpoint = sessionRecording["endpoint"] as? String {
-                            self.config.snapshotEndpoint = endpoint
-                        }
-                        self.sessionReplayFlagActive = self.isRecordingActive(featureFlags, sessionRecording)
-                        self.storage.setDictionary(forKey: .sessionReplay, contents: sessionRecording)
-                    }
+                    self.processSessionRecordingConfig(data, featureFlags: featureFlags)
```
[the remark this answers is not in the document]
marandaneto: not sure i follow.

app starts - read from cache
if remote config is enabled, fetch remote config and update cache and global state if any
if remote config is disabled and preload flags is enabled, fetch decide and update cache and global state if any

until one of the requests are completed and the cache is updated, session recording will be used the value from the cache

PR #327, opened 2025-04-08
feat:  Add more details such as version, id, and reason to $feature_flag_called events

haacked:

Add support for `/decide?v=4` (see https://github.com/PostHog/posthog/pull/29751). Similar to how it's done in https://github.com/PostHog/posthog-js-lite/pull/427.

## Updates to the `$feature_flag_called` event

When capturing the `$feature_flag_called` event, additional information are now captured:

The end result is `$feature_flag_called` events will have additional properties:

| Property Name                |  Descirption
| ------------------------ | -------------
| `$feature_flag_version` | The version of the feature flag. This is visible in the diffs of the feature flag activity log.
| `$feature_flag_reason` | The reason the feature flag matched or didn't match.
| `$feature_flag_id`          | The database id of the feature flag (Just in case a flag was deleted and then recreated with the same key, we can determine the difference.

## Backwards compatibility:

The changes are all backwards compatible with `/decide?v=3`.

Uni tests. I don't know how to manually test it yet.

## :pencil: Checklist


PostHog/PostHogFeatureFlags.swift
```diff
@@ -17,12 +17,18 @@ class PostHogFeatureFlags {
     private var loadingFeatureFlags = false
     private var sessionReplayFlagActive = false
 
+    private var flags: [String: Any]?
```
marandaneto, line 20: can we just use the `featureFlags` and `featureFlagPayloads`?

PostHog/PostHogSDK.swift
```diff
@@ -887,10 +887,26 @@ let maxRetryDelay = 30.0
...
 
         if shouldCapture {
-            let properties = [
+            let requestId = featureFlags?.lastRequestId ?? ""
+            let details = featureFlags?.getFeatureFlagDetails(flagKey)
+
+            var properties = [
                 "$feature_flag": flagKey,
                 "$feature_flag_response": flagValue ?? "",
+                "$feature_flag_request_id": requestId,
```
marandaneto, line 896: we do set `"$feature_flag_response"` if the value is nil with an empty string.
now we do the same with `$feature_flag_request_id`, maybe we should just not send those values

PostHog/PostHogStorage.swift
```diff
@@ -87,12 +87,14 @@ class PostHogStorage {
         case replayQeueue = "posthog.replayFolder"
         case enabledFeatureFlags = "posthog.enabledFeatureFlags"
         case enabledFeatureFlagPayloads = "posthog.enabledFeatureFlagPayloads"
+        case flags = "posthog.flags"
```
marandaneto, line 90: can we reuse the `enabledFeatureFlags` and `enabledFeatureFlagPayloads`? otherwise there will be cached values that will be outdated, probably not deleted unless the user call `reset`, and the app will start without having cached flags but they do under a different storage name.
is the v4 data structure different so that we have to use a new storage key?

haacked: The [decide v4 response](https://posthog.com/docs/api/decide) is very different. It contains more metadata about a flag and how it got evaluated. That information is not in `enabledFeatureFlags` and `enabledFeatureFlagPayloads`.

I wrote this in a back-compat manner, so we still use `enabledFeatureFlags` and `enabledFeatureFlagPayloads` when calling `isFeatureEnabled` etc.

The main difference is now when we capture a `$feature_flag_called` event, we check to see if there's anything in `posthog.flags` and if so, we use that to add extra metadata to the event.

In a future follow-up, we could even remove `posthog.enabledFeatureFlags` and `posthog.enabledFeatureFlagPayloads` and calculate both of those from `posthog.flags`. I didn't want to do that right now because I wanted to make the minimal change possible.

PostHog/PostHogSDK.swift
haacked: > we do set "$feature_flag_response" if the value is nil with an empty string.

Oh, that's actually incorrect. We want to know if the value is `nil` because that means something. I'll fix that.

PostHog/PostHogStorage.swift
marandaneto: oh i see, thanks for the context

PostHog/PostHogRemoteConfig.swift
```diff
@@ -36,6 +39,10 @@ class PostHogRemoteConfig {
     private let dispatchQueue = DispatchQueue(label: "com.posthog.RemoteConfig",
                                               target: .global(qos: .utility))
 
+    var lastRequestId: String? {
+        requestId ?? storage.getString(forKey: .requestId)
```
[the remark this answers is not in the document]
haacked: Done!

haacked:
Thanks for the review @ioannisj! I made the lock change. Before I merge, I wanted to double check that was what you had in mind.

PR #329, opened 2025-04-10
feat: start and stop integrations on opt out

### LINKED ISSUE #280: Stop all integrations like session replay and autocapture on opt-out
### Description

Original discussion: https://github.com/PostHog/posthog-ios/pull/276#discussion_r1905431037

When opting-out, we should `stop()` all integrations since they won't be processed by the SDK, and then `start()` again on next opt-in. 

For session replay, when opting-in a new session should be created.

- [x] iOS
- [ ] Android
- [ ] React Native
- [ ] Flutter

ioannisj:

Closes #280 

Start and stop integrations on `optIn()` and `optOut()` calls

CleanShot 2025-04-10 at 15 56 50

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## Next
 
-- feat: stop integrations on optOut() ([#329](https://github.com/PostHog/posthog-ios/pull/329))
+- feat: start/stop integrations when calling optIn() or optOut() ([#329](https://github.com/PostHog/posthog-ios/pull/329))
 
 ## 3.22.1 - 2025-04-10
 


pauldambra (approved):
going off the gif more than anything else considering my swift knowledge 🙈 😅

pauldambra:
oh! one thing js does maybe badly here... when some opts out will this finish flushing whatever is already in the queue?

in JS we drop anything not already sent and some people want to stop recording manually but it means we stop at a random point sometime before the opt out

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## Next
 
+- feat: stop integrations on optOut() ([#329](https://github.com/PostHog/posthog-ios/pull/329))
```
marandaneto, line 3: maybe be specific about resuming integration on optIn as well?

PostHog/PostHogSDK.swift
```diff
@@ -898,6 +901,10 @@ let maxRetryDelay = 30.0
             config.optOut = false
             storage?.setBool(forKey: .optOut, contents: false)
         }
+
+        setupLock.withLock {
+            installIntegrations()
```
marandaneto, line 906: will integrations capture dupe events/or do dupe things if its installed multiple times? eg https://github.com/PostHog/posthog-ios/blob/506e40a3a873fdf8dd8bceebd07e999732196502/PostHog/App%20Life%20Cycle/PostHogAppLifeCycleIntegration.swift#L61-L70

PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -23,7 +23,6 @@
 
         private weak var postHog: PostHogSDK?
 
-        private var timer: Timer?
```
marandaneto, line 26: is there timer a left over from the other PR?

marandaneto (approved):

marandaneto:
@ioannisj i think `sessionReplay` should be for the automatic recording, you can add a comment to the toggle if needed.
but if its disabled, and you call `startSessionRecording` it should work (unless its disabled server side).
It's how its documented here as well https://posthog.com/docs/session-replay/installation?tab=iOS#manually-control-session-recordings
so its correct IMO.
You'd need to fix the code to allow calling `startSessionRecording` even if `sessionReplay` is disabled, but respecting the remote config.

pauldambra:
i think if we can stay as close to the JS SDK behaviour as possible then it makes everything simpler (even if we'd maybe not pick that today 😅)

so init disabled and then start recording should be supported 👍

PR #337, opened 2025-04-18
fix: Reimplement `isfeatureEnabled` using `getFeatureFlag`

### LINKED ISSUE #336: `isFeatureEnabled` should send the variant values of a multivariate flag instead of the boolean values
### Version

3.24.0

### Steps to Reproduce

- Create a multivariate feature flag with a 50% split between the two variants (a classic A/B test flag)
- Instrument it in code with `isFeatureEnabled`
- notice that it sends boolean values for `$feature_flag_response` instead of the variant values.

### Expected Result

- it should send the variant values for the `$feature_flag_response` when evaluating a  multivariate flag, regardless if it's `isFeatureEnabled` or `getFeatureFlag`

All of our other SDKs do this, so the iOS SDK should as well.

### Actual Result

See notes above.

haacked:
Changes `isFeatureEnabled` to delegate to `getFeatureFlag` in order to get the result. This ensures that the `$feature_flag_called` event sends the correct `$feature_flag_response`.

Also added some bin scripts: See PostHog/meta#309 for context.

Fixes #336

Unit tests

## :pencil: Checklist

### DIFF
--- a/PostHog/PostHogSDK.swift
+++ b/PostHog/PostHogSDK.swift
@@ -844,21 +844,8 @@ let maxRetryDelay = 30.0
     }
 
     @objc public func isFeatureEnabled(_ key: String) -> Bool {
-        if !isEnabled() {
-            return false
-        }
-
-        guard let remoteConfig else {
-            return false
-        }
-
-        let value = remoteConfig.isFeatureEnabled(key)
-
-        if config.sendFeatureFlagEvent {
-            reportFeatureFlagCalled(flagKey: key, flagValue: value)
-        }
-
-        return value
+        let result = getFeatureFlag(key)
+        return result is String ? true : (result as? Bool) ?? false
     }
 
     @objc public func getFeatureFlagPayload(_ key: String) -> Any? {
--- a/PostHogTests/PostHogSDKTest.swift
+++ b/PostHogTests/PostHogSDKTest.swift
@@ -246,6 +246,29 @@ class PostHogSDKTest: QuickSpec {
             sut.close()
         }
 
+        it("send feature flag event with variant response for isFeatureEnabled when enabled") {
+            let sut = self.getSut(preloadFeatureFlags: true, sendFeatureFlagEvent: true)
+
+            waitDecideRequest(server)
+            expect(sut.isFeatureEnabled("string-value")) == true
+
+            let events = getBatchedEvents(server)
+
+            expect(events.count) == 1
+
+            let event = events.first!
+            expect(event.event) == "$feature_flag_called"
+            expect(event.properties["$feature_flag"] as? String) == "string-value"
+            expect(event.properties["$feature_flag_response"] as? String) == "test"
+            expect(event.properties["$feature_flag_request_id"] as? String) == "0f801b5b-0776-42ca-b0f7-8375c95730bf"
+            expect(event.properties["$feature_flag_id"] as? Int) == 3
+            expect(event.properties["$feature_flag_version"] as? Int) == 1
+            expect(event.properties["$feature_flag_reason"] as? String) == "Matched condition set 1"
+
+            sut.reset()
+            sut.close()
+        }
+
         it("send feature flag event for getFeatureFlag when enabled") {
             let sut = self.getSut(preloadFeatureFlags: true, sendFeatureFlagEvent: true)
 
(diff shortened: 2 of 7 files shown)


haacked:
Once this is reviewed, I'll bump the changelog and version.

PostHog/PostHogSDK.swift
```diff
@@ -844,21 +844,8 @@ let maxRetryDelay = 30.0
...
-
-        let value = remoteConfig.isFeatureEnabled(key)
-
-        if config.sendFeatureFlagEvent {
-            reportFeatureFlagCalled(flagKey: key, flagValue: value)
-        }
-
-        return value
+        let result = getFeatureFlag(key)
+        return result is String ? true : (result as? Bool) ?? false
```
marandaneto, line 848: `getFeatureFlag` returns `Any`, so it can be something different than String and Boo,l and in this ca,se it'd evaluate to `fal,se` which is wrong

marandaneto: the correct check would be like this https://github.com/PostHog/posthog-ios/blob/1b011434301250ca75098c40fecad9ba3d209307/PostHog/PostHogRemoteConfig.swift#L384-L393

marandaneto: Also since you dont use `isFeatureEnabled` from `PostHogRemoteConfig` anymore, maybe we should remove it?

haacked: It's too bad Swift doesn't have union types. In practice, a feature flag result will only be `null`, `bool`, or a `string`. If `getFlagValue` returned something other than one of those types, that would be unexpected. In that case, I think returning `false` is appropriate OR we should throw an exception. What do you think?

marandaneto: > a feature flag result will only be null, bool, or a string

TIL, so is numbers stringified?
In my opinion, we should never throw an exception; otherwise, clients would always need to wrap their code in try/catch.

marandaneto: btw Android does the same so would need fixing https://github.com/PostHog/posthog-android/blob/d269ddb7a1e07590b00d68cd04b62d1b87716554/posthog/src/main/java/com/posthog/PostHog.kt#L737-L765

haacked: > TIL, so is numbers stringified?

Not really, because there are no numbers for flag results.

We have two types of flags. Boolean and multivariate. A boolean flag result is always a boolean. A variant flag result is the variant key, which is a string.

For example, if you create a multivariate flag, and enter a number as a key, it's stored like so:

```json
{
  "groups": [...],
  "payloads": {},
  "multivariate": {
    "variants": [
      {
        "key": "1234",
        "name": "",
        "rollout_percentage": 0
      }
    ]
  }
}
```

> btw Android does the same so would need fixing [PostHog/posthog-android@d269ddb/posthog/src/main/java/com/posthog/PostHog.kt#L737-L765](https://github.com/PostHog/posthog-android/blob/d269ddb7a1e07590b00d68cd04b62d1b87716554/posthog/src/main/java/com/posthog/PostHog.kt#L737-L765)

Thanks! I was planning to go through every client to see if any others have this issue.

Are you OK with the current changes or is there anything else you'd like to see changed?

marandaneto: thanks for the context.
@haacked https://github.com/PostHog/posthog-ios/pull/337#discussion_r2051402773, just this comment, IMO, so we can delete code that isn't used anymore.

CHANGELOG.md
```diff
@@ -1,4 +1,6 @@
-## Next
+## 3.24.1 - 2025-04-18
```
marandaneto, line 1:
```suggestion
## 3.24.1 - 2025-04-22
```

marandaneto (approved):
2 comments left otherwise LGTM
@ioannisj also take a look so you are aware of the changes

PostHog/PostHogSDK.swift
haacked: Oh yeah! I somehow missed that comment. I'll take care of it.

haacked: Done!

CHANGELOG.md
```diff
@@ -1,4 +1,6 @@
-## Next
+## 3.24.1 - 2025-04-22
```
marandaneto, line 1: @ioannisj should have been 2025-04-`23` here

PR #340, opened 2025-04-22
fix: swizzle viewDidLayoutSubviews instead of layoutSubviews

### LINKED ISSUE #333: iPad floating keyboard hang
### Version

3.23.0

### Steps to Reproduce

Hi, I have a problem on iPad with floating keyboard. When i click to textField (does not mather if in UIKit or SwiftUI) then application hang, CPU is on 100% and memory continue to rise. Sometimes it happens on second text field, when keyboard changes from text input to numberPad.
I can see these types of error in debug console 

Image
Image

I manage once to pause the app and it was stuck on `func sendEvent(_ event: UIEvent)` 
When I dissable PostHog, everythink works correctly.

### Expected Result

Application should not freeze

### Actual Result

Application freezes

ioannisj:

Closes #333 

It looks like swizzling UIView.layoutSubviews is causing an infinite loop when running on a real iPad with the floating keyboard enabled. Interestingly, this issue doesn’t occur in the simulator, which made it harder to diagnose initially. I tried swizzling with an empty implementation method  just to confirm if recursion was the issue, but even then, I couldn’t prevent the loop — which is quite odd.

I came across this [related discussion](https://github.com/Juanpe/SkeletonView/issues/352#issuecomment-755866354) in the SkeletonView GitHub repo that seems to validate that this is indeed happening. The fix there was to properly unswizzle in their case, which doesn't fit our use case. 

I'm still in the dark on what the root causing may be, but a potential fix is to avoid `UIView.layoutSubviews` and instead rely on `UIViewController.viewDidLayoutSubviews` to detect view layouts. 

Tested locally on a real device

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 ## Next
 
 - fix: app hangs on iPad with floating keyboard when session replay is enabled ([#340](https://github.com/PostHog/posthog-ios/pull/340))
+
+## 3.24.1 - 2025-04-23
+
+- fix: Send correct `$feature_flag_response` for the `$feature_flag_called` event when calling `isFeatureEnabled` ([#337](https://github.com/PostHog/posthog-ios/pull/337))
 - fix: support ISO8601 dates with missing milliseconds ([#338](https://github.com/PostHog/posthog-ios/pull/338))
 
 ## 3.24.0 - 2025-04-17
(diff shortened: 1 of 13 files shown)


marandaneto:
@ioannisj just make sure that everything that depends on the former still works (eg session replay, screen view events, etc).
i'd test on react native as well just to be sure.

marandaneto (approved):

marandaneto (approved):

Issue #344, opened 2025-04-28
Feature Flags Always Return False


reddavis:
### Version

3.24.2

### Steps to Reproduce

1. setup a feature flag
2. set it to 100%

![Image](https://github.com/user-attachments/assets/a6040bb7-c2c6-4bf7-8b63-e6d1dc7aafb1)

4. call `isFeatureEnabled` and see it returns false.
5. go to the PostHog FF dashboard, and see that its showing it returned false for the user.

![Image](https://github.com/user-attachments/assets/bbd7a80c-0930-4d84-a46a-14d21399045f)

6. use an older version of the SDK (i went back to 3.20.0)
7. call `isFeatureEnabled` and see it returns true.
5. go to the PostHog FF dashboard, and see that its showing it returned true for the user.

![Image](https://github.com/user-attachments/assets/4062b808-1593-4044-bee4-5461b998be5c)

### Expected Result

Expected to return true

### Actual Result

It returns false

marandaneto:
@ioannisj can you debug this? maybe one of the latest changes broke something

bb-git:
Actually in 3.30.0 it is also not working (still or again). I downgraded to 3.20.0 and it was working.

PR #345, opened 2025-04-28
chore(flags): use new `/flags` endpoint instead of `/decide`

dmarticus:

Update the iOS SDK to use the new `/flags` endpoint instead of the legacy `/decide` one.

### DIFF
--- a/PostHog/PostHogApi.swift
+++ b/PostHog/PostHogApi.swift
@@ -192,18 +192,18 @@ class PostHogApi {
         }.resume()
     }
 
-    func decide(
+    func flags(
         distinctId: String,
         anonymousId: String?,
         groups: [String: String],
         completion: @escaping ([String: Any]?, _ error: Error?) -> Void
     ) {
         guard let url = getEndpointURL(
-            "/decide",
-            queryItems: URLQueryItem(name: "v", value: "4"),
+            "/flags",
+            queryItems: URLQueryItem(name: "v", value: "2"),
             relativeTo: config.host
         ) else {
-            hedgeLog("Malformed decide URL error.")
+            hedgeLog("Malformed flags URL error.")
             return completion(nil, nil)
         }
 
@@ -226,13 +226,13 @@ class PostHogApi {
         do {
             data = try JSONSerialization.data(withJSONObject: toSend)
         } catch {
-            hedgeLog("Error parsing the decide body: \(error)")
+            hedgeLog("Error parsing the flags body: \(error)")
             return completion(nil, error)
         }
 
         URLSession(configuration: config).uploadTask(with: request, from: data!) { data, response, error in
             if error != nil {
-                hedgeLog("Error calling the decide API: \(String(describing: error))")
+                hedgeLog("Error calling the flags API: \(String(describing: error))")
                 return completion(nil, error)
             }
 
@@ -240,20 +240,20 @@ class PostHogApi {
 
             if !(200 ... 299 ~= httpResponse.statusCode) {
                 let jsonBody = String(describing: try? JSONSerialization.jsonObject(with: data!, options: .allowFragments) as? [String: Any])
-                let errorMessage = "Error calling decide API: status: \(httpResponse.statusCode), body: \(jsonBody)."
+                let errorMessage = "Error calling flags API: status: \(httpResponse.statusCode), body: \(jsonBody)."
                 hedgeLog(errorMessage)
 
                 return completion(nil,
                                   InternalPostHogError(description: errorMessage))
             } else {
-                hedgeLog("Decide called successfully.")
+                hedgeLog("Flags called successfully.")
             }
 
             do {
                 let jsonData = try JSONSerialization.jsonObject(with: data!, options: .allowFragments) as? [String: Any]
                 completion(jsonData, nil)
             } catch {
-                hedgeLog("Error parsing the decide response: \(error)")
+                hedgeLog("Error parsing the flags response: \(error)")
                 completion(nil, error)
             }
         }.resume()
--- a/PostHog/PostHogRemoteConfig.swift
+++ b/PostHog/PostHogRemoteConfig.swift
@@ -237,7 +237,7 @@ class PostHogRemoteConfig {
             self.loadingFeatureFlags = true
         }
 
-        api.decide(distinctId: distinctId,
+        api.flags(distinctId: distinctId,
                    anonymousId: anonymousId,
                    groups: groups)
         { data, _ in
@@ -260,7 +260,7 @@ class PostHogRemoteConfig {
 
                 // Safely handle optional data
                 guard var data = data else {
-                    hedgeLog("Error: Decide response data is nil")
+                    hedgeLog("Error: Flags response data is nil")
                     self.notifyFeatureFlagsAndRelease(data)
                     return callback(nil)
                 }
@@ -282,7 +282,7 @@ class PostHogRemoteConfig {
                 guard let featureFlags = data["featureFlags"] as? [String: Any],
                       let featureFlagPayloads = data["featureFlagPayloads"] as? [String: Any]
                 else {
-                    hedgeLog("Error: Decide response missing correct featureFlags format")
+                    hedgeLog("Error: Flags response missing correct featureFlags format")
 
                     self.notifyFeatureFlagsAndRelease(data)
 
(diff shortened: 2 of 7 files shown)


PostHog.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name             = "PostHog"
-  s.version          = "3.24.2"
+  s.version          = "3.25.0"
```
marandaneto, line 3: versioning is done by the release process https://github.com/PostHog/posthog-ios/blob/main/RELEASING.md

PostHog/PostHogVersion.swift
```diff
@@ -9,7 +9,7 @@ import Foundation
 
 // if you change this, make sure to also change it in the podspec and check if the script scripts/bump-version.sh still works
 // This property is internal only
-public var postHogVersion = "3.24.2"
+public var postHogVersion = "3.25.0"
```
marandaneto, line 12: same as https://github.com/PostHog/posthog-ios/pull/345/files#r2065547767

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## Next
 
+## 3.25.0 – 2025-04-29
```
marandaneto, line 3: similar to https://github.com/PostHog/posthog-ios/pull/345/files#r2065547767
thats why we have the 'Next'

marandaneto:
@ioannisj can you review it?
Also, I'd debug https://github.com/PostHog/posthog-ios/issues/344 first since its a different endpoint now (and likely easier to reproduce if there's a bug)

PostHog.podspec
dmarticus: got it, will amend

dmarticus:
@ioannisj I fixed the issue with https://github.com/PostHog/posthog/pull/31798, can you give it a try on your side again?

marandaneto:
similar to https://github.com/PostHog/posthog-android/pull/245#issuecomment-2893824253 but here https://github.com/PostHog/posthog-ios/blob/479ce5ba9ce899dfe969dea7c88cd110adb800e7/PostHog/PostHogRemoteConfig.swift#L187C18-L187C35 and probably more places, @ioannisj would know more

dmarticus:
@ioannisj with this commit [`ce2f2ac` (#345)](https://github.com/PostHog/posthog-ios/pull/345/commits/ce2f2ac85475576600e131a69c39aacc0f87ce52) I've added support for all of the `/decide` fields to the `/flags` endpoint, how does this look to you?

PostHog/PostHogApi.swift
```diff
@@ -192,18 +192,29 @@ class PostHogApi {
...
         completion: @escaping ([String: Any]?, _ error: Error?) -> Void
     ) {
-        guard let url = getEndpointURL(
-            "/decide",
-            queryItems: URLQueryItem(name: "v", value: "4"),
-            relativeTo: config.host
-        ) else {
-            hedgeLog("Malformed decide URL error.")
+        let url: URL?
+        if config.remoteConfig {
```
dmarticus: I don't need to load the rest of the config if we're using RemoteConfig

PostHog/PostHogApi.swift
```diff
@@ -192,18 +192,29 @@ class PostHogApi {
...
+                queryItems: URLQueryItem(name: "v", value: "2"),
+                relativeTo: config.host
+            )
+        } else {
+            url = getEndpointURL(
+                "/flags",
+                queryItems: URLQueryItem(name: "v", value: "2"), URLQueryItem(name: "config", value: "true"),
+                relativeTo: config.host
+            )
+        }
```
dmarticus: btw i fully used Claude on this, not sure if it's idiomatic.

PostHog/PostHogApi.swift
ioannisj: Hey @dmarticus, I think this won't fix the session replays that depend on flags issue. We do fetch `sessionRecording` from remote config to get `linkedFlag` value, but the way the code is currently structured the check expects both `sessionRecording` **and** feature flag values to be present when calling [isRecordingActive](https://github.com/PostHog/posthog-ios/blob/d6fde6cea4fbb2f968f116aa04572c39297a8c2b/PostHog/PostHogRemoteConfig.swift#L192)

As it stands, session recording will only be evaluated from remote config payload (if config.remoteConfig is true), but it will always do so with cached flags (from a previous app run). 

1. enable session recording flag on dashboard
2. first run, session recording evaluates to false (no cached flags) 
3. second run, session recording evaluates to true (from previously cached flags) 
4. turn off session recording flag on dashboard
5. app run, session recording evaluates to true (from previously cached flags) 
6. additional run, session recording correctly evaluates to false

So here we can either: 
1. restructure the code so that we cache `linkedFlag` from remote config call and re-evaluate when /flags is called with fresh flag values
2. always call with `/flags&config=true` so that we get the whole response for now

cc: @marandaneto can fact check me on this

dmarticus: okay, let me see if I follow.  the flow goes:

`initAfterStorage`, which does this

```
      if (this._disableRemoteConfig === false) {
        this.reloadRemoteConfigAsync()
      } else {
        this.logMsgIfDebug(() => console.info('PostHog Debug', `Remote config is disabled.`))
        if (options?.preloadFeatureFlags !== false) {
          this.logMsgIfDebug(() => console.info('PostHog Debug', `Feature flags will be preloaded from Flags API.`))
          // if RemoteConfig is disabled, we need to use the `/flags?v=2&config=true` endpoint to get the feature flags and the various other configs
          this.reloadFeatureFlags({ withConfig: true })
        } else {
          this.logMsgIfDebug(() => console.info('PostHog Debug', `preloadFeatureFlags is disabled.`))
        }
      }
```

so it either loads the remote config, or loads the feature flags + all the other config options that were originally on `/decide`.

`reloadRemoteConfigAsync` goes like this
```
  private async _remoteConfigAsync(): Promise<PostHogRemoteConfig | undefined> {
    this._remoteConfigResponsePromise = this._initPromise
      .then(() => {
        let remoteConfig = this.getPersistedProperty<Omit<PostHogRemoteConfig, 'surveys'>>(
          PostHogPersistedProperty.RemoteConfig
        )

        this.logMsgIfDebug(() => console.log('PostHog Debug', 'Cached remote config: ', JSON.stringify(remoteConfig)))

        return super.getRemoteConfig().then((response) => {
          if (response) {
            const remoteConfigWithoutSurveys = { ...response }
            delete remoteConfigWithoutSurveys.surveys

            this.logMsgIfDebug(() =>
              console.log('PostHog Debug', 'Fetched remote config: ', JSON.stringify(remoteConfigWithoutSurveys))
            )

            if (this.disableSurveys === false) {
              const surveys = response.surveys

              let hasSurveys = true

              if (!Array.isArray(surveys)) {
                // If surveys is not an array, it means there are no surveys (its a boolean instead)
                this.logMsgIfDebug(() => console.log('PostHog Debug', 'There are no surveys.'))
                hasSurveys = false
              } else {
                this.logMsgIfDebug(() =>
                  console.log('PostHog Debug', 'Surveys fetched from remote config: ', JSON.stringify(surveys))
                )
              }

              if (hasSurveys) {
                this.setPersistedProperty<SurveyResponse['surveys']>(
                  PostHogPersistedProperty.Surveys,
                  surveys as Survey[]
                )
              } else {
                this.setPersistedProperty<SurveyResponse['surveys']>(PostHogPersistedProperty.Surveys, null)
              }
            } else {
              this.setPersistedProperty<SurveyResponse['surveys']>(PostHogPersistedProperty.Surveys, null)
            }
            // we cache the surveys in its own storage key
            this.setPersistedProperty<Omit<PostHogRemoteConfig, 'surveys'>>(
              PostHogPersistedProperty.RemoteConfig,
              remoteConfigWithoutSurveys
            )

            this.cacheSessionReplay('remote config', response)

            // we only dont load flags if the remote config has no feature flags
            if (response.hasFeatureFlags === false) {
              // resetting flags to empty object
              this.setKnownFeatureFlagDetails({ flags: {} })

              this.logMsgIfDebug(() => console.warn('Remote config has no feature flags, will not load feature flags.'))
            } else if (this.preloadFeatureFlags !== false) {
              this.reloadFeatureFlags()
            }

            if (!response.supportedCompression?.includes(Compression.GZipJS)) {
              this.disableCompression = true
            }

            remoteConfig = response
          }

          return remoteConfig
        })
      })
      .finally(() => {
        this._remoteConfigResponsePromise = undefined
      })
    return this._remoteConfigResponsePromise
  }
  ```
  
  which, if I'm understanding correctly, loads the session recording info from remote config and then loads fresh flags via `reloadFeatureFlags` if the remote config says to load feature flags.

So to me, it looks like all the data is set after `initAfterStorage` completes.  Am I missing something?

PR #347, opened 2025-04-29
feat: clear cached flags when hasFeatureFlags is false

ioannisj:

Clear cached feature flags when remote config responds with `hasFeatureFlags false`

## :pencil: Checklist

### DIFF
--- a/PostHog/PostHogRemoteConfig.swift
+++ b/PostHog/PostHogRemoteConfig.swift
@@ -28,7 +28,6 @@ class PostHogRemoteConfig {
     private var remoteConfig: [String: Any]?
     private var remoteConfigDidFetch: Bool = false
     private var featureFlagPayloads: [String: Any]?
-    private let requestIdLock = NSLock()
     private var requestId: String?
 
     /// Internal, only used for testing
@@ -41,7 +40,7 @@ class PostHogRemoteConfig {
                                               target: .global(qos: .utility))
 
     var lastRequestId: String? {
-        requestIdLock.withLock {
+        featureFlagsLock.withLock {
             requestId ?? storage.getString(forKey: .requestId)
         }
     }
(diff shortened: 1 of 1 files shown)


PostHog/PostHogRemoteConfig.swift
```diff
@@ -247,12 +253,8 @@ class PostHogRemoteConfig {
                 {
                     // swiftlint:disable:next line_length
                     hedgeLog("Warning: Feature flags quota limit reached - clearing all feature flags and payloads. See https://posthog.com/docs/billing/limits-alerts for more information.")
-                    self.featureFlagsLock.withLock {
-                        // Clear both feature flags and payloads
-                        self.setCachedFeatureFlags([:])
-                        self.setCachedFeatureFlagPayload([:])
-                    }
 
+                    self.clearFeatureFlags()
```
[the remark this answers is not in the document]
marandaneto: that was added by @dmarticus in multiple SDKs so I believe so

PostHog/PostHogRemoteConfig.swift
```diff
@@ -483,6 +485,14 @@ class PostHogRemoteConfig {
         }
     }
 
+    private func clearFeatureFlags() {
```
marandaneto, line 488: I believe we should reset cache/var `requestId` here too

PostHog/PostHogRemoteConfig.swift
```diff
@@ -76,10 +76,16 @@ class PostHogRemoteConfig {
...
                     guard let self else { return }
+
                     let hasFeatureFlags = remoteConfig?[self.hasFeatureFlagsKey] as? Bool == true
-                    let preloadFeatureFlags = self.config.preloadFeatureFlags
-                    // reload feature flags if not previously loaded
-                    if hasFeatureFlags, preloadFeatureFlags {
+
+                    // if server responds with `hasFeatureFlags: false`, then there are no more active flags on the account
+                    guard hasFeatureFlags else {
+                        return clearFeatureFlags()
```
marandaneto, line 84: i think we should also set `loadingFeatureFlags: true` here since its kinda loaded, we know there's nothing server side

marandaneto:
always tag @PostHog/team-feature-flags team for feature flag reviews as well.

PostHog/PostHogRemoteConfig.swift
marandaneto: oh yeah my bad, i meant the same as on Android `isFeatureFlagsLoaded` but I've mistaken the var name, we don't use it here.
so IMO we should notify though because its the same as loaded, we know there's nothing, and we're intentionally not calling the decide API, which means the app might be 'awaiting' for the callback/notification and it'll never arrive.

PostHog/PostHogRemoteConfig.swift
```diff
@@ -450,6 +461,16 @@ class PostHogRemoteConfig {
         return value
     }
 
+    // To be called after acquiring `requestIdLock`
```
marandaneto, line 464: can we actually reuse `featureFlagsLock` instead? `requestId` is only set/cleared within the flags context anyway

PostHog/PostHogRemoteConfig.swift
```diff
@@ -483,6 +504,17 @@ class PostHogRemoteConfig {
...
 
+    private func clearFeatureFlags() {
+        featureFlagsLock.withLock {
+            setCachedFlags([:])
+            setCachedFeatureFlags([:])
+            setCachedFeatureFlagPayload([:])
+        }
+        requestIdLock.withLock {
+            setCachedRequestId(nil)
+        }
```
marandaneto, line 515: this https://github.com/PostHog/posthog-ios/pull/347/files#r2108998573

PR #349, opened 2025-05-07
fix: session manager not rotating session when moving from background

ioannisj:

Edge case where session ID is not rotated for session replay when user moves from background to foreground, after max idle time. 

The order of getting session id (which runs the checks for inactivity) and `touchSession()` (which marks last activity) mattered here. So sometimes session would not be rotated when moving to foreground after 30+ mins of inactivity because touchSession() would be called first (depends on timing of notification from the system). We are now rotating the session if >30mins passed between touchSession() calls as well. 

Added/fixed a unit test

## :pencil: Checklist

### DIFF
--- a/PostHog/PostHogSessionManager.swift
+++ b/PostHog/PostHogSessionManager.swift
@@ -143,17 +143,20 @@ import Foundation
     func touchSession() {
         guard isNotReactNative() else { return }
 
+        let (currentSessionId, lastActive) = sessionLock.withLock {
+            (sessionId, sessionActivityTimestamp)
+        }
+
+        guard currentSessionId != nil else { return }
+
         let timeNow = now()
         let timestamp = timeNow.timeIntervalSince1970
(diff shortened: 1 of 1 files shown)


pauldambra (approved):
🙌

PostHog/PostHogSessionManager.swift
```diff
@@ -143,7 +143,15 @@ import Foundation
     func touchSession() {
         guard isNotReactNative() else { return }
 
-        let timestamp = now().timeIntervalSince1970
+        let timeNow = now()
+        let timestamp = timeNow.timeIntervalSince1970
+        let lastActive = sessionLock.withLock { sessionActivityTimestamp }
```
marandaneto, line 148: should we check if theres an active `sessionId`? we do on L156
not need to lock twice for getting `sessionActivityTimestamp`, reuse the `lastActive` below?

PR #350, opened 2025-05-07
feat: capture console logs for session replay

ioannisj:
## :bulb: Motivation and Context

Closes #258 

Docs PR: https://github.com/PostHog/posthog.com/pull/11473
Replay Settings PR: https://github.com/PostHog/posthog/pull/32178

| Dev | Recording | 
|----------|--------------|
|    |        | 

## :green_heart: How did you test it?

Locally

## :pencil: Checklist

- [ ] I reviewed the submitted code.
- [ ] I added tests to verify the changes.
- [ ] I updated the docs if needed.
- [ ] No breaking change or entry added to the changelog.

PR #352, opened 2025-05-30
fix: pause network capture on app backgrounded

ioannisj:

Helps with https://github.com/PostHog/posthog-js/issues/2164

Stop capturing network logs in the background as discussed [here](https://github.com/PostHog/posthog-ios/pull/350#discussion_r2079815135). This also helps with extended sessions in React Native since RN could not check and rotate the session id for activity that happens on the native layer in the background.  

This is a small refactor of the network capture logic to conform to `PostHogSessionReplayPlugin` protocol - so no major changes there. Logic to pause/resume all plugins on background/foreground events was already there. 

Tested locally, no new tests

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Next
 
 - fix: pause network capture when app is backgrounded ([#352](https://github.com/PostHog/posthog-ios/pull/352))
+- fix: prevent duplicate application lifecycle events ([#354](https://github.com/PostHog/posthog-ios/pull/354))
 
 ## 3.26.1 - 2025-05-30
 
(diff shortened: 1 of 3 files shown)


PostHog/Replay/Plugins/Network/PostHogSessionReplayNetworkPlugin.swift
```diff
@@ -0,0 +1,87 @@
...
+            self.postHog = postHog
+            isActive = true
+
+            do {
+                sessionSwizzler = try URLSessionSwizzler(
+                    shouldCapture: shouldCaptureNetworkSample,
+                    onCapture: handleNetworkSample
+                )
+                sessionSwizzler?.swizzle()
+                hedgeLog("[Session Replay] Network telemetry plugin started")
```
marandaneto:
```suggestion
            do {
                sessionSwizzler = try URLSessionSwizzler(
                    shouldCapture: shouldCaptureNetworkSample,
                    onCapture: handleNetworkSample
                )
                sessionSwizzler?.swizzle()
                hedgeLog("[Session Replay] Network telemetry plugin started")
                isActive = true
```

otherwise if there's an exception, isActive = true but not really

PR #355, opened 2025-06-12
fix: do not call callback with invalid flags

marandaneto:
## :pencil: Checklist


PostHog/PostHogRemoteConfig.swift
```diff
@@ -273,7 +273,7 @@ class PostHogRemoteConfig {
                       let featureFlagPayloads = data["featureFlagPayloads"] as? [String: Any]
                 else {
                     hedgeLog("Error: Decide response missing correct featureFlags format")
-                    self.notifyFeatureFlagsAndRelease(data)
+                    self.notifyFeatureFlagsAndRelease(nil)
```
marandaneto, line 276: otherwise we'd pass flags with invalid/broken format

PR #356, opened 2025-06-12
fix: unify storage path for `appGroupIdentifier` across targets

hoppsen:

When integrating PostHog into both my app and its widget extension, I discovered that each target was getting its own AnonymousId. Although I’d set:

```swift
let config = PostHogConfig(apiKey: ..., host: ...)
config.appGroupIdentifier = "group.com.hoppsen.current_app"
...
```

Debugging the code and folders I noticed that the code I now changed appended the `bundleIdentifer` to the folder path, which in my case resulted in multiple folders (one per target).

Because of this, the widget and the main app weren’t sharing the same AnonymousId. This change removes the bundle-ID suffix so that all targets point to one shared container under the app group.

```
// Before fix:
.../AppGroup/123456789/Library/Application%20Support/com.hoppsen.current_app/phc_...
.../AppGroup/123456789/Library/Application%20Support/com.hoppsen.current_app.WidgetExtension/phc_...

// After fix:
.../AppGroup/123456789/Library/Application%20Support/group.com.hoppsen.current_app/phc_...
```

I tested it within my local environment and ran the updated unit test to confirm the suffix no longer includes the bundle identifier.

## :pencil: Checklist

- [x] ~I updated the docs if needed.~

---
Thanks for an amazing product! ❤️ Please let me know if I’ve misunderstood the `bundleIdentifier`’s intended role.

### DIFF
--- a/PostHog/PostHogStorage.swift
+++ b/PostHog/PostHogStorage.swift
@@ -47,8 +47,6 @@ func applicationSupportDirectoryURL() -> URL {
 func appGroupContainerUrl(config: PostHogConfig) -> URL? {
     guard let appGroupIdentifier = config.appGroupIdentifier else { return nil }
 
-    let bundleIdentifier = getBundleIdentifier()
-
     #if os(tvOS)
         // tvOS: Due to stricter sandbox rules, creating "Application Support" directory is not possible on tvOS
         let librarySubPath = "Library/Caches/"
@@ -59,7 +57,7 @@ func appGroupContainerUrl(config: PostHogConfig) -> URL? {
     let url = FileManager.default
         .containerURL(forSecurityApplicationGroupIdentifier: appGroupIdentifier)?
         .appendingPathComponent(librarySubPath)
-        .appendingPathComponent(bundleIdentifier)
+        .appendingPathComponent(appGroupIdentifier)
 
     if let url {
         createDirectoryAtURLIfNeeded(url: url)
--- a/PostHogTests/PostHogStorageTest.swift
+++ b/PostHogTests/PostHogStorageTest.swift
@@ -25,12 +25,20 @@ class PostHogStorageTest {
 
     @Test("returns the app group container dir URL")
     func returnsTheAppGroupContainerDirURL() {
+        let appGroupIdentifier = "group.com.posthog.some_identifier"
         let config = PostHogConfig(apiKey: "123")
-        config.appGroupIdentifier = "some_identifier"
+        config.appGroupIdentifier = appGroupIdentifier
         let url = appGroupContainerUrl(config: config)!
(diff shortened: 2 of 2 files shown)


hoppsen:
Hi @ioannisj ! 👋  
Just tagging you in case you haven’t seen this yet—would love your feedback when you get a moment.

hoppsen:
> I wonder if we should also migrate existing files similar to what we do [here](https://github.com/PostHog/posthog-ios/blob/d6fde6cea4fbb2f968f116aa04572c39297a8c2b/PostHog/PostHogStorage.swift#L109) so that we don't miss any events? So copying from `Library/Application%20Support/<bundle_identifier>/phc_...` to `Library/Application%20Support/<appGroupIdentifier>/phc_...`. This migration would be a bit more tricky since we'll have duplicates - for example the new `posthog.queueFolder` would be created from main target migration, so extension migration will have to copy all contents in this existing folder

@ioannisj Migrating existing data into the shared app-group folder makes a lot of sense to avoid lost events. Would you like me to update this PR or are you already looking into it?

hoppsen:
@ioannisj Just fixed the SwiftLint issues 🫠

hoppsen:
@ioannisj Just fixed the SwiftLint issues 🫠

PostHog/PostHogStorage.swift
```diff
@@ -77,6 +76,122 @@ func getBundleIdentifier() -> String {
...
+ Removes a directory if it's empty.
+
+ - Parameters:
+   - url: The directory URL to potentially remove
+ - Returns: `true` if the directory was removed, `false` otherwise
+ */
+@discardableResult
+func removeIfEmpty(_ url: URL) -> Bool {
+    let remainingItems = try? FileManager.default.contentsOfDirectory(at: url, includingPropertiesForKeys: nil, options: [])
+    if remainingItems?.isEmpty ?? true {
```
ioannisj:
```suggestion
    if remainingItems?.isEmpty == true {
```

ioannisj: Otherwise this would attempt to remove if line above throws. I'd rather remove the folder only when we are confident that the directory is empty

hoppsen: Applied the change 🚀

hoppsen: I think this should be the same 🤔 

```diff
- if remainingItems?.isEmpty == true {
+ if remainingItems?.isEmpty {
```

ioannisj: Pretty sure that won't compile because of optional Bool?

hoppsen: Yeah true 🤦🏼‍♂️, we would need to do `let remainingItems, remainingItems.isEmpty {` but then `... == true` is shorter 👍🏼

PR #357, opened 2025-06-12
feat: add support for beforeSend function to edit or drop events

wowrumal:
Required to filter events that the user doesn't want to send to the server.

Wrote a test + manual

## :pencil: Checklist

### DIFF
--- a/PostHog/Models/PostHogEvent.swift
+++ b/PostHog/Models/PostHogEvent.swift
@@ -84,7 +84,7 @@ import Foundation
     }
 }
 
-enum PostHogEventName: String, CaseIterable {
+enum PostHogKnownUnsafeEditableEvent: String {
     case snapshot = "$snapshot"
     case pageview = "$pageview"
     case pageleave = "$pageleave"
@@ -100,7 +100,7 @@ enum PostHogEventName: String, CaseIterable {
     case featureEnrollmentUpdate = "$feature_enrollment_update"
     case featureFlagCalled = "$feature_flag_called"
 
-    static func isUnsafeEditable(_ name: String) -> Bool {
-        PostHogEventName(rawValue: name) != nil
+    static func contains(_ name: String) -> Bool {
+        PostHogKnownUnsafeEditableEvent(rawValue: name) != nil
     }
 }
--- a/PostHog/Models/PostHogEvent.swift
+++ b/PostHog/Models/PostHogEvent.swift
@@ -7,7 +7,7 @@
 
 import Foundation
 
-public class PostHogEvent {
+@objc(PostHogEvent) public class PostHogEvent: NSObject {
     public var event: String
     public var distinctId: String
     public var properties: [String: Any]
--- a/PostHog/Models/PostHogEvent.swift
+++ b/PostHog/Models/PostHogEvent.swift
@@ -8,13 +8,13 @@
 import Foundation
 
 @objc(PostHogEvent) public class PostHogEvent: NSObject {
-    public var event: String
-    public var distinctId: String
-    public var properties: [String: Any]
-    public var timestamp: Date
-    public var uuid: UUID
+    @objc public var event: String
+    @objc public var distinctId: String
+    @objc public var properties: [String: Any]
+    @objc public var timestamp: Date
+    @objc public private(set) var uuid: UUID
     // Only used for Replay
-    public var apiKey: String?
+    var apiKey: String?
 
     init(event: String, distinctId: String, properties: [String: Any]? = nil, timestamp: Date = Date(), uuid: UUID = UUID.v7(), apiKey: String? = nil) {
         self.event = event
@@ -81,5 +81,26 @@ import Foundation
         }
 
         return json
+    }
+}
+
+enum PostHogEventName: String, CaseIterable {
+    case snapshot = "$snapshot"
+    case pageview = "$pageview"
+    case pageleave = "$pageleave"
+    case set = "$set"
+    case surveyDismissed = "survey dismissed"
+    case surveySent = "survey sent"
+    case surveyShown = "survey shown"
+    case identify = "$identify"
+    case groupidentify = "$groupidentify"
+    case createAlias = "$create_alias"
+    case clientIngestionWarning = "$$client_ingestion_warning"
+    case webExperimentApplied = "$web_experiment_applied"
+    case featureEnrollmentUpdate = "$feature_enrollment_update"
+    case featureFlagCalled = "$feature_flag_called"
+
+    static func isUnsafeEditable(_ name: String) -> Bool {
+        PostHogEventName(rawValue: name) != nil
     }
 }
--- a/PostHog/PostHogConfig.swift
+++ b/PostHog/PostHogConfig.swift
@@ -6,6 +6,8 @@
 //
 import Foundation
 
+public typealias BeforeSendBlock = (PostHogEvent) -> PostHogEvent?
+
 @objc(PostHogConfig) public class PostHogConfig: NSObject {
     enum Defaults {
         #if os(tvOS)
@@ -66,9 +68,13 @@ import Foundation
 
     /// Hook that allows to sanitize the event properties
     /// The hook is called before the event is cached or sent over the wire
+    @available(*, deprecated, message: "Use beforeSend instead")
     @objc public var propertiesSanitizer: PostHogPropertiesSanitizer?
     /// Determines the behavior for processing user profiles.
     @objc public var personProfiles: PostHogPersonProfiles = .identifiedOnly
+    /// Hook that allows to sanitize the event
+    /// The hook is called before the event is cached or sent over the wire
+    @objc public var beforeSend: BeforeSendBlock = { $0 }
 
     /// The identifier of the App Group that should be used to store shared analytics data.
     /// PostHog will try to get the physical location of the App Group’s shared container, otherwise fallback to the default location
--- a/PostHog/PostHogConfig.swift
+++ b/PostHog/PostHogConfig.swift
@@ -6,7 +6,16 @@
 //
 import Foundation
 
-public typealias BeforeSendBlock = (PostHogEvent?) -> PostHogEvent?
+public typealias BeforeSendBlock = (PostHogEvent) -> PostHogEvent?
+
+@objc public final class BoxedBeforeSendBlock: NSObject {
+    @objc public let block: BeforeSendBlock
+
+    @objc(block:)
+    public init(block: @escaping BeforeSendBlock) {
+        self.block = block
+    }
+}
 
 @objc(PostHogConfig) public class PostHogConfig: NSObject {
     enum Defaults {
@@ -72,9 +81,6 @@ public typealias BeforeSendBlock = (PostHogEvent?) -> PostHogEvent?
     @objc public var propertiesSanitizer: PostHogPropertiesSanitizer?
     /// Determines the behavior for processing user profiles.
     @objc public var personProfiles: PostHogPersonProfiles = .identifiedOnly
-    /// Hook that allows to sanitize the event
-    /// The hook is called before the event is cached or sent over the wire
-    @objc public var beforeSend: BeforeSendBlock = { $0 }
 
     /// The identifier of the App Group that should be used to store shared analytics data.
     /// PostHog will try to get the physical location of the App Group’s shared container, otherwise fallback to the default location
@@ -176,5 +182,34 @@ public typealias BeforeSendBlock = (PostHogEvent?) -> PostHogEvent?
         if #available(iOS 15.0, *) {
             _surveys = value
         }
+    }
+
+    /// Hook that allows to sanitize the event
+    /// The hook is called before the event is cached or sent over the wire
+    private var beforeSend: BeforeSendBlock = { $0 }
+
+    private static func buildBeforeSendBlock(_ blocks: [BeforeSendBlock]) -> BeforeSendBlock {
+        { event in
+            blocks.reduce(event) { event, block in
+                event.flatMap(block)
+            }
+        }
+    }
+
+    public func setBeforeSend(_ blocks: [BeforeSendBlock]) {
+        beforeSend = Self.buildBeforeSendBlock(blocks)
+    }
+
+    public func setBeforeSend(_ blocks: BeforeSendBlock...) {
+        setBeforeSend(blocks)
+    }
+
+    @available(*, unavailable, message: "Use setBeforeSend(_ blocks: BeforeSendBlock...) instead")
+    @objc public func setBeforeSend(_ blocks: [BoxedBeforeSendBlock]) {
+        setBeforeSend(blocks.map(\.block))
+    }
+
+    func runBeforeSend(_ event: PostHogEvent) -> PostHogEvent? {
+        beforeSend(event)
     }
 }
--- a/PostHog/PostHogConfig.swift
+++ b/PostHog/PostHogConfig.swift
@@ -6,7 +6,7 @@
 //
 import Foundation
 
-public typealias BeforeSendBlock = (PostHogEvent) -> PostHogEvent?
+public typealias BeforeSendBlock = (PostHogEvent?) -> PostHogEvent?
 
 @objc(PostHogConfig) public class PostHogConfig: NSObject {
     enum Defaults {
--- a/PostHog/PostHogSDK.swift
+++ b/PostHog/PostHogSDK.swift
@@ -782,17 +782,28 @@ let maxRetryDelay = 30.0
         queue.add(event)
     }
 
-    func buildEvent(event: String, distinctId: String, properties: [String: Any], timestamp: Date = Date()) -> PostHogEvent? {
+    func buildEvent(event eventName: String, distinctId: String, properties: [String: Any], timestamp: Date = Date()) -> PostHogEvent? {
         let sanitizedProperties = sanitizeProperties(properties)
 
         let event = PostHogEvent(
-            event: event,
+            event: eventName,
             distinctId: distinctId,
             properties: sanitizedProperties,
             timestamp: timestamp
         )
 
-        return config.beforeSend(event)
+        let resultEvent = config.beforeSend(event)
+
+        if resultEvent == nil {
+            let originalMessage = "PostHog event \(eventName) was dropped"
+            let message = PostHogEventName.isUnsafeEditable(eventName)
+                ? "\(originalMessage). This can cause unexpected behavior."
+                : originalMessage
+
+            hedgeLog(message)
+        }
+
+        return resultEvent
     }
 
     @objc(groupWithType:key:)
--- a/PostHog/PostHogSDK.swift
+++ b/PostHog/PostHogSDK.swift
@@ -796,7 +796,7 @@ let maxRetryDelay = 30.0
 
         if resultEvent == nil {
             let originalMessage = "PostHog event \(eventName) was dropped"
-            let message = PostHogEventName.isUnsafeEditable(eventName)
+            let message = PostHogKnownUnsafeEditableEvent.contains(eventName)
                 ? "\(originalMessage). This can cause unexpected behavior."
                 : originalMessage
 
--- a/PostHog/PostHogSDK.swift
+++ b/PostHog/PostHogSDK.swift
@@ -461,13 +461,12 @@ let maxRetryDelay = 30.0
                 userProperties: sanitizeDictionary(userProperties),
                 userPropertiesSetOnce: sanitizeDictionary(userPropertiesSetOnce)
             )
-            let sanitizedProperties = sanitizeProperties(properties)
 
-            queue.add(PostHogEvent(
-                event: "$identify",
-                distinctId: distinctId,
-                properties: sanitizedProperties
-            ))
+            guard let event = buildEvent(event: "$identify", distinctId: distinctId, properties: properties) else {
+                return
+            }
+
+            queue.add(event)
 
             remoteConfig?.reloadFeatureFlags()
 
@@ -570,7 +569,7 @@ let maxRetryDelay = 30.0
             return
         }
 
-        let isSnapshotEvent = event == "$snapshot"
+        var isSnapshotEvent = event == "$snapshot"
         let eventTimestamp = timestamp ?? now()
         let eventDistinctId = distinctId ?? getDistinctId()
 
@@ -587,22 +586,29 @@ let maxRetryDelay = 30.0
                                          groups: groups,
                                          appendSharedProps: !isSnapshotEvent,
                                          timestamp: timestamp)
-        let sanitizedProperties = sanitizeProperties(properties)
+
+        // Sanitize is now called in buildEvent
+        let posthogEvent = buildEvent(
+            event: event,
+            distinctId: eventDistinctId,
+            properties: properties,
+            timestamp: eventTimestamp
+        )
+
+        guard let posthogEvent else {
+            return
+        }
+
+        // Reevaluate if this is a snapshot event because the event might have been updated by the beforeSend hook
+        isSnapshotEvent = posthogEvent.event == "$snapshot"
 
         // if this is a $snapshot event and $session_id is missing, don't process then event
-        if isSnapshotEvent, sanitizedProperties["$session_id"] == nil {
+        if isSnapshotEvent, posthogEvent.properties["$session_id"] == nil {
             return
         }
 
         // Session Replay has its own queue
         let targetQueue = isSnapshotEvent ? replayQueue : queue
-
-        let posthogEvent = PostHogEvent(
-            event: event,
-            distinctId: eventDistinctId,
-            properties: sanitizedProperties,
-            timestamp: eventTimestamp
-        )
 
         targetQueue?.add(posthogEvent)
 
@@ -636,13 +642,12 @@ let maxRetryDelay = 30.0
         let distinctId = getDistinctId()
 
         let properties = buildProperties(distinctId: distinctId, properties: props)
-        let sanitizedProperties = sanitizeProperties(properties)
 
-        queue.add(PostHogEvent(
-            event: "$screen",
-            distinctId: distinctId,
-            properties: sanitizedProperties
-        ))
+        guard let event = buildEvent(event: "$screen", distinctId: distinctId, properties: properties) else {
+            return
+        }
+
+        queue.add(event)
     }
 
     func autocapture(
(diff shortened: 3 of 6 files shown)


PostHog/PostHogSDK.swift
```diff
@@ -604,6 +604,10 @@ let maxRetryDelay = 30.0
             timestamp: eventTimestamp
         )
 
+        if config.eventsSanitizer?.shouldDrop(event, properties: sanitizedProperties) == true {
```
haritowa: I think it is better to use guard here

haritowa: Also, I think it is better to send the whole `posthogEvent` to this method, just in case(if it is a value type/non mutable type)

marandaneto:
agree with `before_send` the way to go https://posthog.com/docs/libraries/js/features#amending-or-sampling-events

PostHog/PostHogSDK.swift
```diff
@@ -461,13 +461,12 @@ let maxRetryDelay = 30.0
...
             )
-            let sanitizedProperties = sanitizeProperties(properties)
 
-            queue.add(PostHogEvent(
-                event: "$identify",
-                distinctId: distinctId,
-                properties: sanitizedProperties
-            ))
+            guard let event = buildEvent(event: "$identify", distinctId: distinctId, properties: properties) else {
+                return
```
marandaneto, line 466: we should add logging when events are dropped

PostHog/PostHogSDK.swift
```diff
@@ -771,13 +774,25 @@ let maxRetryDelay = 30.0
...
+        let event = PostHogEvent(
+            event: event,
             distinctId: distinctId,
-            properties: sanitizedProperties
-        ))
+            properties: sanitizedProperties,
+            timestamp: timestamp
+        )
+
+        return config.beforeSend(event)
```
marandaneto: i think we should not run beforeSend for `$snapshot` events (just a hunch), we have to check the JS SDK and do the same if thats the case

PostHog/Models/PostHogEvent.swift
```diff
@@ -7,7 +7,7 @@
 
 import Foundation
 
-public class PostHogEvent {
+@objc(PostHogEvent) public class PostHogEvent: NSObject {
```
ioannisj, line 10: None of the properties would be accessible in Objc here, so we need to either `@objcMembers` or decorate each property with `@objc` (I personally prefer this)

I would also change the visibility of the following: 

```
...
@objc public private(set) var uuid: UUID // so people don't send dupes
var apiKey: String? // Only used for Replay internally
```

@marandaneto is `apiKey` used outside of this module and needs to be public if you know?

PostHog/PostHogConfig.swift
```diff
@@ -66,9 +68,13 @@ import Foundation
 
     /// Hook that allows to sanitize the event properties
     /// The hook is called before the event is cached or sent over the wire
+    @available(*, deprecated, message: "Use beforeSend instead")
```
ioannisj, line 80: Pretty sure we do, but just to verify, we are retaining the functionality for people who already use this right?

PostHog/PostHogSDK.swift
```diff
@@ -771,13 +774,25 @@ let maxRetryDelay = 30.0
...
+
+        guard let event = buildEvent(event: "$groupidentify", distinctId: distinctId, properties: properties) else {
+            return
+        }
+
+        queue.add(event)
+    }
+
+    func buildEvent(event: String, distinctId: String, properties: [String: Any], timestamp: Date = Date()) -> PostHogEvent? {
         let sanitizedProperties = sanitizeProperties(properties)
```
ioannisj, line 786: 👍

PostHog/PostHogSDK.swift
ioannisj: I don't see any special handling [here](https://github.com/PostHog/posthog-js/blob/caf21938af1e901d63ed5c88d2e508d08c38163f/src/posthog-core.ts#L967) or any dedicated tests for this - not sure if I'm missing something though

PostHog/PostHogConfig.swift
```diff
@@ -6,6 +6,8 @@
 //
 import Foundation
 
+public typealias BeforeSendBlock = (PostHogEvent) -> PostHogEvent?
```
ioannisj, line 9: https://github.com/PostHog/posthog-ios/pull/357#issuecomment-2972826745

> The js version also has a [chaining version](https://posthog.com/docs/libraries/js/features#chaining-before_send-functions) of before_send init, yet we decided to omit this, because the iOS SDK handles this kind of stuff differently: via props

We could maybe change the signature to `(PostHogEvent?) -> PostHogEvent?` so that people can compose: 

```
sampleByDistinctId(0.5)(
    sampleByEvent("Event Name")(
         sampleBySessionId(0.1)(
			event
         )
    )
)
```

Would be awesome to also provide these functions here as well [js](https://github.com/PostHog/posthog-js/blob/caf21938af1e901d63ed5c88d2e508d08c38163f/src/customizations/before-send.ts#L72) but could come with a later improvement/PR

PostHog/PostHogSDK.swift
marandaneto: Better to check with replay team.
Dropped snapshots can cause a broken recording

PostHog/Models/PostHogEvent.swift
marandaneto: as per comment just `snapshot` API endpoint needs that so it does not need to be public

PostHog/PostHogConfig.swift
haritowa: We thought about some "seamless migration," i.e., to mark this prop as write-only and deprecated, and actually just set up a beforeSend closure with sanitizer logic. Yet, I think it is vague for the end user because they can override their property sanitizer when they start to use the beforeSend hook. I think the js lib has this prop deprecated(yet still available) too

PostHog/PostHogSDK.swift
haritowa: We are ready to add this into PR, waiting for your clarification

ioannisj: Hey @haritowa, so the JS lib just [logs a warning](https://github.com/PostHog/posthog-js/blob/b61616775fb18c5da2f2159018de0808c96a5390/packages/browser/src/posthog-core.ts#L2301-L2305) when a [knownUnsafeEditableEvent](https://github.com/PostHog/posthog-js/blob/b61616775fb18c5da2f2159018de0808c96a5390/packages/browser/src/types.ts#L11) is skipped/rejected by before_send. We should do the same here as well along with https://github.com/PostHog/posthog-ios/pull/357#discussion_r2149308979

PostHog/PostHogConfig.swift
haritowa: What should we do if the `beforeSend` hook received nil and returned non-nil? :D

Should we send it? Should we wrap `beforeSend` with a nil check and filter our such invocations?
If we introduce this signature, we have to handle such cases...

Actually, swift compiler will gladly allow us to pass `(T?) -> T?` to `(T) -> T?` prop, yet I am not sure how this will work for objc consumers

```swift
import Foundation

class DataClass {
  var prop: String

  init(prop: String) {
    self.prop = prop
  }
}

class Manager {
  var hook: (DataClass) -> DataClass? = { $0 }
}

func mutateFunction(input: DataClass?) -> DataClass? {
  input?.prop = "test 2"
  return input
}

let manager = Manager()
let initialDataClass = DataClass(prop: "test 1")

print(manager.hook(initialDataClass)?.prop) // Optional("test 1")
manager.hook = mutateFunction

print(manager.hook(initialDataClass)?.prop) // Optional("test 2")
```

haritowa: Afaik there are no such cases in the code right now, yet we allow future contributors to blindly send nils to hook which may introduce bugs later. Also, I think it is not great interface-wise too, because it forces consumers to reason about "why can I get nil in the first place". I think we should just keep the signature as is, and it will work event for chaining(at least in swift), or we could introduce some `build(mutations: (PostHogEvent) -> PostHogEvent?...)` later

PostHog/PostHogSDK.swift
haritowa: > Hey @haritowa, so the JS lib just [logs a warning](https://github.com/PostHog/posthog-js/blob/b61616775fb18c5da2f2159018de0808c96a5390/packages/browser/src/posthog-core.ts#L2301-L2305) when a [knownUnsafeEditableEvent](https://github.com/PostHog/posthog-js/blob/b61616775fb18c5da2f2159018de0808c96a5390/packages/browser/src/types.ts#L11) is skipped/rejected by before_send. We should do the same here as well along with [#357 (comment)](https://github.com/PostHog/posthog-ios/pull/357#discussion_r2149308979)

Sounds good, but the iOS SDK doesn't have log levels, so this will be just a regular log. Is that okay? Also, can we somehow allow editing the `knownUnsafeEditableEvent` list from the outside? This way, we can silence these warnings in case we know what we are doing. Our log console is a mess after this change :D

PostHog/PostHogSDK.swift
wowrumal: Added logging after beforeSend call returns nil event as in JS library

PostHog/Models/PostHogEvent.swift
wowrumal: Added @objc property decorations for each property except `apiKey`. Also added `private(set)` for  `uuid` property

PostHog/PostHogSDK.swift
haritowa: I thought a bit more about this warning, and I think it still does not cover all the funky edge cases. For example, if the user modifies a snapshot event, we will not print any warnings because it is not `nil` after the mutation. This is consistent with the JS implementation, yet I think it is not the correct behavior.

The simplest way to fix it is to compare events before and after the hook and print a warning if:
-  the event changed and the original event was `knownUnsafeEditableEvent`
-  the event changed and the mutated event became `knownUnsafeEditableEvent`

Alternatively, we could introduce a separate hook for unsafe events, but again, this is a bit messy and also not aligned with the JS implementation.

Lastly, we can keep things as they are (print a warning only if the hook returns `nil` for unsafe events)

Also, we don't copy `PostHogEvent` either before or after the `hook` invocation. I am not sure if this is okay because before our change, the event object wasn't exposed to consumers (i.e., it wasn't possible to modify this object or capture it in the hook closure and modify it later).

We tested the "modified outside of hook closure" case, though, and PostHog actually sent the correct unmodified event; not sure why :D

haritowa:
☝️ we shipped some updates, could you please resolve already fixed comments so we can focus on the remaining bits

PostHog/PostHogConfig.swift
ioannisj: Yeah, keeping the existing functionality should be the way to go with this I think

PostHog/PostHogConfig.swift
ioannisj: The idea of making the parameter optional is just to make it easier for SDK users to compose their own `beforeSend` chains. In Swift, optional chaining makes this very ergonomic (it’s just language sugar though) - for Objective-C, a simple null check would be enough. From our side, we'll be calling a single function - and we'll never call it with nil.

This mirrors what the JS SDK [does](https://github.com/PostHog/posthog-js/blob/559f0975be4b4819fbe74527e69b6a5ec884c539/packages/browser/src/types.ts#L276) btw — it uses the same endofunction. 

>Afaik there are no such cases in the code right now, yet we allow future contributors to blindly send nils to hook which may introduce bugs later. Also, I think it is not great interface-wise too, because it forces consumers to reason about "why can I get nil in the first place". I think we should just keep the signature as is, and it will work event for chaining(at least in swift), or we could introduce some build(mutations: (PostHogEvent) -> PostHogEvent?...) later

Yeah, I do understand your pov here. Sticking with (PostHogEvent) -> PostHogEvent? is okay as well - it just means that for more advanced `beforeSend` chains, SDK users will need to do some extra work - which is acceptable as well. 

I don't really have a strong preference here. @marandaneto what's your take on this since we'll probably have to mirror this on other mobile SDKs as well - we should be consistent with the api

PostHog/Models/PostHogEvent.swift
```diff
@@ -83,3 +83,24 @@ public class PostHogEvent {
         return json
     }
 }
+
+enum PostHogEventName: String, CaseIterable {
```
ioannisj: nit: type name does not accurately reflect what we do here - there are other posthog events that are not or won't be in this list. (e.g $autocapture).  I’d suggest sticking with something like PostHogKnownUnsafeEditableEvent to better reflect its intent/scope?

PostHog/PostHogSDK.swift
ioannisj: > Also, we don't copy PostHogEvent either before or after the hook invocation. I am not sure if this is okay because before our change, the event object wasn't exposed to consumers (i.e., it wasn't possible to modify this object or capture it in the hook closure and modify it later).

Ideally, we’d use immutable value types here, which would avoid these concerns entirely. But that would complicate interop with Objective-C, since Objective-C can only work with classes (reference types), not structs/enums. So for ObjC would need to user a reference type anw. 

The reason there's no need to copy is because `PostHogEvent` is a reference type, so it's passed by reference in the closure and is mutable.  

> We tested the "modified outside of hook closure" case, though, and PostHog actually sent the correct unmodified event; not sure why :D

Right after beforeSend() runs, the event is added to a queue and serialized to disk. By the time the event is actually flushed, it’s deserialized into a completely fresh instance - so any mutations done outside the closure won’t affect what gets sent, unless they happen before serialization finishes (probably in a concurrent queue).

If we ever want stricter safety guarantees, we could consider deep-copying before queuing — but for now, the window for these race conditions is quite narrow I would say and more tied to SDK misuse or unsafe async behavior? 

cc: @marandaneto

PostHog/Models/PostHogEvent.swift
wowrumal: Changed type name to `PostHogKnownUnsafeEditableEvent`

PostHog/PostHogSDK.swift
haritowa: > Right after beforeSend() runs, the event is added to a queue and serialized to disk

Got it, thanks

PostHog/PostHogConfig.swift
marandaneto: JS is either one single fn or an array of fn.
I prefer `(PostHogEvent) -> PostHogEvent?` because we will never call the beforeSend fn with null, but we can return null, so I'd focus on the semantics here.

People can always chain themselves..

The option would be to make `beforeSend` an array for easy chaining.

`beforeSend: BeforeSendBlock[]?`

So people can chain by `beforeSend = [fn1, fn2, ...]`

When the SDK iterates over the `beforeSend` array, we only call the next fn if the returned event isn't dropped already.

This is what JS does and I'd stick to it.

PostHog/PostHogSDK.swift
marandaneto: > If we ever want stricter safety guarantees, we could consider deep-copying before queuing — but for now, the window for these race conditions is quite narrow I would say and more tied to SDK misuse or unsafe async behavior?

agree, no need to handle those use cases, it'd just add complexity to the SDK for people misusing the before send callback

PostHog/PostHogConfig.swift
haritowa: I tried different approaches but the bottom line is: looks like we can't expose an array of closures param to Obj-C in a typesafe way. [More details here](https://www.perplexity.ai/search/how-to-use-swift-bridging-with-sfhy5SJpTJe_1NVhWnvqYA)

Without obj-c I think this implementation will actually be good:
```swift
    /// Hook that allows to sanitize the event
    /// The hook is called before the event is cached or sent over the wire
    private var beforeSendBlock: BeforeSendBlock = { $0 }

    // This method is used to build the beforeSendBlock from the array of blocks
    // It calls each block in order if the previous block returns a non-nil value
    // If the last block returns a nil value -- nil is returned
    private func buildBeforeSendBlock(_ blocks: [BeforeSendBlock]) -> BeforeSendBlock {
        { blocks.reduce(Optional($0)) { event, block in
            event.flatMap(block)
        } }
    }

    public func setBeforeSendBlock(_ blocks: BeforeSendBlock...) {
        beforeSendBlock = buildBeforeSendBlock(blocks)
    }

    public func appendBeforeSendBlock(_ block: BeforeSendBlock...) {
        beforeSendBlock = buildBeforeSendBlock([beforeSendBlock] + block)
    }
```

_I am reminding that we are doing all this kind of stuff to port js sugar to esoteric obj-c users..._

Just to be clear, I tried your exact proposed API too:
CleanShot 2025-06-18 at 14 29 30@2x

If you are **really want to implement it**, we could add something like [this](https://www.perplexity.ai/search/objc-only-attribute-swift-UrG50lbhTmOipiRChJcuHw). Still feels goofy af, but we can and check how it feels after the implementation :)

## Some obj-c love approach
Also, I created a quick prototype showcasing how we could almost achieve everything you want without messy badly typed code:
- allows us to assign array of hooks in swift
- working get/set semantics for swift (will be impossible to do with closure fuse on set)
- allows us to append hooks both from obj-c and swift
- allows us to remove all hooks both from obj-c and swift

This approach will both implement all sugar for swift version, yet exposing all essential methods to obj-c

```swift
@propertyWrapper
public struct CallbackChain<T> {
    private var callbacks: [(T) -> T?] = []
    
    public var wrappedValue: [(T) -> T?] {
        get { callbacks }
        set { callbacks = newValue }
    }
    
    public var projectedValue: (T) -> T? {
        { input in
            guard !callbacks.isEmpty else { return input }
            return callbacks.reduce(Optional(input)) { $0.flatMap($1) }
        }
    }
    
    public init(wrappedValue: [(T) -> T?] = []) {
        self.callbacks = wrappedValue
    }
}

...
    @CallbackChain public var beforeSend: [BeforeSendBlock] = []
    
    /// Add a beforeSend block (Objective-C compatible)
    @objc public func appendBeforeSendBlock(_ block: @escaping (PostHogEvent) -> PostHogEvent?) {
        beforeSend.append(block)
    }
    
    /// Remove all beforeSend blocks (Objective-C compatible)
    @objc public func removeAllBeforeSendBlocks() {
        beforeSend.removeAll()
    }
    
    /// Execute the beforeSend chain (internal SDK use only)
    func executeBeforeSendChain(_ event: PostHogEvent) -> PostHogEvent? {
        return $beforeSend(event)
    }
```

Or we could we use something like this
```swift
    @CallbackChain public var beforeSend: [BeforeSendBlock] = []
    
    // Sugar for Objective-C compatibility
    @objc public var beforeSendBlock: BeforeSendBlock? {
        get {
            return $beforeSend
        }
        set {
            beforeSend = [newValue].compactMap { $0 }
        }
    }

    /// Add a beforeSend block (Objective-C compatible)
    @available(*, unavailable, message: "Use beforeSend instead")
    @objc public func appendBeforeSendBlock(_ block: @escaping (PostHogEvent) -> PostHogEvent?) {
        beforeSend.append(block)
    }
    
    /// Execute the beforeSend chain (internal SDK use only)
    func executeBeforeSendChain(_ event: PostHogEvent) -> PostHogEvent? {
        return $beforeSend(event)
    }
 ```

File: PostHog/PostHogSDK.swift (hunk 3 above)
Comment (haritowa) [reply PostHog/PostHogSDK.swift]: Let's figure out the remaining questions in this thread:
> I thought a bit more about this warning, and I think it still does not cover all the funky edge cases.

https://github.com/PostHog/posthog-ios/pull/357#discussion_r2150169586

> Also, can we somehow allow editing the knownUnsafeEditableEvent list from the outside? This way, we can silence these warnings in case we know what we are doing. Our log console is a mess after this change :D

https://github.com/PostHog/posthog-ios/pull/357#discussion_r2150095809

Sorry for me being annoying, we want to finalize this contribution and use it in our release :)

File: PostHog/PostHogConfig.swift (hunk 7 above)
Comment (ioannisj) [reply PostHog/PostHogConfig.swift:9]: @haritowa Yeah, I think the only way to expose an array of closures to Objc is by boxing in an NSObject. Can we maybe have something simple like this: 

```swift
    // Swift Only (note: no @objc)
    public var beforeSend: [BeforeSendBlock] = [{ $0 }]
    
    // Objc
    @available(*, unavailable)
    @objc public func setBeforeSend(_ boxed: [BoxedBeforeSendBlock]) {
        beforeSend = boxed.map(\.block)
    }
```

With BoxedBeforeSendBlock being: 

```swift
@objc public final class BoxedBeforeSendBlock: NSObject {
    let block: BeforeSendBlock

    init(block: @escaping BeforeSendBlock) {
        self.block = block
    }
}
```

We can use the `@available(*, unavailable)` trick to hide objc flavor from swift users as you suggested

haritowa: How will the objc usage look in this scenario? Do they have to wrap their blocks into objects? Because it actually looks like we are hurting the most popular way to use the hook (just a single block instead of a block wrapped in an object wrapped in an array) for some less popular cases.

Also, there is no way to get the current hook closure from objc with this API? Is that okay?

ioannisj: > How will the objc usage look in this scenario? Do they have to wrap their blocks into objects? Because it actually looks like we are hurting the most popular way to use the hook (just a single block instead of a block wrapped in an object wrapped in an array) for some less popular cases.
> 
> 

Yeah objc users will have to wrap each closure in an object this way, which I think is an okay compromise. We can of course go with the appendBeforeSendBlock approach you suggested as well if you think this would improve the objc api. Swift users can use the `beforeSend` property directly. 

> Also, there is no way to get the current hook closure from objc with this API? Is that okay?

If I'm not mistaken none of the suggestions above achieved this either since we can only interface with NSObject[]. I think it's okay since I can't see any meaningful use case of reading an array of closures you provided yourself to the SDK back

ioannisj: If we want full symmetry between the two APIs, we could even make `beforeSend` internal and expose a Swift-facing method (accepting an array or variadic list of BeforeSendBlocks) - effectively making both Swift and Objective-C write-only

haritowa: Ok, looks good. I'll store the `beforeSend` closure as a single flattened closure, is it ok? And also I will handle the case where empty array was passed(will just reset to `{ $0 }` in this case)

PostHog/PostHogSDK.swift
ioannisj: > I thought a bit more about this warning, and I think it still does not cover all the funky edge cases. For example, if the user modifies a snapshot event, we will not print any warnings because it is not nil after the mutation.

I think we can skip the full before/after comparison—doing a deep compare on [String: Any] feels like it's going to be expensive - since JS tolerates this risk, I’d argue we can too. So I’m inclined to keep things consistent and avoid the extra overhead here.

> Also, can we somehow allow editing the knownUnsafeEditableEvent list from the outside? This way, we can silence these warnings in case we know what we are doing. Our log console is a mess after this change :D

Just to confirm—are you mainly trying to reduce log noise during development? Cause setting config.debug = false would already help.

I’m not too sure about exposing knownUnsafeEditableEvent for editing—it feels a bit arbitrary and I’m not convinced others would benefit from managing that themselves. I’d lean towards leaving it as-is for now, and maybe later we introduce something like config.minDebugLevel to control verbosity more broadly. If this indeed an issue for you, how about an all-or-nothing `config.silenceBeforeSendLogs`?

PostHog/PostHogConfig.swift
haritowa: Done, yet I am not sure how to test it with objc. Do you have some objc related tests?

haritowa:
I think we should update PR title, it mutated into something completely different 😂

PostHog/PostHogSDK.swift
haritowa: Yeah, I think we will just live with these logs, I don't want to miss something important by silencing all the logs. And I agree, this feels wrong to expose internal stuff just to solve our minor issue

PostHog/PostHogConfig.swift
haritowa: Ah, I almost forgot. We actually got some cool Swift sugar, and a single hook closure can be set like [this](https://github.com/PostHog/posthog-ios/pull/357/files#diff-ccdae913e9a4a620023a87b8f54f7fb465761883177a4a7d91aeec9d584feb72R848) (even with our variadic implementation). I think this API is 🔥

```swift
config.setBeforeSend { $0.event == "first_event" ? nil : $0 }
```

Yet all of the below is valid too:
```swift
setBeforeSend { $0.event == "test_event" ? nil : $0 }

setBeforeSend(
    { $0.event == "other_event" ? nil : $0 },
    { $0.event = "modified_event"; return $0 }
)

setBeforeSend([
    { $0.event == "other_event" ? nil : $0 },
    { $0.event = "modified_event"; return $0 }
])
```

PostHogTests/PostHogSDKTest.swift
```diff
@@ -640,6 +685,207 @@ class PostHogSDKTest: QuickSpec {
...
+                        ]
+                    )
+
+                    sut.capture("first_event")
+                    sut.capture("second_event")
+                    sut.capture("third_event")
+
+                    // first event is skipped by the first block
+                    // second event is modified by the second block and not skipped by the third block(because it became "modified_event")
+                    // third event is modified by the third block
```
ioannisj:
```suggestion
                    // third event is modified by the second block
```

ioannisj (commented):
Thanks all — I think this is already in great shape. Really appreciate the effort and thought put into this! ❤️

Only thing left is to add a changelog entry so CI passes, and then I think we’re good to merge. 🚀

PostHog/PostHogConfig.swift
ioannisj: > Done, yet I am not sure how to test it with objc. Do you have some objc related tests?

Sorry missed this. No we don't have any. There's just a `PostHogObjCExample` sample we use to test the API

PostHogTests/PostHogSDKTest.swift
haritowa: 😬 oops

ioannisj (approved):

PostHog/Models/PostHogEvent.swift
```diff
@@ -83,3 +83,24 @@ public class PostHogEvent {
...
+    case surveyDismissed = "survey dismissed"
+    case surveySent = "survey sent"
+    case surveyShown = "survey shown"
+    case identify = "$identify"
+    case groupidentify = "$groupidentify"
+    case createAlias = "$create_alias"
+    case clientIngestionWarning = "$$client_ingestion_warning"
+    case webExperimentApplied = "$web_experiment_applied"
+    case featureEnrollmentUpdate = "$feature_enrollment_update"
+    case featureFlagCalled = "$feature_flag_called"
```
marandaneto, line 101: @ioannisj some of those events arent used in this library, lets remove them to save some bundle bits

PR #359, opened 2025-06-19
feat: add renderSurvey delegate

ioannisj:
### Add support for custom survey rendering 

This PR introduces a new capability to override the default survey rendering logic. While the primary use case is to allow the Flutter SDK to render surveys natively in Flutter, this also enables any SDK user to implement a fully custom survey UI

## :pencil: Checklist


anfriis:
Looking forward to this! 
Would it be possible to also change the default top padding or a possibility to customize it?
It looks weird the it doesn't have any padding between the white top bar and the title, so we are holding back on using surveys at the moment in our app:
Survey

anfriis:
> Hey @anfriis — the goal of this feature is to give you full control over survey rendering when needed which will unlock full customization but is definitely more development intensive!
> 
> You’re totally right about the padding — while it should be customizable via the survey editor in the PostHog app, the iOS implementation is currently ignoring that setting. I’ll get a fix out for this soon and will update you when it’s live. Sorry that this has been a blocker for using surveys in your app — really appreciate you flagging it!

Thank for the quick feedback.
Yeah, I can imagine!
With the padding fixed the design is decent enough for us to use it :)
Thanks 🥇

PostHog/Models/Surveys/PostHogSurvey+Display.swift
```diff
@@ -0,0 +1,132 @@
...
+            case .emoji:
+                scale == 3 ? 1 ... 3 : 1 ... 5
+            case .number:
+                switch scale {
+                case 7: 1 ... 7
+                case 10: 0 ... 10
+                default: 1 ... 5
+                }
+            default:
+                1 ... 5
```
marandaneto: I think we should use an enum value here, just so that we avoid hardcoded values and reuse them across the repo

marandaneto: juar saw it, can we reuse this here?

```
    var rawValue: Int {
        switch self {
        case .threePoint: 3
        case .fivePoint: 5
        case .sevenPoint: 7
        case .tenPoint: 10
        case let .unknown(scale): scale
        }
    }
```

PostHog/Surveys/PostHogSurveyIntegration.swift
```diff
@@ -562,6 +802,44 @@
...
+        }
+
+        enum SevenPoint {
+            static let negatives = 1 ... 3
+            static let neutrals = 4 ... 4
+        }
+
+        enum TenPoint {
+            static let detractors = 0 ... 6
+            static let passives = 7 ... 8
```
marandaneto, line 839: can we reuse the enum for the hardcoded values

PostHog/Surveys/Utils/EmojiRating.swift
```diff
@@ -53,14 +53,14 @@
 
         // swiftlint:disable:next cyclomatic_complexity
         @ViewBuilder private func emoji(for value: Int) -> some View {
-            if emojiRange.range.count == 3 {
+            if range.count == 3 {
```
marandaneto: reuse enum?

PostHog/Surveys/Utils/EmojiRating.swift
```diff
@@ -107,7 +91,7 @@
                     VStack(spacing: 40) {
                         EmojiRating(
                             selectedValue: $selectedValue,
-                            emojiRange: .oneToFive,
+                            range: 1 ... 5,
```
marandaneto: reuse enum

PostHog/Surveys/Utils/NumberRating.swift
```diff
@@ -109,7 +95,7 @@
                     VStack(spacing: 15) {
                         NumberRating(
                             selectedValue: $selectedValue,
-                            numberRange: .oneToFive,
+                            range: 0 ... 10,
```
marandaneto: reuse range vlaue?

marandaneto (commented):
left a few comments but overall all good

marandaneto:
@ioannisj 
> Docs PR: TODO

i'd not document this for now until its being used by Flutter and proven to work just fine, we might need to change things after the first version, and it'd be cool to be flexible here.
i'd also add a comment on the public methods that this is internal only for now

PostHog/Models/Surveys/PostHogSurvey+Display.swift
ioannisj: yeap, I can add the range logic to this enum actually. For PostHogDisplaySurveyQuestion however - which is exposed to ObjC - I think we'll need stick to `scaleLowerBound`/`scaleUpperBound` simple type since we can't expose an enum with associated values

marandaneto (approved):
besides the comments i already left, LGTM

PR #368, opened 2025-07-14
feat: do not clear feature flags on remote config error

### LINKED ISSUE #367: Feature flags do not load from cache if app has network connectivity issues when opening
### Version

3.28.2

### Steps to Reproduce

I am able to consistently reproduce this issue. Put shortly, users that open the app without connectivity will see their feature flags as `null`.

Steps:
1. Have `preloadFeatureFlags = true`
2. Open the app with connectivity, have the feature flags properly load
3. Close the app
4. Use a proxy (I am using Proxyman) to set the network failures to the Posthog host to 100%
5. Open the app

### Expected Result

Feature flags that were cached on step 2 would load, even without connectivity.

### Actual Result

The feature flags that were cached on step 2. won't load, instead they will all return as `nil`.

### Potential Cause 
I was able to trace, and the connectivity error causes the flags to return as `nil`:

https://github.com/PostHog/posthog-ios/blob/784145ddd2f7de3a6d6fcce0e50c04f29d33b338/PostHog/PostHogApi.swift#L275-L278

https://github.com/PostHog/posthog-ios/blob/784145ddd2f7de3a6d6fcce0e50c04f29d33b338/PostHog/PostHogRemoteConfig.swift#L262-L266

Could the SDK differentiate between network errors vs. server errors?

ioannisj:

Closes #367 

Do not clear local flags on remote config error or when `hasFeatureFlags` is not `false`

Added a couple of unit tests

## :pencil: Checklist

### DIFF
--- a/PostHog/PostHogRemoteConfig.swift
+++ b/PostHog/PostHogRemoteConfig.swift
@@ -74,25 +74,25 @@ class PostHogRemoteConfig {
         }) {
             dispatchQueue.async {
                 self.reloadRemoteConfig { [weak self] remoteConfig in
+                    guard let self else { return }
+
                     // if there's no remote config response, skip
-                    guard let self, let remoteConfig else { return }
-
-                    // Check if the server explicitly responded with hasFeatureFlags key
-                    if let hasFeatureFlagsValue = remoteConfig[self.hasFeatureFlagsKey] {
-                        let hasFeatureFlags = hasFeatureFlagsValue as? Bool == true
-
-                        if !hasFeatureFlags {
-                            // Server responded with hasFeatureFlags: false, meaning no active flags on the account
-                            clearFeatureFlags()
-                            // need to notify cause people may be waiting for flags to load
-                            notifyFeatureFlags([:])
-                            return
-                        }
+                    guard let remoteConfig else {
+                        hedgeLog("Remote config response is missing, skipping loading flags")
+                        notifyFeatureFlags(nil)
+                        return
                     }
 
-                    // If we reach here, either hasFeatureFlags is true or not present
-                    if self.config.preloadFeatureFlags {
-                        // Reload feature flags based on config
+                    // Check if the server explicitly responded with hasFeatureFlags key
+                    if let hasFeatureFlagsBoolValue = remoteConfig[self.hasFeatureFlagsKey] as? Bool, !hasFeatureFlagsBoolValue {
+                        hedgeLog("hasFeatureFlags is false, clearing flags and skipping loading flags")
+                        // Server responded with explicit hasFeatureFlags: false, meaning no active flags on the account
+                        clearFeatureFlags()
+                        // need to notify cause people may be waiting for flags to load
+                        notifyFeatureFlags([:])
+                    } else if self.config.preloadFeatureFlags {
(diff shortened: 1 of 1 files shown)


PostHog/PostHogRemoteConfig.swift
```diff
@@ -74,17 +74,25 @@ class PostHogRemoteConfig {
...
             dispatchQueue.async {
                 self.reloadRemoteConfig { [weak self] remoteConfig in
-                    guard let self else { return }
-
-                    let hasFeatureFlags = remoteConfig?[self.hasFeatureFlagsKey] as? Bool == true
+                    // if there's no remote config response, skip
+                    guard let self, let remoteConfig else { return }
+
+                    // Check if the server explicitly responded with hasFeatureFlags key
+                    if let hasFeatureFlagsValue = remoteConfig[self.hasFeatureFlagsKey] {
```
marandaneto: not sure if the server will ever respond without a `hasFeatureFlagsKey`, but i think we should not clear flags if thats the case just for being defensive.

PostHog/PostHogRemoteConfig.swift
```diff
@@ -74,17 +74,25 @@ class PostHogRemoteConfig {
         }) {
             dispatchQueue.async {
                 self.reloadRemoteConfig { [weak self] remoteConfig in
-                    guard let self else { return }
-
-                    let hasFeatureFlags = remoteConfig?[self.hasFeatureFlagsKey] as? Bool == true
+                    // if there's no remote config response, skip
+                    guard let self, let remoteConfig else { return }
```
marandaneto: i think we should call `notifyFeatureFlags(nil)`

marandaneto: https://github.com/PostHog/posthog-ios/blob/784145ddd2f7de3a6d6fcce0e50c04f29d33b338/PostHog/PostHogRemoteConfig.swift#L159-L161
is another case of logging + calling notifyFeatureFlags(nil)

PostHog/PostHogRemoteConfig.swift
ioannisj: Yeah being a little bit defensive here, but in case hasFeatureFlags is missing for some reason, we'll try to load flags anw

PostHog/PostHogRemoteConfig.swift
```diff
@@ -74,17 +74,25 @@ class PostHogRemoteConfig {
...
+
+                    // Check if the server explicitly responded with hasFeatureFlags key
+                    if let hasFeatureFlagsValue = remoteConfig[self.hasFeatureFlagsKey] {
+                        let hasFeatureFlags = hasFeatureFlagsValue as? Bool == true
```
marandaneto: if for some reason this is not a `Bool`, it will evaluate to `false` and we will clear flags, maybe the right solution here would be to fallback to `preloadFeatureFlags` instead

PostHog/PostHogRemoteConfig.swift
ioannisj: Done

PostHog/PostHogRemoteConfig.swift
ioannisj: True, done

marandaneto (approved):

PR #373, opened 2025-07-21
fix: Cache and pass person and group properties

### LINKED ISSUE #366: Attach person and group properties to /flags request
### Description

Follow up from this discussion here https://github.com/PostHog/posthog-ios/issues/358#issuecomment-2994984674

- [x] iOS
- [ ] Android
- [ ] React Native

haacked:
Fixes #366

This fixes a race condition where feature flags return incorrect values when person properties haven't been processed by server-side ingestion yet. Previously, the `/flags` API only sent `distinct_id`, `$anon_distinct_id`, and `$groups` but not person or group properties, causing flags to be evaluated without the latest property values.

This implementation allows person and group properties to be sent directly in feature flag requests, eliminating the timing dependency on async ingestion.

- Added new SDK methods `setPersonProperties()` and `setGroupProperties()` with comprehensive documentation
- Updated PostHogExample app with test function demonstrating both person and group properties
- Verified using Proxyman that `person_properties` and `group_properties` are correctly sent in `/flags` API requests
- Tested thread-safe property storage and automatic flag reloading
- Confirmed properties are cleared on `reset()` and properly integrated with `identify()` calls

## :pencil: Checklist


haacked:
Linting errors unrelated to this PR. Got a fix for that: https://github.com/PostHog/posthog-ios/pull/374

haritowa:
Haha I just though of implementing it :D

PostHog/PostHogRemoteConfig.swift
```diff
@@ -251,9 +257,14 @@ class PostHogRemoteConfig {
...
         }
 
+        let personProperties = getPersonPropertiesForFlags()
+        let groupProperties = getGroupPropertiesForFlags()
+
         api.flags(distinctId: distinctId,
                   anonymousId: anonymousId,
-                  groups: groups)
+                  groups: groups,
+                  personProperties: personProperties.isEmpty ? nil : personProperties,
```
ioannisj: This ternary check here is redundant given the existing !isEmpty check inside the method, but harmless.

PostHog/PostHogRemoteConfig.swift
```diff
@@ -442,6 +453,55 @@ class PostHogRemoteConfig {
         return flags
     }
 
+    func setPersonPropertiesForFlags(_ properties: [String: Any]) {
+        personPropertiesForFlagsLock.withLock {
+            // Merge properties additively, similar to JS SDK behavior
+            for (key, value) in properties {
```
ioannisj: Usually done by `personPropertiesForFlags.merge(properties, uniquingKeysWith: {_, new in new})` but this will work as well

PostHog/PostHogSDK.swift
```diff
@@ -468,6 +472,9 @@ let maxRetryDelay = 30.0
 
             queue.add(event)
 
+            // Automatically set person properties for feature flags during identify event
+            setPersonPropertiesForFlagsIfNeeded(userProperties)
```
ioannisj: Should we be doing the same for groupProperties in group() call?

PostHog/PostHogRemoteConfig.swift
```diff
@@ -442,6 +453,55 @@ class PostHogRemoteConfig {
         return flags
     }
 
+    func setPersonPropertiesForFlags(_ properties: [String: Any]) {
```
ioannisj, line 462: My initial intuition was that both person and group properties would also be cached and managed in disk, not only memory, since technically people can skip identify call on next launch if distinct_id has not changed? Is that not needed?

PostHog/PostHogSDK.swift
```diff
@@ -831,6 +847,146 @@ let maxRetryDelay = 30.0
...
+    ///     "plan": "premium"
+    /// ])
+    ///
+    /// // Now feature flags will be evaluated with these properties
+    /// let flagValue = PostHogSDK.shared.isFeatureEnabled("new_feature")
+    /// ```
+    ///
+    /// - Parameter properties: Dictionary of person properties to include in flag evaluation
+    /// - SeeAlso: `setPersonPropertiesForFlags(_:reloadFeatureFlags:)` to control flag reloading behavior
+    @objc public func setPersonPropertiesForFlags(_ properties: [String: Any]) {
```
ioannisj, line 924: Is the intent here with this family of methods to allow users to manage person and group properties specifically for flags, while skipping the $identify, $set, and $groupidentify events (since we are exposing as public APIs)?

My first thought and assumption was that we’d want to set those properties automatically for flags when users interact with the APIs that actually capture those events. e.g identify() and group()

PostHog/PostHogRemoteConfig.swift
```diff
@@ -442,6 +453,55 @@ class PostHogRemoteConfig {
...
+
+    func setGroupPropertiesForFlags(_ groupType: String, properties: [String: Any]) {
+        groupPropertiesForFlagsLock.withLock {
+            // Merge properties additively for this group type
+            if groupPropertiesForFlags[groupType] == nil {
+                groupPropertiesForFlags[groupType] = [:]
+            }
+            for (key, value) in properties {
+                groupPropertiesForFlags[groupType]![key] = value
+            }
```
ioannisj: We can use this swift sugar here to avoid force unwrapping

```suggestion
            for (key, value) in properties {
                groupPropertiesForFlags[groupType, default: [:]][key] = value
            }
```

PostHog/PostHogSDK.swift
```diff
@@ -351,6 +351,10 @@ let maxRetryDelay = 30.0
         }
         PostHogSessionManager.shared.resetSession()
 
+        // Clear person and group properties for flags
+        remoteConfig?.resetPersonPropertiesForFlags()
+        remoteConfig?.resetGroupPropertiesForFlags()
```
marandaneto, line 358: not 100% sure but i think `storage?.reset(...)` above would clear the storage already, @ioannisj can you confirm?

PostHog/PostHogApi.swift
```diff
@@ -196,6 +196,8 @@ class PostHogApi {
         distinctId: String,
         anonymousId: String?,
         groups: [String: String],
+        personProperties: [String: Any]? = nil,
```
marandaneto:
```suggestion
        personProperties: [String: Any],
```

can probably be non optional since we will attach default person props anyway

PostHog/PostHogRemoteConfig.swift
marandaneto: I'd remove, less code/less runtime cost

PostHog/PostHogRemoteConfig.swift
marandaneto: `merging` or `merge` are cleaner, less code.

PostHog/PostHogSDK.swift
marandaneto: yes, reason is here https://posthog.com/docs/libraries/js/features#overriding-server-properties

PostHog/PostHogSDK.swift
ioannisj: See comment https://github.com/PostHog/posthog-ios/pull/373#discussion_r2224347587

We don't currently disk cache person and group properties and we are not reading from disk either (unless I missed something) I think this should change, but waiting for comment from @haacked

PostHog/PostHogSDK.swift
marandaneto: we should `sanitizeDictionary(userProperties)` here too since we do that already, we can cache in a var and reuse the result here

PostHog/PostHogSDK.swift
```diff
@@ -478,6 +485,9 @@ let maxRetryDelay = 30.0
                     userProperties: userProperties,
                     userPropertiesSetOnce: userPropertiesSetOnce)
 
+            // Automatically set person properties for feature flags
+            setPersonPropertiesForFlagsIfNeeded(userProperties)
```
marandaneto: same as https://github.com/PostHog/posthog-ios/pull/373/files#r2224586379

PostHog/PostHogApi.swift
ioannisj: 👍 if-lets inside this method should also be changed then

PostHog/PostHogRemoteConfig.swift
```diff
@@ -442,6 +453,55 @@ class PostHogRemoteConfig {
...
+        }
+    }
+
+    func resetGroupPropertiesForFlags(_ groupType: String? = nil) {
```
marandaneto, line 483:
```suggestion
    func resetGroupPropertiesForFlags(_ groupType: String) {
```

you need a type otherwise there's nothing to do, the caller should avoid the call in this case

PostHog/PostHogSDK.swift
marandaneto: yes, `groupIdentify` has to do the same

PostHog/PostHogSDK.swift
```diff
@@ -831,6 +847,146 @@ let maxRetryDelay = 30.0
...
+    ///   - properties: Dictionary of person properties to include in flag evaluation
+    ///   - reloadFeatureFlags: Whether to automatically reload feature flags after setting properties
+    @objc(setPersonPropertiesForFlags:reloadFeatureFlags:)
+    public func setPersonPropertiesForFlags(_ properties: [String: Any], reloadFeatureFlags: Bool) {
+        if !isEnabled() {
+            return
+        }
+
+        let sanitizedProperties = sanitizeDictionary(properties) ?? [:]
+        remoteConfig?.setPersonPropertiesForFlags(sanitizedProperties)
```
marandaneto, line 963: I'd only call `setPersonPropertiesForFlags` if valid and empty.
Then I'd only `reloadFeatureFlags` inside this condition as well, to not reload for nothing

PostHog/PostHogSDK.swift
```diff
@@ -831,6 +847,146 @@ let maxRetryDelay = 30.0
...
+    /// PostHogSDK.shared.resetGroupPropertiesForFlags("organization")
+    ///
+    /// // Clear all group properties
+    /// PostHogSDK.shared.resetGroupPropertiesForFlags()
+    /// ```
+    ///
+    /// - Parameter groupType: The group type to clear properties for, or nil to clear all group properties
+    /// - Note: This method does not automatically reload feature flags. Call `reloadFeatureFlags()`
+    ///         after resetting if you want to immediately refresh flags with the cleared properties.
+    @objc public func resetGroupPropertiesForFlags(_ groupType: String? = nil) {
```
marandaneto, line 983:
```suggestion
    @objc public func resetGroupPropertiesForFlags(_ groupType: String) {
```

marandaneto: @ioannisj do we need a method overload here for better objC uX? otherwise they have to forcefully pass (nil)

PostHog/PostHogSDK.swift
```diff
@@ -831,6 +847,146 @@ let maxRetryDelay = 30.0
...
+    ///
+    /// // Manually reload flags later
+    /// PostHogSDK.shared.reloadFeatureFlags()
+    /// ```
+    ///
+    /// - Parameters:
+    ///   - properties: Dictionary of person properties to include in flag evaluation
+    ///   - reloadFeatureFlags: Whether to automatically reload feature flags after setting properties
+    @objc(setPersonPropertiesForFlags:reloadFeatureFlags:)
+    public func setPersonPropertiesForFlags(_ properties: [String: Any], reloadFeatureFlags: Bool) {
```
marandaneto: we can make `reloadFeatureFlags = true` as default so you dont have to pass the value, less code.
We should also add an overload for this method for better objC UX only with the `properties` param.

PostHog/PostHogSDK.swift
```diff
@@ -831,6 +847,146 @@ let maxRetryDelay = 30.0
...
+    /// ], reloadFeatureFlags: false)
+    ///
+    /// // Manually reload flags later
+    /// PostHogSDK.shared.reloadFeatureFlags()
+    /// ```
+    ///
+    /// - Parameters:
+    ///   - properties: Dictionary of person properties to include in flag evaluation
+    ///   - reloadFeatureFlags: Whether to automatically reload feature flags after setting properties
+    @objc(setPersonPropertiesForFlags:reloadFeatureFlags:)
```
marandaneto:
```suggestion
    @objc(setPersonPropertiesForFlagsWithProperties:reloadFeatureFlags:)
```

PostHog/PostHogSDK.swift
ioannisj: Yeah, similar to this comment here https://github.com/PostHog/posthog-ios/pull/373#discussion_r2224619524

PostHog/PostHogSDK.swift
```diff
@@ -831,6 +847,146 @@ let maxRetryDelay = 30.0
...
+    ///     "seats": 50,
+    ///     "industry": "technology"
+    /// ])
+    /// ```
+    ///
+    /// - Parameters:
+    ///   - groupType: The group type identifier (e.g., "organization", "team")
+    ///   - properties: Dictionary of properties to set for this group type
+    /// - Note: This method automatically reloads feature flags to apply the new properties.
+    @objc public func setGroupPropertiesForFlags(_ groupType: String, properties: [String: Any]) {
```
marandaneto: we need the `@objc` with the param names because there are multiple params, ObjC interop.

PostHog/PostHogSDK.swift
marandaneto: do we need to cache and pass `userPropertiesSetOnce` as well? how is that handled on JS?

marandaneto: the `capture` method also takes the userProps and userSetOnceProps, so we need to call this there too

marandaneto (approved):
left a few comments, we are on the right direction.

marandaneto (changes requested):
left a few comments, we are on the right direction.

PostHog/PostHogSDK.swift
marandaneto: correct, its not flushing to the disk.
its ok to keep a copy in memory to avoid IO every single time its needed, but right now its not caching at all

PostHog/PostHogSDK.swift
```diff
@@ -493,6 +503,12 @@ let maxRetryDelay = 30.0
         return false
     }
 
+    private func setPersonPropertiesForFlagsIfNeeded(_ userProperties: [String: Any]?) {
```
marandaneto: should we only allow doing this and group props if `personProfiles` is != `never`? or check for `hasPersonProcessing`

PostHog/PostHogSDK.swift
haacked: Yes, we should. I added that and followed what the JS lib does.

PostHog/PostHogRemoteConfig.swift
haacked: Good call. Made that change.

PostHog/PostHogSDK.swift
haacked: Implemented caching.

PostHog/PostHogSDK.swift
haacked: Done!

PostHog/PostHogRemoteConfig.swift
```diff
@@ -251,9 +257,14 @@ class PostHogRemoteConfig {
...
 
+        let personProperties = getPersonPropertiesForFlags()
+        let groupProperties = getGroupPropertiesForFlags()
+
         api.flags(distinctId: distinctId,
                   anonymousId: anonymousId,
-                  groups: groups)
+                  groups: groups,
+                  personProperties: personProperties.isEmpty ? nil : personProperties,
+                  groupProperties: groupProperties.isEmpty ? nil : groupProperties)
```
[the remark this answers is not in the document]
haacked: Uh, looks aligned to me.

haacked:
@ioannisj @marandaneto thank you so much for your reviews! It's teaching me a lot about mobile development. Would you mind taking another look? I think I addressed all your feedback.

PostHog/PostHogRemoteConfig.swift
```diff
@@ -442,6 +456,70 @@ class PostHogRemoteConfig {
...
+            if let cachedPersonProperties = storage.getDictionary(forKey: .personPropertiesForFlags) as? [String: Any] {
+                personPropertiesForFlags = cachedPersonProperties
+            }
+        }
+
+        groupPropertiesForFlagsLock.withLock {
+            if let cachedGroupProperties = storage.getDictionary(forKey: .groupPropertiesForFlags) as? [String: [String: Any]] {
+                groupPropertiesForFlags = cachedGroupProperties
+            }
+        }
```
ioannisj, line 532: nit: I would lock inside the if-let to avoid unnecessary locks

PostHog/PostHogSDK.swift
```diff
@@ -478,6 +485,9 @@ let maxRetryDelay = 30.0
                     userProperties: userProperties,
                     userPropertiesSetOnce: userPropertiesSetOnce)
 
+            // Automatically set person properties for feature flags during user property updates
+            setPersonPropertiesForFlagsIfNeeded(userProperties, userPropertiesSetOnce: userPropertiesSetOnce)
```
ioannisj, line 491: @haacked just confirming you saw these comments [here](https://github.com/PostHog/posthog-ios/pull/373#discussion_r2224629368). I assume we don't need to do the same for $groupidentify and `capture()` method?

PostHog/PostHogRemoteConfig.swift
marandaneto: its better the way it is since the `storage` has no locks so we lock the `read` op as well

PostHog/PostHogSDK.swift
marandaneto: @ioannisj https://github.com/PostHog/posthog-ios/pull/373/files#r2224570259
whats about now with caching? do we still need to call these 2 methods?

PostHog/PostHogRemoteConfig.swift
ioannisj: Ah totally true, I stand corrected!

PostHog/PostHogRemoteConfig.swift
```diff
@@ -442,6 +456,70 @@ class PostHogRemoteConfig {
...
+            // Persist to disk
+            storage.setDictionary(forKey: .personPropertiesForFlags, contents: personPropertiesForFlags)
+        }
+    }
+
+    func resetPersonPropertiesForFlags() {
+        personPropertiesForFlagsLock.withLock {
+            personPropertiesForFlags.removeAll()
+            // Clear from disk
+            storage.setDictionary(forKey: .personPropertiesForFlags, contents: personPropertiesForFlags)
```
marandaneto, line 472: i think this should be centralized here https://github.com/PostHog/posthog-ios/blob/03803c80ae67695667a8106c98af92caac94e81f/PostHog/PostHogStorage.swift#L352
same for groups

PostHog/PostHogSDK.swift
ioannisj: I think so, so we can clear mem as well?

PostHog/PostHogSDK.swift
```diff
@@ -831,6 +862,164 @@ let maxRetryDelay = 30.0
...
+    ///   - reloadFeatureFlags: Whether to automatically reload feature flags after setting properties
+    @objc(setPersonPropertiesForFlagsWithProperties:reloadFeatureFlags:)
+    public func setPersonPropertiesForFlags(_ properties: [String: Any], reloadFeatureFlags: Bool = true) {
+        if !isEnabled() {
+            return
+        }
+
+        let sanitizedProperties = sanitizeDictionary(properties) ?? [:]
+        guard !sanitizedProperties.isEmpty else { return }
+        remoteConfig?.setPersonPropertiesForFlags(sanitizedProperties)
```
marandaneto, line 963: should we call `setPersonPropertiesForFlagsIfNeeded` instead?
otherwise we dont check for eg `hasPersonProcessing`

PostHog/PostHogSDK.swift
marandaneto: still missing

PostHog/PostHogStorage.swift
```diff
@@ -212,6 +212,8 @@ class PostHogStorage {
         case remoteConfig = "posthog.remoteConfig"
         case surveySeen = "posthog.surveySeen"
         case requestId = "posthog.requestId"
+        case personPropertiesForFlags = "posthog.personPropertiesForFlags"
+        case groupPropertiesForFlags = "posthog.groupPropertiesForFlags"
```
ioannisj, line 216: Actually https://github.com/PostHog/posthog-ios/pull/373/files#r2227858263 made me click that we need these two keys in `reset(keepAnonymousId:)` as well

PostHog/PostHogSDK.swift
```diff
@@ -831,6 +862,164 @@ let maxRetryDelay = 30.0
...
+    ///   - properties: Dictionary of properties to set for this group type
+    /// - Note: This method automatically reloads feature flags to apply the new properties.
+    @objc public func setGroupPropertiesForFlags(_ groupType: String, properties: [String: Any]) {
+        if !isEnabled() {
+            return
+        }
+
+        let sanitizedProperties = sanitizeDictionary(properties) ?? [:]
+        guard !sanitizedProperties.isEmpty else { return }
+        remoteConfig?.setGroupPropertiesForFlags(groupType, properties: sanitizedProperties)
```
marandaneto, line 1053: we need something similar to `setPersonPropertiesForFlagsIfNeeded` here so we dont dupe the code.
this should be called within the `group` method too.
also, should groups also be guarded by `hasPersonProcessing`? not sure if it depends on persons

PostHog/PostHogSDK.swift
```diff
@@ -831,6 +862,164 @@ let maxRetryDelay = 30.0
...
+
+        // Automatically reload flags to apply the new properties
+        remoteConfig?.reloadFeatureFlags()
```
marandaneto: should we add a method param similar to `reloadFeatureFlags`? so people can disable if needed

marandaneto (commented):
left a few final comments, almost there.

PostHog/PostHogSDK.swift
marandaneto: It's correct, just checked, due to the public API resetting only person props and group props, so clearing cache and disk is in one single place.

PostHog/PostHogSDK.swift
ioannisj: Tested this and ObjC doesn't seem to complain actually (was under the impression as well)

Sample app seems to compile and run just fine with: 

```objc
    [[PostHogSDK shared] setPersonPropertiesForFlags:@{@"prop": @"value"}];
    [[PostHogSDK shared] setPersonPropertiesForFlags:@{@"prop": @"value"} reloadFeatureFlags:NO];
```

marandaneto: Its needed with method overloads, in this case its not but then the generated method names are different than the others.
Its more about semantics.

marandaneto: If both params would be String, you wouldn't know what comes first

PostHog/PostHogSDK.swift
haacked: Ah, yeah, we need to do it for both.

PostHog/PostHogSDK.swift
haacked: I'll add a check for hasPersonProperties() here.

PostHog/PostHogSDK.swift
haacked: > also, should groups also be guarded by `hasPersonProcessing`? not sure if it depends on persons

According to [our docs here](https://posthog.com/docs/product-analytics/group-analytics):

> Events must be [identified](https://posthog.com/docs/data/anonymous-vs-identified-events) to be associated with a group. If the `$process_person_profile event` property ends up being set to false, the event will not be associated with the group.

So I believe this guard is correct.

haacked:
Ok, I merged https://github.com/PostHog/posthog-ios/pull/375 into this branch so it all goes in at once.

PR #375, opened 2025-07-22
feat: Support default properties

haacked:
This is a follow-up to #373 (and targets that branch)

Feature flags that rely on person properties like `$app_version` or `$os_version` often fail to work immediately after SDK initialization because these properties aren't available until `identify()` calls are processed server-side. This creates a race condition where flags return incorrect values during the critical first moments of app usage.

This PR implements automatic person properties similar to the PostHog web SDK, where common device and app properties are automatically set as person properties during SDK initialization.

**Properties automatically set:**
- `$app_version` - App version from bundle
(`CFBundleShortVersionString`)
- `$app_build` - App build number (`CFBundleVersion`)
- `$os_name` - Operating system name (iOS, macOS, etc.)
- `$os_version` - Operating system version
- `$device_type` - Device type (Mobile, Tablet, Desktop, etc.)
- `$locale` - User's current locale

**Key benefits:**
- Eliminates race conditions for feature flag evaluation
- Ensures flags work immediately without waiting for server-side processing
- Maintains consistency with PostHog web SDK behavior
- Includes opt-out configuration for flexibility

- Added comprehensive test suite covering enabled/disabled states and app update scenarios
- Verified automatic property setting during SDK initialization via flag request inspection
- Tested that properties are refreshed on app updates
- Confirmed opt-out functionality works correctly
- Updated example app with usage documentation and testing functionality
- All existing tests continue to pass

## :pencil: Checklist

### DIFF
--- a/PostHog/App Life Cycle/PostHogAppLifeCycleIntegration.swift
+++ b/PostHog/App Life Cycle/PostHogAppLifeCycleIntegration.swift
@@ -142,11 +142,6 @@ final class PostHogAppLifeCycleIntegration: PostHogIntegration {
         }
 
         postHog.capture(event, properties: props)
-
-        // Refresh default person properties on app updates to ensure they're current
-        if event == "Application Updated" {
-            postHog.refreshDefaultPersonProperties()
-        }
     }
 
     private func captureAppOpened() {
--- a/PostHog/App Life Cycle/PostHogAppLifeCycleIntegration.swift
+++ b/PostHog/App Life Cycle/PostHogAppLifeCycleIntegration.swift
@@ -142,6 +142,11 @@ final class PostHogAppLifeCycleIntegration: PostHogIntegration {
         }
 
         postHog.capture(event, properties: props)
+
+        // Refresh default person properties on app updates to ensure they're current
+        if event == "Application Updated" {
+            postHog.refreshDefaultPersonProperties()
+        }
     }
 
     private func captureAppOpened() {
--- a/PostHog/PostHogSDK.swift
+++ b/PostHog/PostHogSDK.swift
@@ -102,7 +102,9 @@ let maxRetryDelay = 30.0
             let api = PostHogApi(config)
 
             config.storageManager = config.storageManager ?? PostHogStorageManager(config)
-            remoteConfig = PostHogRemoteConfig(config, theStorage, api)
+            remoteConfig = PostHogRemoteConfig(config, theStorage, api) { [weak self] in
+                self?.getDefaultPersonProperties() ?? [:]
+            }
 
             #if !os(watchOS)
                 do {
@@ -140,9 +142,6 @@ let maxRetryDelay = 30.0
                 // don't install integrations if in opt-out state
                 installIntegrations()
             }
-
-            // Set default person properties after setup is complete
-            setDefaultPersonProperties()
 
             DispatchQueue.main.async {
                 NotificationCenter.default.post(name: PostHogSDK.didStartNotification, object: nil)
@@ -527,12 +526,12 @@ let maxRetryDelay = 30.0
         remoteConfig?.setPersonPropertiesForFlags(allProperties)
     }
 
-    /// Automatically sets default device and app properties as person properties for feature flag evaluation.
-    /// This ensures feature flags can immediately use properties like $app_version, $os_version, etc.
-    /// without waiting for server-side processing of identify() calls.
-    private func setDefaultPersonProperties() {
-        guard config.setDefaultPersonProperties else { return }
-        guard isEnabled() else { return }
+    /// Returns fresh default device and app properties for feature flag evaluation.
+    /// This ensures feature flags can use current properties like $app_version, $os_version, etc.
+    /// These properties are computed fresh each time they're needed.
+    private func getDefaultPersonProperties() -> [String: Any] {
+        guard config.setDefaultPersonProperties else { return [:] }
+        guard isEnabled() else { return [:] }
 
         let staticContext = context?.staticContext() ?? [:]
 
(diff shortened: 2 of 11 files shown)


PostHog/PostHogSDK.swift
```diff
@@ -509,6 +512,56 @@ let maxRetryDelay = 30.0
...
+
+        if !defaultProperties.isEmpty {
+            let sanitizedProperties = sanitizeDictionary(defaultProperties) ?? [:]
+            remoteConfig?.setPersonPropertiesForFlags(sanitizedProperties)
+        }
+    }
+
+    /// Refreshes the default person properties, typically called during app updates.
+    /// This ensures that properties like $app_version and $app_build are current.
+    internal func refreshDefaultPersonProperties() {
```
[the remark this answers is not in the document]
haacked: this is not applicable because the method is called from PostHogAppLifeCycleIntegration, which is in a separate file but same module, so it needs to remain internal.

PostHog/App Life Cycle/PostHogAppLifeCycleIntegration.swift
```diff
@@ -142,6 +142,11 @@ final class PostHogAppLifeCycleIntegration: PostHogIntegration {
         }
 
         postHog.capture(event, properties: props)
+
+        // Refresh default person properties on app updates to ensure they're current
+        if event == "Application Updated" {
+            postHog.refreshDefaultPersonProperties()
+        }
```
ioannisj: Not sure if this is indeed needed since an `Application Updated` can only happen on a fresh app launch (apps cannot be update while running). We are already calling setDefaultPersonProperties() on SDK init which will already have the most recent $app_version value

PostHog/PostHogSDK.swift
```diff
@@ -141,6 +141,9 @@ let maxRetryDelay = 30.0
                 installIntegrations()
             }
 
+            // Set default person properties after setup is complete
+            setDefaultPersonProperties()
```
marandaneto: We should not cache this in storage, but instead read it at runtime every time, since this could lead to incorrect values.
For example, right now we only update the values if `captureApplicationLifecycleEvents` is enabled which would be a bug already.
The reading cost is minimal since these values are read once and stored in memory during the app's lifecycle.

PostHog/PostHogSDK.swift
```diff
@@ -509,6 +512,60 @@ let maxRetryDelay = 30.0
...
+            defaultProperties["$device_type"] = deviceType
+        }
+
+        // Localization
+        if let locale = dynamicContext["$locale"] {
+            defaultProperties["$locale"] = locale
+        }
+
+        if !defaultProperties.isEmpty {
+            let sanitizedProperties = sanitizeDictionary(defaultProperties)
```
marandaneto: We don't need to sanitize those values. This is used for values given by the user that can or not be serializable/JSON valid

PostHog/PostHogSDK.swift
```diff
@@ -509,6 +512,60 @@ let maxRetryDelay = 30.0
...
 
+    /// Automatically sets default device and app properties as person properties for feature flag evaluation.
+    /// This ensures feature flags can immediately use properties like $app_version, $os_version, etc.
+    /// without waiting for server-side processing of identify() calls.
+    private func setDefaultPersonProperties() {
+        guard config.setDefaultPersonProperties else { return }
+        guard isEnabled() else { return }
+
+        let staticContext = context?.staticContext() ?? [:]
+        let dynamicContext = context?.dynamicContext() ?? [:]
```
marandaneto: Since we only read the `locale` here and this has a higher cost than `staticContext`, it'd make sense to add a param to `dynamicContext` which would skip reading all the unnecessary values, eg just returns `locale` in this case

marandaneto (changes requested):
left a few comments

PostHog/PostHogSDK.swift
haacked: Ok, I addressed this.

PostHog/PostHogSDK.swift
```diff
@@ -524,6 +526,52 @@ let maxRetryDelay = 30.0
...
+            defaultProperties["$os_version"] = osVersion
+        }
+
+        // Device information
+        if let deviceType = staticContext["$device_type"] {
+            defaultProperties["$device_type"] = deviceType
+        }
+
+        // Localization - read directly to avoid expensive dynamicContext call
+        if #available(iOS 16.0, macOS 13.0, tvOS 16.0, watchOS 9.0, *) {
```
marandaneto, line 579: nit: extract this to a [utils](https://github.com/PostHog/posthog-ios/tree/main/PostHog/Utils) method and reuse here and [here](https://github.com/PostHog/posthog-ios/blob/03803c80ae67695667a8106c98af92caac94e81f/PostHog/PostHogContext.swift#L181-L189)

PostHog/PostHogSDK.swift
```diff
@@ -524,6 +526,52 @@ let maxRetryDelay = 30.0
...
+    /// Returns fresh default device and app properties for feature flag evaluation.
+    /// This ensures feature flags can use current properties like $app_version, $os_version, etc.
+    /// These properties are computed fresh each time they're needed.
+    private func getDefaultPersonProperties() -> [String: Any] {
+        guard config.setDefaultPersonProperties else { return [:] }
+        guard isEnabled() else { return [:] }
+
+        let staticContext = context?.staticContext() ?? [:]
+
+        var defaultProperties: [String: Any] = [:]
```
marandaneto, line 555: Another option would be to add a new method to `context` that only returns a subset of all contexts, eg `context?.personPropertiesContext()` so all the properties reading are centralized in the `context` class.
It's easier to refactor if needed later on since we dont have hardcoded key diplicated in multiple places

PostHog/PostHogSDK.swift
marandaneto: or address https://github.com/PostHog/posthog-ios/pull/375#discussion_r2227644578
so this can live within the context class

marandaneto (approved):
left a last comment otherwise LGTM

PR #377, opened 2025-08-12
fix: map thank you message description content type

ioannisj:

When looking into https://github.com/PostHog/posthog-flutter/issues/194 I realised we were not mapping `thankYouMessageDescriptionContentType` to appearance display model

## :pencil: Checklist


PostHog/Surveys/Models/PostHogDisplaySurveyAppearance.swift
```diff
@@ -52,21 +52,21 @@ import Foundation
     public let thankYouMessageCloseButtonText: String?
 
     init(
-        fontFamily: String? = nil,
```
ioannisj, line 55: internal init so no changes to public api. Just removed defaults

marandaneto (approved):

PR #381, opened 2025-08-29
feat: surveys GA

ioannisj:
## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Next
 
 - feat: surveys GA ([#381](https://github.com/PostHog/posthog-ios/pull/381))
+> Note: Surveys are now enabled by default
 
 ## 3.30.1 - 2025-08-12
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Next
 
+- feat: surveys GA ([#381](https://github.com/PostHog/posthog-ios/pull/381))
+
 ## 3.30.1 - 2025-08-12
 
 - fix: map missing content type for Surveys ([#377](https://github.com/PostHog/posthog-ios/pull/377))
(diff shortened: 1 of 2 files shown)


PostHog/PostHogConfig.swift
```diff
@@ -202,7 +199,7 @@ public typealias BeforeSendBlock = (PostHogEvent) -> PostHogEvent?
         return integrations
     }
 
-    var _surveys: Bool = false // swiftlint:disable:this identifier_name
+    var _surveys: Bool = true // swiftlint:disable:this identifier_name
```
marandaneto, line 202: i would still keep it as `false` by default, so people have to turn it on

marandaneto: or should it be enabled by default? good question

marandaneto: RN does not have this because you actually need the `PostHogSurveyProvider` provider, so its the same as flipping the flag to enabled

marandaneto: default to enabled on the web, so i guess its ok

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## Next
 
+- feat: surveys GA ([#381](https://github.com/PostHog/posthog-ios/pull/381))
```
marandaneto, line 3: i'd mention `surveys` is now enabled by default as well

PostHog/PostHogConfig.swift
marandaneto: have to fix here as well https://posthog.com/docs/libraries/ios#all-configuration-options

> /**
>  * Enable Surveys. (off/false by default)
>  * Experimental feature under @_spi(Experimental)
>  */
> configuration.surveys = false

marandaneto (approved):

PR #383, opened 2025-09-25
feat: iOS surveys use the new response question id format

### LINKED ISSUE #323: Surveys - add support for question ids
### Description

From discussion: https://github.com/PostHog/posthog-ios/pull/320#discussion_r2014329715

Support new format for `$survey_questions` (which should include { id, name, response }) and `$survey_response_x` properties

- [x] [iOS](https://github.com/PostHog/posthog-ios/pull/383)
- [x] [Android](https://github.com/PostHog/posthog-android/pull/289)
- [x] [React Native](https://github.com/PostHog/posthog-js/pull/2352)
- [x] [Flutter](https://github.com/PostHog/posthog-flutter/pull/210)

marandaneto:
follow up https://github.com/PostHog/posthog-js/pull/2352/
Closes https://github.com/PostHog/posthog-ios/issues/323

Screenshot 2025-09-25 at 14 23 19

## :pencil: Checklist

### DIFF
--- a/PostHog/Models/Surveys/PostHogSurvey+Display.swift
+++ b/PostHog/Models/Surveys/PostHogSurvey+Display.swift
@@ -19,6 +19,7 @@
             switch self {
             case let .open(question):
                 return PostHogDisplayOpenQuestion(
+                    id: question.id,
                     question: question.question,
                     questionDescription: question.description,
                     questionDescriptionContentType: question.descriptionContentType?.toDisplayContentType(),
@@ -28,6 +29,7 @@
 
             case let .link(question):
                 return PostHogDisplayLinkQuestion(
+                    id: question.id,
                     question: question.question,
                     questionDescription: question.description,
                     questionDescriptionContentType: question.descriptionContentType?.toDisplayContentType(),
@@ -38,6 +40,7 @@
 
             case let .rating(question):
                 return PostHogDisplayRatingQuestion(
+                    id: question.id,
                     question: question.question,
                     questionDescription: question.description,
                     questionDescriptionContentType: question.descriptionContentType?.toDisplayContentType(),
@@ -52,6 +55,7 @@
 
             case let .singleChoice(question), let .multipleChoice(question):
                 return PostHogDisplayChoiceQuestion(
+                    id: question.id,
                     question: question.question,
                     questionDescription: question.description,
                     questionDescriptionContentType: question.descriptionContentType?.toDisplayContentType(),
--- a/PostHog/Surveys/PostHogSurveyIntegration.swift
+++ b/PostHog/Surveys/PostHogSurveyIntegration.swift
@@ -400,7 +400,7 @@
                 hedgeLog("[Surveys] Received a response event for a non-active survey")
                 return nil
             }
-            
+
             // TODO: ideally the handleSurveyResponse should pass the question ID as param but it would break the Flutter SDK for older versions
             let questionId: String
             if index < survey.questions.count {
@@ -494,7 +494,7 @@
                 let responseKey = question.id.isEmpty ? getOldResponseKey(for: index) : getNewResponseKey(for: question.id)
                 var questionData: [String: Any] = [
                     "id": question.id,
-                    "question": question.question
+                    "question": question.question,
                 ]
 
                 if let response = responsesProperties[responseKey] {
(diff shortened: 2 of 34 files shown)


PostHog/Surveys/PostHogSurveyIntegration.swift
```diff
@@ -499,7 +523,6 @@
         /// Sends a `survey dismissed` event to PostHog instance
         private func sendSurveyDismissedEvent(survey: PostHogSurvey) {
             let additionalProperties: [String: Any] = [
-                "$survey_questions": survey.questions.map(\.question),
```
marandaneto, line 502: survey dismissed should not have `$survey_questions`

lucasheriques: in the posthog-js, it also includes the `$survey_questions` property.
it's useful for getting all responses up until when the user dismissed it

lucasheriques (approved):
code looks good, pr is still in draft though

but as mentioned, I'd also include the $survey_questions in the dismissed event, so there's a way to get the responses in case someone answers 1 question (in a 3 question survey) and then dismisses it

PostHog/Surveys/PostHogSurveyIntegration.swift
marandaneto: mmm i dont follow because our APIs only query for survey sent events when getting responses, how is that useful?

ioannisj: if I recall correctly, I copied over from what posthog-js ws doing, thinking that it's part of the partial response feature (was wip back then if I remember correctly), but good point on APIs above. Let's wait for @lucasheriques to comment. If we are not using this anywhere, we can drop from all sdks?

PostHog/Surveys/PostHogSurveyIntegration.swift
```diff
@@ -567,16 +590,23 @@
...
             response: PostHogSurveyResponse,
             nextQuestion: PostHogNextSurveyQuestion
         ) -> [String: PostHogSurveyResponse] {
             activeSurveyLock.withLock {
-                activeSurveyResponses[getResponseKey(for: index)] = response
+                // keeping the old response key format for back compatibility
+                activeSurveyResponses[getOldResponseKey(for: index)] = response
+                if !id.isEmpty {
+                    // setting the new response key format
+                    activeSurveyResponses[getNewResponseKey(for: id)] = response
```
ioannisj, line 608: Should we use something similar to L494 so we don't write to dict twice?

PostHog/Surveys/PostHogSurveyIntegration.swift
lucasheriques: I think makes sense, we can drop it. When we implement partial responses for mobile SDKs (sometime in the future) we'll have access to all responses regardless. Will remove from the JS sdk too

PostHog/Models/Surveys/PostHogSurvey+Display.swift
```diff
@@ -19,6 +19,7 @@
             switch self {
             case let .open(question):
                 return PostHogDisplayOpenQuestion(
+                    id: question.id,
```
ioannisj, line 22: Since we are not using this question id in the survey delegate at all (because that would break Flutter implementation and backward compat) I think we should skip modifying the public PostHogDisplay_ family of models. We anw grab the question id on a response using the index. wdyt?

ioannisj (commented):
Left a couple of minor comments on public APIs. Also, since we are missing some unit tests on survey_sent event, would be cool to add them. I'm positive this won't break Flutter but I'll test soon

PostHog/Models/Surveys/PostHogSurvey+Display.swift
marandaneto: i think we should move away from the older behaviour, the only reason to not do that is to keep back compatibility.
this is just one step closer to moving away from that behaviour when we do the net major bump and actually pass the id as part of the delegate

PostHog/Surveys/PostHogSurveyIntegration.swift
marandaneto: they are different, one of them i am reading a key in a back compatible way and the other one i intentionally want to set both keys if possible

marandaneto:
> Left a couple of minor comments on public APIs. Also, since we are missing some unit tests on survey_sent event, would be cool to add them. I'm positive this won't break Flutter but I'll test soon

@ioannisj would you like to add a test and finish this since our hackathon started and wont have more time this week?

PostHog/Surveys/PostHogSurveyIntegration.swift
ioannisj: Ah yeah, bad read from me here

PR #387, opened 2025-10-07
feat(flags): add evaluation tags to iOS SDK

dmarticus:
## Changes

This PR implements support for evaluation environments, allowing users to specify which environment tags their SDK instance should use when evaluating feature flags.

Key changes:
  - Added evaluation_environments configuration option to PostHogConfig that accepts an array of environment tags (e.g., `['production', 'web', 'checkout']`)
  - Updated methods that call the `/flags` endpoint to include `evaluation_environments` in the request payload when configured
  - Added tests to verify the new functionality

This implementation follows the specification outlined in:
  - PostHog documentation PR: https://github.com/PostHog/posthog.com/pull/12929
  - Related internal PRs: https://github.com/PostHog/posthog/pull/38681 and https://github.com/PostHog/posthog/pull/37719

When configured, only feature flags that have at least one matching evaluation tag will be evaluated. Feature flags with no evaluation tags will always be evaluated for backward compatibility.

Basically, I'm doing what I did [here](https://github.com/PostHog/posthog-js/pull/2369), in response to a [customer request](https://posthog.slack.com/archives/C06KJQ2RH1A/p1753773309230879).

This should be perfectly safe; this field is backward compatible and won't affect existing versions of the app.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## Next
 
-## 3.33.0 - 2025-10-08
+## 3.33.0 - 2025-10-10
 
 - add evaluation tags to iOS SDK ([#387](https://github.com/PostHog/posthog-ios/pull/387))
 


PostHog/PostHogConfig.swift
```diff
@@ -98,6 +98,23 @@ public typealias BeforeSendBlock = (PostHogEvent) -> PostHogEvent?
...
+    /// Example usage:
+    /// ```swift
+    /// config.evaluationEnvironments = ["production", "web", "checkout"]
+    /// ```
+    ///
+    /// This helps ensure feature flags are only evaluated in the appropriate environments
+    /// for your SDK instance.
+    ///
+    /// Default: nil (all flags are evaluated)
+    @objc public var evaluationEnvironments: [String]?
```
ioannisj, line 116: I thought this would not work for objc interop but testing it out in ObjC sample works fine 👍 

```objc
config.evaluationEnvironments = @[@"production", @"mobile"];
```

PR #388, opened 2025-10-10
feat: add config option to disable swizzling

### LINKED ISSUE #362: Avoid swizzling
### Problem Statement

Hello 

Our team is looking into PostHog, my biggest issue is the swizzling of system functions in a third party SDK.

I understand PostHog wants to be the only analytics tool you need, but in reality you're not working in a vacuum. And the current SDK swizzles _a lot_; layoutSubviews, viewDidAppear, Application sendAction & sendEvent, scrollViews, urlSession, ...

I didn't find any settings to avoid swizzling, are there any plans to provide a 'light version'? Does anyone know how badly functionality would be impacted by disabling swizzling?

For reference, our crash reporting tool also uses swizzling but provides [a global disable setting and even allows string matching](https://docs.sentry.io/platforms/apple/guides/ios/configuration/swizzling/) to disable specific instances.

Thanks for any input!

### Solution Brainstorm

_No response_

ioannisj:

Closes #362 

Docs PR: https://github.com/PostHog/posthog.com/pull/13204

- Adds a config option to disable swizzling
- Refactor PostHogSessionManager from singleton to SDK instance member in line with what [js lib](https://github.com/PostHog/posthog-js/blob/3fc39279cb0b179a91eb7f0b4e25ad6be9fcf948/packages/browser/src/posthog-core.ts#L529) does 

## :pencil: Checklist

### DIFF
--- a/PostHog/PostHogSessionManager.swift
+++ b/PostHog/PostHogSessionManager.swift
@@ -32,7 +32,9 @@ import Foundation
 
     func setup(config: PostHogConfig) {
         self.config = config
-        reset()
+        didBecomeActiveToken = nil
+        didEnterBackgroundToken = nil
+        applicationEventToken = nil
         registerNotifications()
         registerApplicationSendEvent()
     }


PostHog/PostHogConfig.swift
```diff
@@ -54,6 +54,15 @@ public typealias BeforeSendBlock = (PostHogEvent) -> PostHogEvent?
 
     @objc public var captureApplicationLifecycleEvents: Bool = true
     @objc public var captureScreenViews: Bool = true
+
+    /// Enable method swizzling for SDK functionality that depends on it
+    ///
+    /// When disabled, functionality that require swizzling (autocapture, screen views, session replay, surveys) will not be installed.
+    /// Note: The SDK may still use minimal internal swizzling for core functionality of managing sessions, but this will be kept to a minimum.
```
marandaneto: this should not happen, if its disabled, there should not be any swizzling

ioannisj: Yeah agreed! This is referring to this comment https://github.com/PostHog/posthog-ios/pull/388#issuecomment-3390007259

marandaneto: i'd just disable the session generation and/or rotation and mention that this feature wont work either
you can also tell people that they could do posthog.startSession posthog.endSession manually, maybe we should add a posthog.rotateSession as well so they have full control manually
about your comment: i didnt understand what you want to do, but my tldr is that you just dont do swizzling so that means just adding some checks and bailing out early for all swizzling

ioannisj: Okay I'll check how js lib handles session manager (and session ids) per instance. It think the cleanest approach here is `refactor so that each SDK instance runs its own session manager (and therefore inherits its config)`, so if a specigic SDK instance config has swizzling disabled, so will it's session manager

marandaneto: tbh i dont follow what is the issue with disabling swizzling and SDK instances and session manager

ioannisj: session manager is a singleton rn, which means it's shared between all the SDK instances. if one instance sets enableSwizzling: false and another later sets it to true, there’s no clean way to respect both. We could say that the first instance that configures with `enableSwizzling: false` turns off swizzling for the shared session manager but I think it would be better if config remains truly instance-scoped

PostHog/PostHogConfig.swift
```diff
@@ -54,6 +54,14 @@ public typealias BeforeSendBlock = (PostHogEvent) -> PostHogEvent?
 
     @objc public var captureApplicationLifecycleEvents: Bool = true
     @objc public var captureScreenViews: Bool = true
+
+    /// Enable method swizzling for SDK functionality that depends on it
+    ///
+    /// When disabled, functionality that require swizzling (like autocapture, screen views, session replay, surveys) will not be installed.
+    ///
+    /// Default: true
+    @objc public var enableSwizzling: Bool = true
```
marandaneto, line 66: we should mention about `$session_id` and multiple instances here

PostHog/PostHogSessionManager.swift
```diff
@@ -20,12 +20,10 @@ import Foundation
         case customSessionId = "Custom session set"
     }
 
-    @objc public static var shared: PostHogSessionManager {
-        DI.main.sessionManager
-    }
+    private let config: PostHogConfig
```
marandaneto: this will break the RN session plugin https://github.com/PostHog/posthog-react-native-session-replay/blob/927168c4dd4d62413ded54acc831720a7b8d72a2/ios/PosthogReactNativeSessionReplay.swift#L29

CHANGELOG.md
```diff
@@ -1,8 +1,12 @@
 ## Next
 
+- feat: add config option to disable swizzling ([#388](https://github.com/PostHog/posthog-ios/pull/388))
+- feat: SDK instance now manages its own session ([#388](https://github.com/PostHog/posthog-ios/pull/388))
+> Note: In case of multiple SDK instances, each instance will now report and manage its own $session_id
```
marandaneto: this is just food for thought:
an ios or macos app is just a single app/user (at least most of them), and if the app opens or closes its a single session (which imo makes sense to be a single session id).
there might be cases where multiple session ids make sense, but nothing comes to mind right now.
i don't know why people would use multiple instances (different project ids? DI? etc) and that would be important to decide what makes more sense.
this would also be a breaking change in case people use multiple instances and expect all events to be tied together.

PostHog/PostHogConfig.swift
ioannisj: This is not specific to this config option though? Added a note in the changelog instead

CHANGELOG.md
ioannisj: Yeah, had the same concerns as well. The main reason I decided for instead of against this change is that js lib handles multiple instances and sessions similarly (linked in PR body). This means that a shared session id across instances in the iOS sdk was never really an intentional feature but more of a side effect of using the singleton pattern. Aligning with JS makes the behavior more predictable and consistent across SDKs, even if it technically introduces a breaking change for those rare (I suppose) multi-instance setups?

PostHog/PostHogSessionManager.swift
ioannisj: Yeah and being public other implementations as well. I'll add back pointing to PostHogSDK.shared.sessionManager. Thanx for this

marandaneto: commented here because it was the first change in the PostHogSessionManager file

PostHog/PostHogConfig.swift
marandaneto: changing this config will affect the behaviour of $session_id so it makes sense to me

ioannisj: Oh yeah, it will limit the session rotation logic, but not how sessions work with multiple instances. I'll add a comment

CHANGELOG.md
marandaneto: probably an oversight when i wrote the sdk
android is the same though https://github.com/PostHog/posthog-android/blob/44dece00a72ba131912be841ba523b2ecbae86f8/posthog/src/main/java/com/posthog/PostHog.kt#L371
so we should fix it as well
the session_id is stored on the disk or in an equivalent location in some cases. How does this work for other SDKs with multiple instances? how dont they overwrite each other?
https://github.com/PostHog/posthog-js/blob/3fc39279cb0b179a91eb7f0b4e25ad6be9fcf948/packages/core/src/posthog-core.ts#L186

ioannisj: Not sure if I'm reading this right but this seems to depend on an instance function [setPersistedProperty](https://github.com/PostHog/posthog-js/blob/3fc39279cb0b179a91eb7f0b4e25ad6be9fcf948/packages/core/src/posthog-core-stateless.ts#L143) which in case of RN and Web for example uses the api key as the storage key? https://github.com/PostHog/posthog-js/blob/3fc39279cb0b179a91eb7f0b4e25ad6be9fcf948/packages/browser/src/consent.ts#L73

ioannisj: And node client uses a mem storage instead? https://github.com/PostHog/posthog-js/blob/3fc39279cb0b179a91eb7f0b4e25ad6be9fcf948/packages/node/src/client.ts#L163

marandaneto: yes https://github.com/PostHog/posthog-js/blob/3fc39279cb0b179a91eb7f0b4e25ad6be9fcf948/packages/react-native/src/posthog-rn.ts#L226
then you have to check `PostHogRNStorage` and `buildOptimisiticAsyncStorage`

marandaneto: does RN use an API key as part of the file path? do we allow multiple instances with the same api key?

ioannisj: > does RN use an API key as part of the file path? do we allow multiple instances with the same api key?

Doesn't look like it does, or could not locate it. Even if it doesn't though, this just means multiple SDK instances setup in that case is not functioning properly (since storage is used for other things as well not just session_id)

ioannisj: > android is the same though [PostHog/posthog-android@44dece0/posthog/src/main/java/com/posthog/PostHog.kt#L371](https://github.com/PostHog/posthog-android/blob/44dece00a72ba131912be841ba523b2ecbae86f8/posthog/src/main/java/com/posthog/PostHog.kt#L371)
so we should fix it as well

Would it be an easy change you think? I can take a look. Do we agree then that this is the direction we are going with session management + instances?

PostHog/PostHogSessionManager.swift
ioannisj: Made a slight refactor to avoid optional (and fallback default that doesn't make sense) here

CHANGELOG.md
marandaneto: I'd rather create an issue and document the why/how and do at some other point so we can start error tracking asap, same for RN.

ioannisj: https://github.com/PostHog/posthog-android/issues/306

PR #389, opened 2025-11-06
fix: feature flags for person and groups and reload flags

marandaneto:
port fixes from https://github.com/PostHog/posthog-android/pull/315

## :pencil: Checklist

### DIFF
--- a/PostHog/PostHogContext.swift
+++ b/PostHog/PostHogContext.swift
@@ -147,6 +147,41 @@ class PostHogContext {
         theSdkInfo
     }
 
+    private lazy var thePersonPropertiesContext: [String: Any] = {
+        let staticCtx = staticContext()
+        let sdkInfo = sdkInfo()
+        var personProperties: [String: Any] = [:]
+
+        // App information
+        if let appVersion = staticCtx["$app_version"] {
+            personProperties["$app_version"] = appVersion
+        }
+        if let appBuild = staticCtx["$app_build"] {
+            personProperties["$app_build"] = appBuild
+        }
+
+        if let appNamespace = staticCtx["$app_namespace"] {
+            personProperties["$app_namespace"] = appNamespace
+        }
+
+        // Operating system information
+        if let osName = staticCtx["$os_name"] {
+            personProperties["$os_name"] = osName
+        }
+        if let osVersion = staticCtx["$os_version"] {
+            personProperties["$os_version"] = osVersion
+        }
+
+        // Device information
+        if let deviceType = staticCtx["$device_type"] {
+            personProperties["$device_type"] = deviceType
+        }
+
+        personProperties.merge(sdkInfo) { _, new in new }
+
+        return personProperties
+    }()
+
     private func platform() -> String {
         var sysctlName = "hw.machine"
 
@@ -202,38 +237,7 @@ class PostHogContext {
     /// Returns person properties context by extracting relevant properties from static context.
     /// This centralizes the logic for determining which properties should be used as person properties.
     func personPropertiesContext() -> [String: Any] {
-        let staticCtx = staticContext()
-        let sdkInfo = sdkInfo()
-        var personProperties: [String: Any] = [:]
-
-        // App information
-        if let appVersion = staticCtx["$app_version"] {
-            personProperties["$app_version"] = appVersion
-        }
-        if let appBuild = staticCtx["$app_build"] {
-            personProperties["$app_build"] = appBuild
-        }
-
-        if let appNamespace = staticCtx["$app_namespace"] {
-            personProperties["$app_namespace"] = appNamespace
-        }
-
-        // Operating system information
-        if let osName = staticCtx["$os_name"] {
-            personProperties["$os_name"] = osName
-        }
-        if let osVersion = staticCtx["$os_version"] {
-            personProperties["$os_version"] = osVersion
-        }
-
-        // Device information
-        if let deviceType = staticCtx["$device_type"] {
-            personProperties["$device_type"] = deviceType
-        }
-
-        personProperties.merge(sdkInfo) { _, new in new }
-
-        return personProperties
+        thePersonPropertiesContext
     }
 
     private func registerNotifications() {


PostHog/PostHogContext.swift
```diff
@@ -225,23 +230,25 @@ class PostHogContext {
...
-        // Localization - read directly to avoid expensive dynamicContext call
-        if #available(iOS 16.0, macOS 13.0, tvOS 16.0, watchOS 9.0, *) {
-            if let languageCode = Locale.current.language.languageCode {
-                personProperties["$locale"] = languageCode.identifier
-            }
-        } else {
-            if let languageCode = Locale.current.languageCode {
-                personProperties["$locale"] = languageCode
-            }
-        }
```
marandaneto, line 244: https://github.com/PostHog/posthog-android/pull/315/files#r2495899139

PostHog/PostHogSDK.swift
```diff
@@ -1083,15 +1114,29 @@ let maxRetryDelay = 30.0
...
+    ///   - groupType: The group type to clear properties for
+    ///   - reloadFeatureFlags: Whether to automatically reload feature flags after setting properties
+    @objc(resetGroupPropertiesForFlagsWithGroupType:reloadFeatureFlags:)
+    public func resetGroupPropertiesForFlags(_ groupType: String, reloadFeatureFlags: Bool = true) {
+        internalResetGroupPropertiesForFlags(groupType: groupType, reloadFeatureFlags: reloadFeatureFlags)
     }
 
     /// Internal implementation for resetting group properties.
-    private func resetGroupPropertiesForFlags(groupType: String?) {
+    private func internalResetGroupPropertiesForFlags(groupType: String?, reloadFeatureFlags: Bool) {
```
marandaneto, line 1139: to avoid name clash with the new method

PostHog/PostHogSDK.swift
```diff
@@ -983,10 +983,28 @@ let maxRetryDelay = 30.0
     /// // Feature flags will now use only server-side properties
     /// let flagValue = PostHogSDK.shared.isFeatureEnabled("feature")
     /// ```
-    ///
-    /// - Note: This method does not automatically reload feature flags. Call `reloadFeatureFlags()`
```
marandaneto, line 987: people might be calling `resetPersonPropertiesForFlags` and then `reloadFeatureFlags`
2 options:
we default  `reloadFeatureFlags` to false to avoid calling `reloadFeatureFlags` twice
we let it as is since the `loadingFeatureFlagsLock` and `loadingFeatureFlags` will do the job and bail out immediatelly on the 2nd call
i chose the later, let me know if thats not ideal.

PostHog/PostHogContext.swift
ioannisj: I assume these were removed on Android as well?

PostHog/PostHogSDK.swift
ioannisj: Yeah the only problem I see is that the reloadFeatureFlags callback will never be called. 

Maybe not for this PR but we need a way to save the callback if there's another reload request in-flight and call it once the original request finishes

PostHog/PostHogContext.swift
marandaneto: they were not added since default flags props didnt exist, you can check the PR above

PostHog/PostHogSDK.swift
marandaneto: not sure i follow, which callback?

PostHog/PostHogContext.swift
```diff
@@ -203,6 +203,7 @@ class PostHogContext {
     /// This centralizes the logic for determining which properties should be used as person properties.
     func personPropertiesContext() -> [String: Any] {
         let staticCtx = staticContext()
+        let sdkInfo = sdkInfo()
```
ioannisj: Since this whole function returns only static info, should this be moved to a lazy var that's calculated only once?

PostHog/PostHogSDK.swift
marandaneto: this is similar to call `identify`, we just call `reloadFeatureFlags` and dont allow the user to pass a callback unless they want to relaod it manually passing a callback

marandaneto: if thats the case, they could call the method with `reloadFeatureFlags: false` and then manually `reloadFeatureFlags(...)`

ioannisj: > people might be calling resetPersonPropertiesForFlags and then reloadFeatureFlags

So if they call reloadFeatureFlags with a callback the code will bail out because there's another call to flags by resetPersonPropertiesForFlags right?

ioannisj: I wonder if we should add a bool parameters to reloadFeatureFlags then that will reset person or group parameters

ioannisj (approved):
Lg left a few comments from discussion

PostHog/PostHogSDK.swift
marandaneto: > So if they call reloadFeatureFlags with a callback the code will bail out because there's another call to flags by resetPersonPropertiesForFlags right?

correct, thats what i meant [here](https://github.com/PostHog/posthog-ios/pull/389/files#r2498698006)

the easy solution would be to just default `reloadFeatureFlags: false` for now if thats a concern

marandaneto: > I wonder if we should add a bool parameters to reloadFeatureFlags then that will reset person or group parameters

not sure i understand, mind clarifying?

ioannisj: What I mean is maybe we can give people a way to do:

```
PostHog.shared.reloadFeatureFlags(
  resetPersonProperties: true
) {
  // completion handler
}
```

Instead of them calling `resetPersonPropertiesForFlags` and then `reloadFeatureFlags`

marandaneto: i dont think thats a good API
there are so many reasons to reload the flags, in this case the `reloadFeatureFlags` method would have a bunch of things (reset groups, reset all, etc).
its a much more natural API to eg
`resetPersonPropertiesForFlags` and flags are automatically reloaded unless opt out
or `resetPersonPropertiesForFlags` never reloads and the user has to reload manually, it seems a much better UX IMO
its the intention and the side effect and not the other way around if that makes sense

marandaneto: i will merge to unblock the fix

ioannisj: Yeap agreed. I think a future improvement is saving the callbacks from reloadFeatureFlags calls in memory and calling them whenever flags reload (for whatever reason) so that we make sure that the logic that people attach there always gets executed

PR #392, opened 2025-11-20
fix: move the feature_flag_called event filter to a better location (after the events filter) 

wowrumal:
My change addresses the behavior of sticky feature flags from [this PostHog tutorial](https://posthog.com/tutorials/sticky-feature-flags). When initially reading and assigning feature flags to a user, I want to avoid sending the $feature_flag_called event, since these evaluations are not real exposures yet.
If I filter out these events using the beforeSend block, I lose all analytics related to feature flag exposures—including the ones I actually want to track. This approach also breaks experiment accuracy, because I can no longer trigger exposure events exactly when needed.
My update ensures that exposure events are sent only at the appropriate time, keeping experiment tracking reliable and analytics data clean.

Existing unit tests, manual testing in own app

## :pencil: Checklist

### DIFF
--- a/PostHog/PostHogSDK.swift
+++ b/PostHog/PostHogSDK.swift
@@ -756,6 +756,22 @@ let maxRetryDelay = 30.0
         return properties
     }
 
+    private func shouldDropFeatureFlagEvent(_ event: PostHogEvent?) -> Bool {
+        guard event?.event == "$feature_flag_called", let flagKey = event?.properties["$feature_flag"] as? String else {
+            return false
+        }
+
+        var shouldCapture = false
+        flagCallReportedLock.withLock {
+            if !flagCallReported.contains(flagKey) {
+                flagCallReported.insert(flagKey)
+                shouldCapture = true
+            }
+        }
+
+        return !shouldCapture
+    }
+
     @objc public func alias(_ alias: String) {
         if !isEnabled() {
             return
@@ -861,7 +877,12 @@ let maxRetryDelay = 30.0
             timestamp: timestamp
         )
 
-        let resultEvent = config.runBeforeSend(event)
+        var resultEvent = config.runBeforeSend(event)
+
+        // if feature $flag_called_event already sent with this key, don't send this event
+        if shouldDropFeatureFlagEvent(resultEvent) {
+            resultEvent = nil
+        }
 
         if resultEvent == nil {
             let originalMessage = "PostHog event \(eventName) was dropped"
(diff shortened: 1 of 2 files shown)


PostHog/PostHogSDK.swift
```diff
@@ -756,6 +756,22 @@ let maxRetryDelay = 30.0
         return properties
     }
 
+    private func shouldDropFeatureFlagEvent(_ event: PostHogEvent?) -> Bool {
+        guard event?.event == "$feature_flag_called", let flagKey = event?.properties["$feature_flag"] as? String else {
+            return false
+        }
+
+        var shouldCapture = false
```
ioannisj, line 764: A bit confusing now since we flipped the logic here, so better rename to `shouldDrop`

PostHog/PostHogSDK.swift
```diff
@@ -1207,39 +1228,28 @@ let maxRetryDelay = 30.0
...
-            if !flagCallReported.contains(flagKey) {
-                flagCallReported.insert(flagKey)
-                shouldCapture = true
-            }
-        }
-
-        if shouldCapture {
-            let requestId = remoteConfig?.lastRequestId ?? ""
-            let details = remoteConfig?.getFeatureFlagDetails(flagKey)
+    public func reportFeatureFlagCalled(flagKey: String, flagValue: Any?) {
```
ioannisj, line 1231: We changed the visibility here to public which I think is what you refer in changelog line `feat: allow user to manage feature_flag_called event`

Not sure I completely get the specific use case here but have you considered `config.sendFeatureFlagEvent = false` and manually capturing `$feature_flag_called`? 

It feels that If someone needs advanced or custom control over $feature_flag_called events, sendFeatureFlagEvent = false + manual capture() is probably more appropriate than expanding the public API. Also, if I'm not mistaken, the web sdk doesn't allow for this either and I think we shouldn't deviate much from what that SDK does

PostHog/PostHogSDK.swift
```diff
@@ -861,7 +877,12 @@ let maxRetryDelay = 30.0
             timestamp: timestamp
         )
 
-        let resultEvent = config.runBeforeSend(event)
+        var resultEvent = config.runBeforeSend(event)
+
+        // if feature $flag_called_event already sent with this key, don't send this event
+        if shouldDropFeatureFlagEvent(resultEvent) {
```
ioannisj, line 883: This changes behavior a bit cause now we will be dropping same-key flag events for people calling `capture("$feature_flag_called")` manually. Before we were running this check only for internal sdk calls. Not sure if more desirable or not but still a change

ioannisj (commented):
Thank you for this @wowrumal. Left some comments but I think @PostHog/team-feature-flags should also have a look

PostHog/PostHogSDK.swift
haritowa: You need internals to build the proper event. Reason, for example

haritowa:
Here it is: https://github.com/PostHog/posthog-ios/pull/393

Yet the PR you are currently looking at still have some point: our `beforeSend` closure can filter out flag called events while the rest SDK will treat it as a sent event(due to marking it as reported in `flagCallReported`)

While our newer PR fixes our specific problem, it is a good idea to still give a second look to the problem I outlined above

ioannisj:
hey guys, closing this one just to keep the web and mobile sdks aligned in terms of implementation. If this is still an issue for you we can reopen

PR #393, opened 2025-11-20
feat: add ability to toggle feature_flag_called event sending

wowrumal:
Provide the ability to disable and reenable automatically sent feature_flag_called event

We want to automatically report feature_flag_called event, but we need to disable this function when we read flags for sticky implementation according [to this guide](https://posthog.com/tutorials/sticky-feature-flags) to send and use this event as exposure event exactly when needed.

Existing unit tests. manual testing in own app

## :pencil: Checklist

### DIFF
--- a/PostHog/PostHogSDK.swift
+++ b/PostHog/PostHogSDK.swift
@@ -1190,6 +1190,16 @@ let maxRetryDelay = 30.0
         return value
     }
 
+    @objc public func setSendFeatureFlagEvent(_ enabled: Bool) {
+        guard isEnabled() else {
+            return
+        }
+
+        setupLock.withLock {
+            config.sendFeatureFlagEvent = enabled
+        }
+    }
+
     @objc public func isFeatureEnabled(_ key: String) -> Bool {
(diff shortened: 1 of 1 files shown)


PostHog/PostHogSDK.swift
```diff
@@ -1190,6 +1190,16 @@ let maxRetryDelay = 30.0
         return value
     }
 
+    @objc public func setSendFeatureFlagEvent(_ enabled: Bool) {
```
marandaneto, line 1193: i'd not add this as a global API, but rather do what i told here https://github.com/PostHog/posthog-android/pull/319/files#r2549035290

haritowa: Should we make it with a default value = current config value?

Also, how should it work when it differs from config value? Should it override config all the time?

The most interesting case is when auto exposure events are off via config, but we pass true as the parameter

wowrumal: @marandaneto I've just created a new PR with this approach
https://github.com/PostHog/posthog-ios/pull/396

PR #396, opened 2025-11-21
feat: add ability to disable sendFeatureFlagEvent for exact feature flag call

wowrumal:
Provide the ability to disable and reenable automatically sent feature_flag_called event

Existing unit tests. manual testing in own app

## :pencil: Checklist


PostHog/PostHogSDK.swift
```diff
@@ -1172,7 +1172,7 @@ let maxRetryDelay = 30.0
         }
     }
 
-    @objc public func getFeatureFlag(_ key: String) -> Any? {
+    @objc public func getFeatureFlag(_ key: String, sendEvent: Bool = true) -> Any? {
```
ioannisj: Afk right now to check what the other SDKs are doing here but I would make the param optional and default nil, so that it can take precedence when present over whatever config value. When nil we can fallback to configuration

ioannisj: Yeah I think the optional value suggestion above makes sense to me
```
public func getFeatureFlag(_ key: String, sendFeatureFlagEvent: Bool? = nil) -> Any?
```
When  `sendFeatureFlagEvent` is `true` -> always send
When `sendFeatureFlagEvent` is `false` -> never send
When `sendFeatureFlagEvent` is `nil` -> use config.sendFeatureFlagEvent

I'd also keep the name of this param and corresponding config the same for clarity

One thing to watch out for is that default parameters are not available in ObjC, so I'd keep the original function signature unchanged for source compatibility 

```swift
@objc public func getFeatureFlag(_ key: String) -> Any? { ... }
```

and add a new function for this with 
```swift
@objc(getFeatureFlagWithKey:sendFeatureFlagEvent:)
public func getFeatureFlag(_ key: String, sendFeatureFlagEvent: Bool? = nil) -> Any? { ... }
```

haritowa: Yeah this was my proposal too. I will slap @wowrumal 3 times for that bad decisions, we already created 3 PRs for the same thing. No way we make another mistake :D

wowrumal: I've updated PR according to the discussion, but we can't use Optional<Bool> in @objc method.
So I've created 2 overloads (w and w/o bool param) for these methods.

image

ioannisj: Looks good! Sorry George for the ping :) 
Waiting for this comment here https://github.com/PostHog/posthog-android/pull/319/files#r2558961646 though so that iOS and Android are aligned

PR #401, opened 2025-11-26
fix: avoid memory leaks on foat conversions

### LINKED ISSUE #399: Memory Leak during intensive auto UI updates
### Version

3.35.0

### Steps to Reproduce

Try to use memory leak monitoring tool from Xcode, also this is not problematic for app's which is not refreshing UI very fast, NSNumber takes very small memory, but this time I have a project where I update UI very fast and NSNumber allocation reach 4GB memory and more, it is increasing geometrical progression, it is very slow growing at the beginning, but at close to end it increasing very fast(200MB per second), end means applications stops working dues to very hight ram usage.

### Expected Result

My real device reach 100% CPU load and gradually increasing Ram usage which reaches limit after a while and this happening  after adding SDK, 

Here is code snippend which should be fixed

`/// A helper type for decoding JSON values, which may be nested objects, arrays, strings, numbers, booleans, or nulls.
private struct JSON: Decodable {
    let value: Any

    init(from decoder: Decoder) throws {
        let container = try decoder.singleValueContainer()

        if container.decodeNil() {
            value = NSNull()
        } else if let object = try? container.decode([String: JSON].self) {
            value = object.mapValues { $0.value }
        } else if let array = try? container.decode([JSON].self) {
            value = array.map(\.value)
        } else if let string = try? container.decode(String.self) {
            value = string
        } else if let bool = try? container.decode(Bool.self) {
            value = bool
        } else if let number = try? container.decode(Double.self) {
            value = NSNumber(value: number)
        } else if let number = try? container.decode(Int.self) {
            value = NSNumber(value: number)
        } else {
            throw DecodingError.dataCorruptedError(
                in: container, debugDescription: "Invalid JSON value"
            )
        }
    }
}`

This code part leads NSNumber allocation issue, please use normal Double and Int instead of NSNumber, This avoids ObjC bridging/boxing and should reduce the NSNumber allocations

### Actual Result

Image

ioannisj:

Closes #399 

## :pencil: Checklist

### DIFF
--- a/PostHog/Replay/Float+Util.swift
+++ b/PostHog/Replay/Float+Util.swift
@@ -7,14 +7,11 @@
 
 import Foundation
 
-extension CGFloat {
-    func toInt() -> Int {
-        NSNumber(value: rounded()).intValue
-    }
-}
-
-extension Double {
-    func toInt() -> Int {
-        NSNumber(value: rounded()).intValue
+extension BinaryFloatingPoint {
+    func toInt() -> Int? {
+        guard isFinite else { return nil }
+        guard self >= Self(Int.min), self <= Self(Int.max) else { return nil }
+        // Since this is used primarily for UI size & position, rounding vs truncating makes sense
+        return Int(rounded())
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Next
 
+- fix: avoid memory leaks on foat conversions ([#401](https://github.com/PostHog/posthog-ios/pull/401))
+
 ## 3.35.0 - 2025-11-07
 
 - fix: call the flags api with the correct groups key name (the api has a back compatible fix already) ([#389](https://github.com/PostHog/posthog-ios/pull/389))
--- a/PostHog/Replay/PostHogReplayIntegration.swift
+++ b/PostHog/Replay/PostHogReplayIntegration.swift
@@ -276,8 +276,8 @@
                         continue
                     }
 
-                    let posX = touch.location.x.toInt()
-                    let posY = touch.location.y.toInt()
+                    let posX = touch.location.x.toInt() ?? 0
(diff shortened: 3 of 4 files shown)


PostHog/Replay/Float+Util.swift
```diff
@@ -7,14 +7,11 @@
...
-
-extension Double {
-    func toInt() -> Int {
-        NSNumber(value: rounded()).intValue
+extension BinaryFloatingPoint {
+    func toInt() -> Int? {
+        guard isFinite else { return nil }
+        guard self >= Self(Int.min), self <= Self(Int.max) else { return nil }
+        // Since this is used primarily for UI size & position, rounding vs truncating makes sense
+        return Int(rounded())
```
marandaneto, line 15: wont that cause the very same issue? instead of NSNumber but now too many Int allocations?
can we actually just reuse the already allocated objects? eg `touch.location.x`

marandaneto (approved):

PostHog/Replay/Float+Util.swift
ioannisj: Being a value type there's no heap alloc, right? Just a numerical cast between two value types. The problem was NSNumber being a reference type and we were alloc and then dealloc right after. 
However, this was under the assumption that we needed Int values for replay payload (since that was the original implementation). If we can use touch.location.x with no checks directly, that would be ideal I think. Do you have any context why we were doing Int(touch.location.x) before?

marandaneto: the compiler is also clever about `NSNumber(value: rounded()).intValue` as long as it knows its a safe conversion and wont heap allocate either but its not guaranteed.
so the changes make sense regarding that.
i'm not sure if we can use doubles, i think i've done Ints because of rrweb but i dont remember for sure.
you can merge as is to get something out and check out if it helps and meanwhile check with the replay team/player code are floating compatible or not

PR #404, opened 2025-12-04
feat: add manual error capture

ioannisj:

#skip-changelog

## :pencil: Checklist


PostHog/Error Tracking/Models/PostHogBinaryImageInfo.swift
```diff
@@ -0,0 +1,51 @@
...
+    let address: UInt64
+
+    /// Size of the binary image in bytes
+    let size: UInt64
+
+    var toDictionary: [String: Any] {
+        var dict: [String: Any] = [
+            "type": "macho",
+            "code_file": name,
+            "image_addr": String(format: "0x%llx", address),
```
marandaneto: `"0x%llx"` lets move this to a const

PostHog/Error Tracking/Models/PostHogStackFrame.swift
```diff
@@ -0,0 +1,62 @@
...
+    let function: String?
+    
+    /// Address of the symbol/function
+    let symbolAddress: UInt64?
+    
+    var toDictionary: [String: Any] {
+        var dict: [String: Any] = [:]
+        
+        dict["instruction_addr"] = String(format: "0x%llx", instructionAddress)
+        dict["platform"] = "ios" // always the same for iOS SDK (may need to revisit)
```
marandaneto: algign this with cymbal changes

PostHog/ErrorTracking/PostHogErrorTrackingConfig.swift
```diff
@@ -0,0 +1,85 @@
...
+    ///
+    /// Frames matching these prefixes will be marked as not in-app,
+    /// unless they also match `inAppIncludes` (which takes precedence).
+    ///
+    /// Example:
+    /// ```swift
+    /// config.errorTrackingConfig.inAppExcludes = [
+    ///     "ThirdPartySDK",
+    ///     "AnalyticsLib"
+    /// ]
```
marandaneto, line 48: examples on inAppIncludes and inAppExcludes dont match, lets make sure to use examples that make sense in the iOS world

PostHog/Error Tracking/PostHogErrorTrackingConfig.swift
```diff
@@ -0,0 +1,85 @@
...
+    /// Default: true
+    @objc public var inAppByDefault: Bool = true
+
+    // MARK: - Initialization
+
+    override public init() {
+        super.init()
+
+        // Auto-add main bundle identifier
+        if let bundleId = Bundle.main.bundleIdentifier {
```
marandaneto: we have a function for this arready `getBundleIdentifier`

PostHog/Error Tracking/PostHogExceptionProcessor.swift
```diff
@@ -0,0 +1,371 @@
...
+    /// - Returns: Dictionary of properties in PostHog's $exception event format
+    static func errorToProperties(
+        _ error: Error,
+        handled: Bool,
+        mechanismType: String,
+        config: PostHogErrorTrackingConfig
+    ) -> [String: Any] {
+        var properties: [String: Any] = [:]
+
+        properties["$exception_level"] = "error" // TODO: figure this out from error wrapped type
```
marandaneto: its error if manual capture or fatal if from a crash handler

PostHog/Error Tracking/PostHogExceptionProcessor.swift
```diff
@@ -0,0 +1,371 @@
...
+    }
+
+    /// Convert a message string to properties
+    ///
+    /// - Parameters:
+    ///   - message: The error message to convert
+    ///   - type: Optional exception type name (defaults to "String")
+    ///   - config: Error tracking configuration for in-app detection
+    /// - Returns: Dictionary of properties in PostHog's $exception event format
+    static func messageToProperties(
```
marandaneto: i dont think we have a capture error with a string anywhere else, right? this is kinda logging

PostHog/Error Tracking/PostHogExceptionProcessor.swift
```diff
@@ -0,0 +1,371 @@
...
+        // Walk exception chain via NSUnderlyingErrorKey
+        nsExceptions.append(exception)
+
+        var current = exception
+        while let underlying = current.userInfo?[NSUnderlyingErrorKey] as? NSException {
+            nsExceptions.append(underlying)
+            current = underlying
+        }
+
+        // Build exceptions (same order as Android)
```
marandaneto: you have to respect the platform order here, you can get an exception with nested exceptions if any and check how xcode/console would print in the console, and use the same order, the posthog UI should have the same experience

PostHog/Error Tracking/PostHogExceptionProcessor.swift
```diff
@@ -0,0 +1,371 @@
...
+    private static func buildException(
+        from exception: NSException,
+        handled: Bool,
+        mechanismType: String,
+        config: PostHogErrorTrackingConfig
+    ) -> [String: Any]? {
+        var exceptionDict: [String: Any] = [:]
+
+        exceptionDict["type"] = exception.name.rawValue
+        exceptionDict["value"] = exception.reason ?? "Unknown exception"
```
marandaneto: i'd rather set nothing, thats what we do in other sdks i think

PostHog/Error Tracking/Utils/PostHogStackTrace.swift
```diff
@@ -0,0 +1,181 @@
...
+import Darwin
+import Foundation
+import MachO
+
+/// Utility for capturing and processing stack traces
+///
+/// This class provides methods to capture stack traces from the current thread
+/// and format them consistently for error tracking.
+///
+enum PostHogStackTrace {
```
marandaneto: should we call this provider/processor as well?

PostHog/PostHogSDK.swift
```diff
@@ -1438,6 +1438,115 @@ let maxRetryDelay = 30.0
...
+    public func captureException(
+        _ error: Error,
+        properties: [String: Any]? = nil
+    ) {
+        guard isEnabled() else { return }
+
+        let errorProperties = PostHogExceptionProcessor.errorToProperties(
+            error,
+            handled: true,
+            mechanismType: "generic-error",
```
marandaneto: i think the default is always `generic`, check other sdks

PostHog/PostHogSDK.swift
```diff
@@ -1438,6 +1438,115 @@ let maxRetryDelay = 30.0
...
+    /// if unexpectedCondition {
+    ///     PostHog.shared.captureException("Unexpected state detected")
+    /// }
+    /// ```
+    ///
+    /// - Parameters:
+    ///   - message: The error message to capture
+    ///   - properties: Optional additional properties to attach to the event
+    @objc(captureExceptionWithMessage:properties:)
+    public func captureException(
```
marandaneto: i dont think this makes sense for ios, i'd rather call captureMessage but i'd only add if other sdks do have as well

PostHog/ErrorTracking/PostHogExceptionProcessor.swift
```diff
@@ -0,0 +1,371 @@
...
+
+    /// Convert Error/NSError to properties
+    ///
+    /// - Parameters:
+    ///   - error: The error to convert
+    ///   - handled: Whether the error was caught/handled
+    ///   - mechanismType: The mechanism that captured the error (e.g., "generic", "NSException")
+    ///   - config: Error tracking configuration for in-app detection
+    /// - Returns: Dictionary of properties in PostHog's $exception event format
+    static func errorToProperties(
```
marandaneto, line 27: we need to make sure that those methods never throw, and swallow if needed
so make sure that everything bails out if its nil, out of range, etc, and do a try catch with swift as well
the main goal here is that if somebody calls captureException(x), the app will never crash, or we did everything posible to prevent that

PostHog/ErrorTracking/PostHogErrorTrackingConfig.swift
ioannisj: I think I added some new examples

PostHog/Error Tracking/PostHogExceptionProcessor.swift
ioannisj: yeah left a comment there since for crash handler we'll probably wrap in a PostHogError (same pattern as Android and Flutter). For now since it's only manual capture then we default to "error". Unless I'm misreading your comment

PostHog/Error Tracking/PostHogExceptionProcessor.swift
ioannisj: This was a response to the comment [here](https://github.com/PostHog/posthog-android/pull/300#discussion_r2422780375) and the fact that Sentry also has a similar API. No strong opinions here, we could remove

PostHog/PostHogSDK.swift
ioannisj: I think this is the same discussion here https://github.com/PostHog/posthog-ios/pull/404#discussion_r2610627751

PostHog/PostHogSDK.swift
ioannisj: My understanding is that mechanism is just meta that makes it easier to identify _how_ the exception was captured (and specific to [platform implementation](https://github.com/PostHog/posthog-js/blob/90450a3724b61aab1a880628fd3f5c2f689fa1fb/packages/core/src/error-tracking/types.ts#L42)). I agree though that maybe we are being too specific on which manual error capture api was called. I changed to “generic” for all

PostHog/Error Tracking/PostHogExceptionProcessor.swift
ioannisj: Updated the comment there to clear this up. Yeah, Xcode will print root cause first, then nest underlying errors

PostHog/Error Tracking/Utils/PostHogStackTrace.swift
ioannisj: Changed

PostHog/ErrorTracking/PostHogExceptionProcessor.swift
ioannisj: Unfortunately do-catch will only work with explicit throws and not runtime crashes. I went through and made sure that nothing obvious can crash (and added some safety checks). To the best of my knowledge, this is safe code but I can't know for sure. I'll stress test though with some for-loops and async queues before releasing

PostHog/Error Tracking/Models/PostHogBinaryImageInfo.swift
```diff
@@ -0,0 +1,51 @@
...
+            "code_file": name,
+            "image_addr": String(format: PostHogStackFrame.hexAddressFormat, address),
+            "image_size": size,
+        ]
+
+        if let uuid = uuid {
+            dict["debug_id"] = uuid
+        }
+
+        if vmAddress > 0 {
```
marandaneto: is 0 invalid? should we filter only negative?

PostHog/Error Tracking/PostHogExceptionProcessor.swift
marandaneto: i'd rather do this with breadcrumbs, or when our UI has a clear distinction between whats a thrown string error or just a capture message (which is kinda a loging) of sorts

PostHog/ErrorTracking/Models/PostHogBinaryImageInfo.swift
```diff
@@ -0,0 +1,51 @@
+//
```
marandaneto, line 1: lets rename the package from `Error Tracking` without an extra space, i've seen issues with that

PostHog/PostHogSDK.swift
marandaneto: I'd remove this for now

PostHogExample/AppDelegate.swift
```diff
@@ -12,18 +12,20 @@ import UIKit
...
+            apiKey: "phc_DOkauJvMj0YFtJsPHqzH6BgpFm79CvU9DPE5E22yRMk",
+            host: "http://localhost:8010"
         )
         // the ScreenViews for SwiftUI does not work, the names are not useful
         config.captureScreenViews = false
         config.captureApplicationLifecycleEvents = false
 //        config.flushAt = 1
 //        config.flushIntervalSeconds = 30
         config.debug = true
+        config.flushAt = 1
```
marandaneto: just uncomment L21 instead

PostHogExample/AppDelegate.swift
```diff
@@ -12,18 +12,20 @@ import UIKit
 class AppDelegate: NSObject, UIApplicationDelegate {
     func application(_: UIApplication, didFinishLaunchingWithOptions _: [UIApplication.LaunchOptionsKey: Any]? = nil) -> Bool {
         let config = PostHogConfig(
-            apiKey: "phc_WKfvDfedaJEDCoUmt9pVa3OWtbbUP1W2ctxwXkt3A3n"
+            apiKey: "phc_DOkauJvMj0YFtJsPHqzH6BgpFm79CvU9DPE5E22yRMk",
+            host: "http://localhost:8010"
```
marandaneto: lets keep the sample as is so its a play and run to test things without having to run posthog or comment code to start with

PostHogExample/AppDelegate.swift
```diff
@@ -12,18 +12,20 @@ import UIKit
...
         config.captureApplicationLifecycleEvents = false
 //        config.flushAt = 1
 //        config.flushIntervalSeconds = 30
         config.debug = true
+        config.flushAt = 1
         config.sendFeatureFlagEvent = false
 
         #if os(iOS)
-            config.sessionReplay = true
+            config.sessionReplay = false
```
marandaneto: same as https://github.com/PostHog/posthog-ios/pull/404/files#r2630728120
all features enabled by default - its easier to catch bugs

marandaneto (approved):
left a few comments otherwis LGTM

PostHogExample/AppDelegate.swift
ioannisj: Ah this was not intentional, reverted

PostHog/Error Tracking/PostHogExceptionProcessor.swift
ioannisj: Removed

PostHog/Error Tracking/Models/PostHogBinaryImageInfo.swift
ioannisj: yeah, removed the check

PostHog/ErrorTracking/Models/PostHogBinaryImageInfo.swift
ioannisj: Done. Renamed a couple of other modules for consistency as well

PR #405, opened 2025-12-05
fix: only report  if the flag value has changed

marandaneto:
https://posthog.slack.com/archives/C03P7NL6RMW/p1764861563168719
Follow up https://github.com/PostHog/posthog-ios/pull/232
Matches android impl

## :pencil: Checklist

### DIFF
--- a/PostHogTests/PostHogSDKTest.swift
+++ b/PostHogTests/PostHogSDKTest.swift
@@ -686,6 +686,34 @@ class PostHogSDKTest: QuickSpec {
             expect(event[1].event).to(equal("force_batch_flush"))
         }
 
+        it("captures $feature_flag_called again when getFeatureFlag returns different value after reloading flags") {
+            let sut = self.getSut(
+                sendFeatureFlagEvent: true,
+                flushAt: 3
+            )
+
+            // First call gets a false value
+            _ = sut.getFeatureFlag("disabled-flag")
+
+            // Change the mock server to return a different value for the same key
+            server.disabledFlag = true
+
+            sut.reloadFeatureFlags {
+                // Second call gets a true value
+                _ = sut.getFeatureFlag("disabled-flag")
+                sut.capture("force_batch_flush")
+            }
+
+            waitFlagsRequest(server)
+
+            let events = getBatchedEvents(server)
+            expect(events.count).to(equal(3))
+
+            expect(events[0].event).to(equal("$feature_flag_called"))
+            expect(events[1].event).to(equal("$feature_flag_called"))
+            expect(events[2].event).to(equal("force_batch_flush"))
+        }
+
         describe("beforeSend hook") {
             let eventTriggers = getBeforeSendEventsConfig()
             let testOtherEventKey = "other_event"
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 ## Next
 
-- chore: bump actions/checkout from 5 to 6 ([#397](https://github.com/PostHog/posthog-ios/pull/397))
```
marandaneto, line 3: we only keep user facing changes in the changelog

PostHog/PostHogSDK.swift
```diff
@@ -41,7 +41,7 @@ let maxRetryDelay = 30.0
     #if !os(watchOS)
         private var reachability: Reachability?
     #endif
-    private var flagCallReported = Set<String>()
+    private var flagCallReported: [String: [Any?]] = .init()
```
ioannisj, line 44: Shouldn't we be storing the last known value here instead of [Any?]

ioannisj: Also this is unbounded so in theory can grow indefinitely. 

Another alternative is to stick to Set<FeatureFlagCalledKey> similar to [Android](https://github.com/PostHog/posthog-android/blob/ea03a27a80e3d25c19a40e84abb35340b386892a/posthog/src/main/java/com/posthog/internal/PostHogFeatureFlagCalledCache.kt#L126) where FeatureFlagCalledKey: Hashable, even though Any? could be a problem in this case

PostHog/PostHogSDK.swift
```diff
@@ -1207,13 +1204,35 @@ let maxRetryDelay = 30.0
...
+        case (nil, nil):
+            return true
+        case let (boolValue as Bool, flagBoolValue as Bool):
+            return boolValue == flagBoolValue
+        case let (stringValue as String, flagStringValue as String):
+            return stringValue == flagStringValue
+        default:
+            return false
+        }
+    }
```
ioannisj, line 1218: This comparison here is incomplete? Are we missing other types like Int, Double, [] etc?

PostHog/PostHogSDK.swift
marandaneto: reset or close will clear the list, so its only gonna grow up to the number of flags, theres no other solution for not capturing repeated events

marandaneto: cannot use `Set` with Any, yes
the current approach is the same as android https://github.com/PostHog/posthog-android/blob/ea03a27a80e3d25c19a40e84abb35340b386892a/posthog/src/main/java/com/posthog/PostHog.kt#L60

PostHog/PostHogSDK.swift
marandaneto: no, getFeatureFlag only returns booleans, strings or null.
getFeatureFlagPayload returns all types, but we dont capture events for reading payloads

PostHog/PostHogSDK.swift
ioannisj: Ah got you, I assumed we were using the lru cache I linked above! All good

ioannisj (approved):

PR #409, opened 2025-12-12
fix: iOS26 session replay masking

### LINKED ISSUE #408: Session replay masking seems to be broken on iOS 26+
### Description

Some recordings seem to ignore the global masking config on iOS 26. 

Ticket: https://posthoghelp.zendesk.com/agent/tickets/44565

ioannisj:

Closes #408 

Docs PR: https://github.com/PostHog/posthog.com/pull/14088

Updates the session replay masking logic to handle SwiftUI behavior changes introduced with the iOS 26 + Xcode 26 and the new SwiftUI rendering engine, where some SwiftUI primitives are rendered as `CALayer` sublayers rather than predictable backing `UIViews`

I could not find a better alternative other than iterating over each view's sublayers to find SwiftUI maskable views. This is limited to iOS 26+ but could potentially have a performance hit for complex hierarchies. 

Note: Because SwiftUI primitive views are now host-less, our approach with `.postHogMask()` and `.postHogNoMask()` is broken as well. I dropped some comments in the code but I'll follow up with a docs update to note this + a fresh issue to track

Manually with sample projects

## :pencil: Checklist

### DIFF
--- a/PostHog/Replay/PostHogReplayIntegration.swift
+++ b/PostHog/Replay/PostHogReplayIntegration.swift
@@ -535,19 +535,7 @@
             // I could not find a consistent pattern to limit this layer search approach,
             // so for iOS 26 we'll need to iterate over the view's sublayers as well to find maskable elements.
             if #available(iOS 26.0, *) {
-                for layer in view.layer.sublayers ?? [] {
-                    if swiftUITextBasedViewTypes.contains(where: layer.isKind(of:)) {
-                        if isTextInputSensitive(view) {
-                            maskableWidgets.append(layer.toAbsoluteRect(window))
-                        }
-                    }
-
-                    if swiftUIImageLayerTypes.contains(where: layer.isKind(of:)) {
-                        if isSwiftUIImageSensitive(view) {
-                            maskableWidgets.append(layer.toAbsoluteRect(window))
-                        }
-                    }
-                }
+                findMaskableLayers(view.layer, view, window, &maskableWidgets)
             }
 
             // this can be anything, so better to be conservative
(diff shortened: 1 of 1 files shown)


PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -512,6 +525,31 @@
...
+            //
+            // Note: prior to iOS 26, primitive views like Text, Image, Button were drawn
+            // inside a host view (subclass of UIView), however starting iOS 26 this seems
+            // to no longer be the case and all SwiftUI primitives seem to be somehow drawn
+            // without an underlying UIView host view (but as sublayers of other parent views).
+            //
+            // I could not find a consistent pattern to limit this layer search approach,
+            // so for iOS 26 we'll need to iterate over the view's sublayers as well to find maskable elements.
+            if #available(iOS 26.0, *) {
+                for layer in view.layer.sublayers ?? [] {
```
pauldambra: am assuming we don't need to recurse in and find sublayers of `layer`

ioannisj: Empirically, this seems to catch all the SwiftUI primitive views. They appear to be added only as direct sublayers to views, but there's no official documentation on how this works internally. Testing on some sample apps, this seems to do the trick. I'd prefer to roll out with this single level scan and see how it works in the wild? Hope we won't have to scan deeper than this

ioannisj: @pauldambra I was wrong, your comment made me look twice. Checked with a more complex UI hierarchy and indeed we do need to recurse here. thanx

PR #415, opened 2025-12-19
fix: SwiftUI masking view modifiers on iOS 26

### LINKED ISSUE #410: SwiftUI .postHogMask() and .postHogNoMask() don't work on SwiftUI primitive views on iOS 26
### Description

Context: https://github.com/PostHog/posthog-ios/pull/409

On iOS 26 some SwiftUI primitive views are no longer mapped reliably to a backing `UIView`. Our approach with `.postHogMask()` and `.postHogNoMask()` depends on finding a SwiftUI backing view and injecting UIKit views so that we can later mask during session replay capture. This no longer works with Xcode 26 for iOS 26 

Primitive views affected are `Text` `Image/AsyncImage` and `Button`

ioannisj:

Closes #410 

Docs: https://github.com/PostHog/posthog.com/pull/14178

Took a while longer that I hoped but finally figured this one out. Lot's of trial and error.  

A first iteration of this fix was changing too much the view masking logic inside `PostHogReplayIntegration.swift` that I was comfortable with, so I decided to take a different approach. 

Now most of the meat of the changes are inside the `PostHogTagViewModifier.swift`. For iOS 26 we're recursively checking for new sublayers that SwiftUI is using and marking them as either `. postHogNoMask` or `.postHogNoCapture`. Then it's a simple check in `findMaskableWidgets` to handle masking. This way, these changes are a bit contained on the usage of these modifiers and not the general masking logic. 

Can't be 100% confident we do catch all cases here but I did try and test this on a number of sample projects and compared the output of iOS 18.0 and iOS 26

### With .postHogMask()

<table>
<tr>
 <td> iOS 18
 <td> iOS 26
<tr>
 <td> 
ios18-1

 <td> ios26-1
<tr>
<td>
ios18-2
<td>
ios26-2
</table>

### With .postHogNoMask()

<table>
<tr>
 <td> iOS 18
 <td> iOS 26
<tr>
 <td> 
ios18-nomask-1

 <td> 
ios26-nomask-1

</table>

Sample projects

## :pencil: Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,6 +7,9 @@ on:
 jobs:
   cancel-previous-workflow:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      actions: write
     steps:
       - name: Cancel Previous Runs
         uses: styfle/cancel-workflow-action@85880fa0301c86cca9da44039ee3bb12d3bedbfa # pin@0.12.1


PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -603,6 +619,51 @@
             }
         }
 
+        /// Debug helper to print the view hierarchy with indentation
```
marandaneto: I'd remove or comment out those lines to avoid bundle size increases

marandaneto (approved):
other than the debugging methods, LGTM

PostHog/Replay/PostHogReplayIntegration.swift
ioannisj: Yeah agreed. Thought it would be useful in the future but I can always save a snippet. Will remove

PR #417, opened 2025-12-22
feat: add crash reporting

ioannisj:

Stacked on top of https://github.com/PostHog/posthog-ios/pull/404
#skip-changelog

Adds automatic crash reporting to the PostHog iOS SDK using `PLCrashReporter`. 
Crashes are captured and sent as `$exception` events with level "fatal" on the **next** app launch.

### Crash Types Handled: 
- Uncaught NSExceptions
- POSIX signals (e.g., SIGSEGV, SIGABRT, SIGBUS)
- Mach exceptions (e.g., EXC_BAD_ACCESS, EXC_CRASH)

### Implementation notes
- Context Preservation: Event context (distinct_id, groups, session_id, properties) is serialized to PLCrashReporter.customData on every context change. PLCrashReporter persists this on disk. We then read this back when processing a crash so we can reconstruct event properties at the time of crash (and not at the time of processing)
- Crash Report Processing: On next launch, pending crash reports are processed before enabling the crash reporter for new crashes. For now, this is a best effort, in case of failure to process we simply purge the crash report. As a future improvement we could implement a retry mechanism and a check for corrupted reports
- Debug Images: Only binary images referenced in the stack trace are included
- Debugger Detection: Crash handler is automatically disabled when a debugger is attached
- Platform Support: iOS, macOS, tvOS only (watchOS has a stub)

### Limitations
- Swift crash messages: Swift crashes (like fatalError,force unwraps) appear as SIGTRAP without the actual error message (PLCrashReporter doesn't expose __crash_info section). Sentry and Bugsnag seem to have figured this out already. Left this out for now since it will require modifying PLCrashReporter 
- NSException chaining: Underlying errors via NSUnderlyingErrorKey are not captured (PLCrashReporter only exposes name/reason)

- Manual testing. Will need symbolication before we know if the stack traces are healthy
- Added some unit tests on utils and processing

## :pencil: Checklist

### DIFF
--- a/PostHog.xcodeproj/project.pbxproj
+++ b/PostHog.xcodeproj/project.pbxproj
@@ -2150,6 +2150,27 @@
 		};
 /* End PBXResourcesBuildPhase section */
 
+/* Begin PBXShellScriptBuildPhase section */
+		DAAA28DD2F43037300BC69DB /* Upload symbols */ = {
+			isa = PBXShellScriptBuildPhase;
+			buildActionMask = 2147483647;
+			files = (
+			);
+			inputFileListPaths = (
+			);
+			inputPaths = (
+			);
+			name = "Upload symbols";
+			outputFileListPaths = (
+			);
+			outputPaths = (
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+			shellPath = /bin/sh;
+			shellScript = "#${BUILD_DIR%/Build/*}/SourcePackages/checkouts/posthog-ios/build-tools/upload-symbols\n\"${SRCROOT}/build-tools/upload-symbols\"\n";
+		};
+/* End PBXShellScriptBuildPhase section */
+
 /* Begin PBXSourcesBuildPhase section */
 		3AA34CF3296D951A003398F4 /* Sources */ = {
 			isa = PBXSourcesBuildPhase;
--- a/build-tools/upload-symbols
+++ b/build-tools/upload-symbols
@@ -1,0 +1,86 @@
+#!/bin/bash
+#
+# PostHog Debug Symbols Upload Script
+# https://posthog.com/docs/error-tracking/upload-source-maps/ios
+#
+# Xcode Build Phase Setup:
+#   SPM:        "${BUILD_DIR%/Build/*}/SourcePackages/checkouts/posthog-ios/build-tools/upload-symbols"
+#   CocoaPods:  "${PODS_ROOT}/PostHog/build-tools/upload-symbols"
+#
+# Input Files (required for script sandboxing):
+#   ${DWARF_DSYM_FOLDER_PATH}/${DWARF_DSYM_FILE_NAME}
+#   ${DWARF_DSYM_FOLDER_PATH}/${DWARF_DSYM_FILE_NAME}/Contents/Resources/DWARF/${PRODUCT_NAME}
+#   ${DWARF_DSYM_FOLDER_PATH}/${DWARF_DSYM_FILE_NAME}/Contents/Info.plist
+#   $(TARGET_BUILD_DIR)/$(EXECUTABLE_PATH)
+#
+# Build Settings:
+#   DEBUG_INFORMATION_FORMAT = DWARF with dSYM File (script will skip DWARF-only builds)
+#
+# Environment Variables (optional):
+#   POSTHOG_CLI_INSTALL_DIR - Custom directory containing posthog-cli binary
+#
+
+# Validate environment
+if [ -z "${DWARF_DSYM_FOLDER_PATH}" ]; then
+    echo "warning: DWARF_DSYM_FOLDER_PATH not set"
+    exit 0
+fi
+
+if [ ! -d "${DWARF_DSYM_FOLDER_PATH}" ]; then
+    echo "warning: dSYM folder not found: ${DWARF_DSYM_FOLDER_PATH}"
+    exit 0
+fi
+
+# Check if folder contains any dSYM bundles
+if [ -z "$(find "${DWARF_DSYM_FOLDER_PATH}" -name '*.dSYM' -type d 2>/dev/null)" ]; then
+    echo "info: No dSYM bundles found in ${DWARF_DSYM_FOLDER_PATH}"
+    exit 0
+fi
+
+# Find posthog-cli (Xcode doesn't load shell profiles)
+# Priority: env var override > well-known location > npm global > PATH fallback
+if [ -f "$HOME/.posthog/posthog-cli" ]; then
+  PH_CLI_PATH="$HOME/.posthog/posthog-cli"
+else
+  # Check if installed via npm -g @posthog/cli
+  NPM_GLOBAL_PREFIX=$(npm prefix -g 2>/dev/null)
+  if [ -n "$NPM_GLOBAL_PREFIX" ] && [ -f "$NPM_GLOBAL_PREFIX/bin/posthog-cli" ]; then
+    PH_CLI_PATH="$NPM_GLOBAL_PREFIX/bin/posthog-cli"
+  else
+    # Check if installed as local dependency
+    NPM_LOCAL_ROOT=$(npm root 2>/dev/null)
+    if [ -n "$NPM_LOCAL_ROOT" ] && [ -f "$NPM_LOCAL_ROOT/.bin/posthog-cli" ]; then
+      PH_CLI_PATH="$NPM_LOCAL_ROOT/.bin/posthog-cli"
+    else
+      # Fallback to searching common locations
+      export PATH="/usr/local/bin:/opt/homebrew/bin:$HOME/.cargo/bin:$HOME/.local/bin:$HOME/.posthog:$PATH"
+      PH_CLI_PATH=$(command -v posthog-cli 2>/dev/null)
+    fi
+  fi
+fi
+
+if [ -z "$PH_CLI_PATH" ] || [ ! -x "$PH_CLI_PATH" ]; then
+    echo "error: posthog-cli not found, install with: npm install -g @posthog/cli"
+    exit 1
+fi
+
+# Build CLI arguments
+CLI_ARGS="--directory $DWARF_DSYM_FOLDER_PATH"
+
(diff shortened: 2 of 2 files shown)


PostHog.podspec
```diff
@@ -25,6 +25,12 @@ Pod::Spec.new do |s|
 
   s.frameworks = 'Foundation'
 
+  # PLCrashReporter dependency (not available on watchOS)
+  # Using ~> 1.8 for minimum compatibility with host apps
```
marandaneto, line 29: any reason to choose this version specifically?

PostHog/ErrorTracking/PostHogCrashReportProcessor.swift
```diff
@@ -0,0 +1,401 @@
...
+
+            // Build debug images for symbolication (only images referenced in stack trace)
+            let debugImages = buildDebugImages(from: report, stackFrames: stackFrames)
+            if !debugImages.isEmpty {
+                properties["$debug_images"] = debugImages
+            }
+
+            // Add crash metadata
+            if let uuidRef = report.uuidRef {
+                properties["$crash_report_id"] = CFUUIDCreateString(nil, uuidRef) as String
```
marandaneto, line 48: what should we do with this info? should we add this to taxonomy?

PostHog/ErrorTracking/PostHogCrashReportProcessor.swift
```diff
@@ -0,0 +1,401 @@
...
+
+                exception["mechanism"] = [
+                    "type": "mach_exception",
+                    "handled": false,
+                    "synthetic": false,
+                    "meta": [
+                        "mach": [
+                            "exception": machException.type,
+                            "code": machException.codes.first,
+                            "subcode": machException.codes.count > 1 ? machException.codes[1] : nil,
```
marandaneto, line 116: lets skip nil values

PostHog/Error Tracking/PostHogCrashReportProcessor.swift
```diff
@@ -0,0 +1,401 @@
...
+                debugImages.append(binaryImage)
+            }
+
+            return debugImages.map(\.toDictionary)
+        }
+
+        // MARK: - Helpers
+
+        /// Format string for zero-padded 64-bit hex addresses (e.g., "0x00007fff12345678")
+        static let hexAddressPaddedFormat = "0x%016llx"
```
marandaneto: make it private

PostHog/Error Tracking/PostHogCrashReportProcessor.swift
```diff
@@ -0,0 +1,401 @@
...
+            8: "EXC_MACH_SYSCALL",
+            9: "EXC_RPC_ALERT",
+            10: "EXC_CRASH",
+            11: "EXC_RESOURCE",
+            12: "EXC_GUARD",
+            13: "EXC_CORPSE_NOTIFY",
+        ]
+
+        private static func machExceptionName(_ type: UInt64) -> String {
+            machExceptionNames[type] ?? "EXC_UNKNOWN(\(type))"
```
marandaneto: probably better with EXC_UNKNOWN_X instead of EXC_UNKNOWN(X), more readable

PostHog/ErrorTracking/PostHogErrorTrackingAutoCaptureIntegration.swift
```diff
@@ -0,0 +1,179 @@
+//
+//  PostHogCrashReporterIntegration.swift
+//  PostHog
+//
+//  Created by Ioannis Josephides on 14/12/2025.
+//
+
+import Foundation
+
+#if os(iOS) || os(macOS) || os(tvOS)
```
marandaneto, line 10: is the manual capture also available for the 3 OSs?

PostHog/ErrorTracking/PostHogErrorTrackingAutoCaptureIntegration.swift
```diff
@@ -0,0 +1,179 @@
...
+
+        var requiresSwizzling: Bool { false }
+
+        private weak var postHog: PostHogSDK?
+        private var crashReporter: PLCrashReporter?
+
+        func install(_ postHog: PostHogSDK) throws {
+            try PostHogCrashReporterIntegration.integrationInstalledLock.withLock {
+                if PostHogCrashReporterIntegration.integrationInstalled {
+                    throw InternalPostHogError(description: "Crash report integration already installed to another PostHogSDK instance.")
```
marandaneto, line 25: we can just log and bail out isntead of throwing, this would be more expensive at runtime

PostHog/Error Tracking/PostHogErrorTrackingConfig.swift
```diff
@@ -12,6 +12,24 @@ import Foundation
...
+    /// - Uncaught `NSException`s
+    ///
+    /// Crashes are persisted to disk and sent as `$exception` events with level "fatal" **on the next app launch**
+    ///
+    /// - Note: Crash reporting is not available on watchOS (this will be a no-op).
+    /// - Note: Crash reporting is automatically disabled when a debugger is attached,
+    ///   as the debugger intercepts signals before the crash handler can process them.
+    ///
+    /// Default: false
+    @objc public var enableCrashHandler: Bool = false
```
marandaneto: i think we call this just autocapture on flutter? i think we should unify naming if that makes sense

PostHog/PostHogSDK.swift
```diff
@@ -620,6 +641,33 @@ let maxRetryDelay = 30.0
...
+            skipBuildProperties: false
+        )
+    }
+
+    /// Internal capture method that handles all event capture logic.
+    ///
+    /// - Parameters:
+    ///   - skipBuildProperties: When true, skips buildProperties call and uses properties as-is.
+    ///     Used by crash reporting to capture events with pre-built crash-time context.
+    func captureInternal(
```
marandaneto, line 791: this seems to be a public API now, how can we make this internal only for us? we should not expose this to customers

PostHog.podspec
ioannisj: 1.8.0 is the earliest version that [supports saving custom data](https://github.com/microsoft/plcrashreporter/blob/master/CHANGELOG.md#version-180) along with the crash report (PLCrashReporter.customData), which we use to persist context at crash time so we can accurately rebuild all the event properties on subsequent app launches. So minimum version that has the features we need

PostHog/PostHogSDK.swift
ioannisj: This is an internal method that accepts an additional `shouldBuildProperties: Bool` parameter (used internally when capturing an exception event). I think the public API remains unchanged

PostHog/Error Tracking/PostHogErrorTrackingConfig.swift
ioannisj: flutter uses `captureX` for different error types, but Android and RN do use `autoCapture`. 
Updated config to `autoCapture` and integration to `PostHogErrorTrackingAutoCaptureIntegration` for consistency

PostHog/ErrorTracking/PostHogCrashReportProcessor.swift
ioannisj: This is a uuid that PLCrashReport generates. It's crash metadata, so I figured we wouldn't want to lose it and it won't hurt including it, even though tbh I can't think of any uses of it. Maybe deduplication of reports BE side but that's highly unlikely since we delete the crash report once processed. I'm okay removing this

PostHog/ErrorTracking/PostHogCrashReportProcessor.swift
ioannisj: nil values are removed with `.compactMapValues { $0 }` below, for both "meta" and "mach" keys. Not super efficient but the number of keys is really small here so negligible I believe

PostHog/Error Tracking/PostHogCrashReportProcessor.swift
ioannisj: done

PostHog/Error Tracking/PostHogCrashReportProcessor.swift
ioannisj: agreed, done

PostHog/ErrorTracking/PostHogErrorTrackingAutoCaptureIntegration.swift
ioannisj: No, manual capture is available on watchOS as well. 
For this class, we have fallback stub  implementation for watchOS, which I think works well. To be consistent on the public API, however, I'll mark this config unavailable for watchOS

PostHog/ErrorTracking/PostHogErrorTrackingAutoCaptureIntegration.swift
ioannisj: This is a pattern we use for other integrations as well. We catch on install step and log in PostHogSDK. We could change the signature of this method to return a truthy for success or failure instead, but I'll skip for this PR to keep the scope down

PostHog/ErrorTracking/PostHogCrashReportProcessor.swift
marandaneto: i'd either leave without it or use this as the event uuid

PostHog/ErrorTracking/PostHogErrorTrackingAutoCaptureIntegration.swift
marandaneto: mm not sure, unless only on iOS, if the integration is installed already, just return only, i dont see a reason to throw, its just doing unnecessary cpu cycles.
for a next PR, I'd just return instead of throwing

PostHog/PostHogSDK.swift
marandaneto: nvm, its not public, my bad

marandaneto:
> Build is green locally but fails in CI, looking into it

Whats missing here for this to be merged?

PR #418, opened 2025-12-23
chore: codeql advanced

ioannisj:

Based on https://github.com/PostHog/posthog/pull/43050 and https://github.com/PostHog/posthog-android/pull/356
#skip-changelog

## :pencil: Checklist

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -21,8 +21,6 @@ jobs:
       - uses: maxim-lobanov/setup-xcode@60606e260d2fc5762a71e64e74b2174e8ea3c8bd # v1.6.0
         with:
           xcode-version: latest-stable
-      - name: Download Platforms
-        run: xcrun xcodebuild -downloadAllPlatforms
       - name: Initialize CodeQL
         uses: github/codeql-action/init@v4
         with:


PostHogExampleVisionOS/PostHogExampleVisionOS.xcodeproj/project.pbxproj
```diff
@@ -177,6 +177,7 @@
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				ASSETCATALOG_COMPILER_GLOBAL_ACCENT_COLOR_NAME = AccentColor;
+				CODE_SIGN_IDENTITY = "";
```
ioannisj, line 180: These changes avoid code signing when building on generic destinations

rafaeelaudibert (approved):
CI job makes sense to me, I don't understand the changes on Makefile tho

ioannisj:
> CI job makes sense to me, I don't understand the changes on Makefile tho

The main change is removing `xcodebuild -downloadAllPlatforms` which downloads all runtimes and sdks. 

CodeQL job was failing with low disk space. I guess because it's running on limited resources? But anw, the change should be beneficial for the other build workflows as well which are green as well. 

Didn't touch any release-related scripts

Piccirello:
This looks good, though the only thing it's missing is scanning of GitHub Actions. We previously got that with the default CodeQL setup, but no longer do. Could you add that back?

ioannisj:
> This looks good, though the only thing it's missing is scanning of GitHub Actions. We previously got that with the default CodeQL setup, but no longer do. Could you add that back?

~@Piccirello  Sorry was OOO and missed this. Would [this](https://github.com/PostHog/posthog-ios/pull/433) do the job~

Okay I see this was already done with https://github.com/PostHog/posthog-ios/pull/429, using a matrix as well which is a better approach. Thanx

Issue #421, opened 2026-01-01
App hang on Mac Catalyst: processInfo.hostName blocks main thread during SDK initialization


sbenodiz:
### Version

**PostHog iOS SDK version:** 3.36.2

### Steps to Reproduce

 When running an iOS app on macOS via Mac Catalyst, the app hangs for 2+ seconds during PostHog SDK initialization. The hang is caused by processInfo.hostName performing a synchronous DNS lookup on the main thread.

  Environment

  - PostHog iOS SDK version: 3.19.x (issue introduced in 3.19.3 via PR #287)
  - Platform: Mac Catalyst (iPad app running on macOS)
  - macOS version: macOS 15.x (Sequoia)
  - Device: MacBook Pro (Apple Silicon)

  Stack Trace

  PostHogSDK.setup(config)
    → PostHogSDK.installIntegrations
      → PostHogAppLifeCycleIntegration.install
        → PostHogAppLifeCycleIntegration.captureAppInstallOrUpdated
          → PostHogSDK.capture
            → PostHogSDK.buildProperties
              → PostHogContext.staticContext
                → PostHogContext.theStaticContext.getter
                  → properties["$device_name"] = processInfo.hostName  // ← BLOCKS HERE
                    → _NSSwiftProcessInfo.hostName.getter
                      → -[NSHost name]
                        → -[NSHost blockingResolveUntil:]
                          → semaphore_wait_trap  // Blocking DNS lookup

  Root Cause

  In PostHogContext.swift around line 89, for Mac Catalyst apps:

  properties["$device_name"] = processInfo.hostName

  The ProcessInfo.hostName property performs a synchronous DNS reverse lookup via NSHost, which can block for seconds (or up to 60 seconds on network timeout). Apple explicitly warns against using this on the main thread:

  "All calls to NSHost are synchronous — they block until the response is fetched. If a network error occurs, this could result in a 60 second delay."
  — https://www.cocoawithlove.com/2009/11/drop-in-fix-for-problems-with-nshost.html

  "Host names don't make practical sense on iOS... Don't use hostName on iOS."
  — https://developer.apple.com/forums/thread/778377

  Suggested Fix

  Option 1: Use SCDynamicStoreCopyComputerName (non-blocking)

  import SystemConfiguration

  #if targetEnvironment(macCatalyst)
  if let computerName = SCDynamicStoreCopyComputerName(nil, nil) as String? {
      properties["$device_name"] = computerName
  }
  #endif

[...]

PR #422, opened 2026-01-02
fix: macos device name no dns lookup

marandaneto:
Closes https://github.com/PostHog/posthog-ios/issues/421
lightweight version from https://github.com/devicekit/DeviceKit

run tests

## :pencil: Checklist

### DIFF
--- a/PostHog/PostHogContext.swift
+++ b/PostHog/PostHogContext.swift
@@ -236,8 +236,9 @@ class PostHogContext {
             // Mac13,x is Mac Studio
             // Future models might follow similar patterns
             if model.hasPrefix("Mac14,2") || model.hasPrefix("Mac14,15") ||
-               model.hasPrefix("Mac15,3") || model.hasPrefix("Mac15,6") ||
-               model.hasPrefix("Mac15,7") {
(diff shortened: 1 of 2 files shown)


PostHog/PostHogContext.swift
```diff
@@ -205,6 +209,52 @@ class PostHogContext {
         return String(cString: machine)
     }
 
+    /// Converts Mac hardware identifiers to user-friendly names
+    /// For example: "MacBookPro18,3" -> "MacBook Pro"
+    private func macModelToFriendlyName(_ model: String) -> String {
```
ioannisj, line 214: Since we don’t map device models for iOS either, should we skip mapping for mac as well and do this service side instead as part of https://github.com/PostHog/posthog/issues/18025?

ioannisj (approved):
Left a comment but approved from me to unblock you. Thanx, this was a silly mistake

PostHog/PostHogContext.swift
marandaneto: yep i think the right solution is backend, but removing would be bad (break. change), i would vote for fixing this for now (performance decrease/kinda wrong info), and later remove the sdk change part with a backend change that makes it back compatible

PR #423, opened 2026-01-02
fix: parse $app_build as integer when possible

marandaneto:
fix https://x.com/hoppsen1/status/2006623742084993495?s=20
android is a number already

running and unit tests

## :pencil: Checklist

### DIFF
--- a/PostHog/App Life Cycle/PostHogAppLifeCycleIntegration.swift
+++ b/PostHog/App Life Cycle/PostHogAppLifeCycleIntegration.swift
@@ -123,7 +123,12 @@ final class PostHogAppLifeCycleIntegration: PostHogIntegration {
             if previousVersion != nil {
                 props["previous_version"] = previousVersion
             }
-            props["previous_build"] = previousVersionCode
+            // Try to parse as Int first, then fallback to String
+            if let prevBuildInt = Int(previousVersionCode) {
+                props["previous_build"] = prevBuildInt
+            } else {
+                props["previous_build"] = previousVersionCode
+            }
         }
 
         var syncDefaults = false
@@ -134,7 +139,12 @@ final class PostHogAppLifeCycleIntegration: PostHogIntegration {
         }
 
         if versionCode != nil {
-            props["build"] = versionCode
+            // Try to parse as Int first, then fallback to String
+            if let buildInt = Int(versionCode) {
+                props["build"] = buildInt
+            } else {
+                props["build"] = versionCode
+            }
             userDefaults.setValue(versionCode, forKey: "PHGBuildKeyV2")
             syncDefaults = true
         }
--- a/PostHog/App Life Cycle/PostHogAppLifeCycleIntegration.swift
+++ b/PostHog/App Life Cycle/PostHogAppLifeCycleIntegration.swift
@@ -141,7 +141,7 @@ final class PostHogAppLifeCycleIntegration: PostHogIntegration {
             syncDefaults = true
         }
 
-        if let versionCode = versionCode {
+        if let versionCode {
             // Try to parse as Int first, then fallback to String
             if let buildInt = Int(versionCode) {
                 props["build"] = buildInt
--- a/PostHog/App Life Cycle/PostHogAppLifeCycleIntegration.swift
+++ b/PostHog/App Life Cycle/PostHogAppLifeCycleIntegration.swift
@@ -1,5 +1,3 @@
-// swiftlint:disable cyclomatic_complexity
-
 //
 //  PostHogAppLifeCycleIntegration.swift
 //  PostHog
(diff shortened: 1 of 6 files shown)


PostHog/App Life Cycle/PostHogAppLifeCycleIntegration.swift
```diff
@@ -123,7 +123,14 @@ final class PostHogAppLifeCycleIntegration: PostHogIntegration {
             if previousVersion != nil {
                 props["previous_version"] = previousVersion
             }
-            props["previous_build"] = previousVersionCode
+            // Try to parse as Int first, then fallback to String
+            if let previousVersionCode = previousVersionCode {
```
ioannisj: Nit: we can just `if let previousVersionCode {` here I think

PostHog/PostHogContext.swift
```diff
@@ -38,7 +38,14 @@ class PostHogContext {
...
         if let appBuild = infoDictionary?["CFBundleVersion"] {
-            properties["$app_build"] = appBuild
+            // Try to parse as Int first, then fallback to original value
+            if let appBuildString = appBuild as? String,
+               let appBuildInt = Int(appBuildString)
+            {
+                properties["$app_build"] = appBuildInt
+            } else {
+                properties["$app_build"] = appBuild
+            }
```
ioannisj: Can we simplify by casting directly to String first, since that will be the fallback?

```

if let appBuild = infoDictionary?["CFBundleVersion"] as? String {
    if let appBuildInt = Int(appBuild) {
        properties["$app_build"] = appBuildInt
    } else {
        properties["$app_build"] = appBuild
    }
}
```

ioannisj (approved):

ptrkstr:
I recently encountered @ezanalemma's mismatch issue because:
- developing locally my project has the default `1` value
- distributed through Github Action CI it takes on the `${{ github.run_number }}.${{ github.run_attempt }}` i.e. `12.1`.

A part of me thinks that it should always be treated as a String due to the [Apple docs](https://developer.apple.com/documentation/bundleresources/information-property-list/cfbundleversion) containing the following:

> You can also abbreviate the build version by using only one or two integers, where missing integers in the format are interpreted as zeros. For example, 0 specifies 0.0.0, 10 specifies 10.0.0, and 10.5 specifies 10.5.0.

So even though one may see a single integer, this is a shorthand for Apple  seeing it padded with zeros and periods.

marandaneto:
@ptrkstr its more common that the version x.x.x comes from `CFBundleShortVersionString` and we set that one as string in `$app_version`

> You can also abbreviate the build version by using only one or two integers, where missing integers in the format are interpreted as zeros. For example, 0 specifies 0.0.0, 10 specifies 10.0.0, and 10.5 specifies 10.5.0.

`CFBundleVersion` usually is a number, like build 1, 2 or 3, despite being possible to also be x, x.x or x.x.x

we try to parse `CFBundleVersion` as `Int` and fallback to String if not possible, so `$app_build` should work both ways

ptrkstr:
That makes sense, thanks for answering @marandaneto!

ptrkstr:
@marandaneto wanted to ask a slight tangent question here to avoid raising an unnecessary PR:

These lifecycle events:
- `Application Installed`
- `Application Updated`
- `Application Opened`

include an additional `build` property which duplicates the existing `$app_build`. 

image

Given `$app_build` is already included on every event, I'm curious if there has been a consideration to remove the redundant `build` property?

<details>
  <summary>Also occurs for version/$app_version</summary>
  image
</details>

If there is reason to keep it for legacy reasons or to match other SDKs, there is a slight discrepancy of parsing `build` whereby:
- `Application Installed` checks for Int parsing first and falls back to String ✅ [source](https://github.com/PostHog/posthog-ios/blob/a6f3ab1ee07df004aa480474ecbfa78cd24920b2/PostHog/AppLifeCycle/PostHogAppLifeCycleIntegration.swift#L150-L159)
- `Application Updated` checks for Int parsing first and falls back to String ✅ [source](https://github.com/PostHog/posthog-ios/blob/a6f3ab1ee07df004aa480474ecbfa78cd24920b2/PostHog/AppLifeCycle/PostHogAppLifeCycleIntegration.swift#L133-L140)
- `Application Opened` always uses String ⚠️ [source](https://github.com/PostHog/posthog-ios/blob/a6f3ab1ee07df004aa480474ecbfa78cd24920b2/PostHog/AppLifeCycle/PostHogAppLifeCycleIntegration.swift#L186-L200)

I think it would be worth having a reusable function that all `build`/`$app_build` locations use to ensure the build is extracted in a consistent way.

marandaneto:
good point, its to keep back compatibility, `build` and `previous_build`, i think both 

we do right here:
https://github.com/PostHog/posthog-ios/blob/a6f3ab1ee07df004aa480474ecbfa78cd24920b2/PostHog/AppLifeCycle/PostHogAppLifeCycleIntegration.swift#L153-L155
but not here https://github.com/PostHog/posthog-ios/blob/a6f3ab1ee07df004aa480474ecbfa78cd24920b2/PostHog/AppLifeCycle/PostHogAppLifeCycleIntegration.swift#L195-L197
wanna open a pr?

ptrkstr:
@marandaneto have made the change here: https://github.com/PostHog/posthog-ios/pull/576

PR #424, opened 2026-01-02
feat: SDK Compliance

jose-sequeira:

We need a language-agnostic way to verify that all PostHog SDKs implement the capture API correctly. The [SDK Test Harness](https://github.com/PostHog/posthog-sdk-test-harness) provides a shared contract and test runner. This PR integrates it into this SDK.

This adds:
- An SDK adapter that wraps this SDK and exposes the standard REST interface for the test harness
- A CI workflow that runs compliance tests on every PR and push to the default branch
- A Dockerfile for running the adapter in the test harness environment

The test harness runs a mock PostHog server, exercises the SDK through the adapter, and verifies behavior matches the contract defined in [CONTRACT.yaml](https://github.com/PostHog/posthog-sdk-test-harness/blob/main/CONTRACT.yaml) (event format, retries, compression, deduplication, etc.).

#skip-changelog

The compliance tests run as part of this PR's CI. The adapter was also tested locally via `docker-compose`.

## :pencil: Checklist


.github/workflows/sdk-compliance.yml
```diff
@@ -0,0 +1,188 @@
...
+# NOTE: This workflow is adapted from PostHog/posthog-sdk-test-harness/.github/workflows/test-sdk-action.yml
+# Key difference: Uses HYBRID RUNNER architecture because iOS SDK requires macOS (Darwin frameworks)
+# - Standard action: Runs everything in Docker on ubuntu-latest
+# - This workflow: Runs adapter natively on macos-latest, test harness in Docker
+#
+# Why: iOS SDK depends on SystemConfiguration, Foundation, UIKit which don't exist on Linux
+
+on:
+  push:
+    branches: [main, master]
```
ioannisj:
```suggestion
    branches: [main]
```

sdk_compliance_adapter/Sources/main.swift
```diff
@@ -0,0 +1,286 @@
...
+
+// Middleware to log all requests
+app.middleware.use(RouteLoggingMiddleware())
+
+// Health endpoint
+app.get("health") { req async throws -> Response in
+    let packageVersion = "3.0.0" // TODO: Read from Package.swift or version file
+    let health = [
+        "sdk_name": "posthog-ios",
+        "sdk_version": packageVersion,
```
ioannisj: These are publicly available in PostHog module in [PostHogVersion.swift](https://github.com/PostHog/posthog-ios/blob/a462aa29d862bd3c51ca76ddd025eae66f66647b/PostHog/PostHogVersion.swift#L12-L14), so I believe we can use them directly

sdk_compliance_adapter/Sources/URLSessionSwizzle.swift
```diff
@@ -0,0 +1,39 @@
+import Foundation
+import ObjectiveC
+
+/// Swizzle URLSession(configuration:) initializer to inject our RequestInterceptor
```
ioannisj: Since we now expose a `urlSessionConfiguration` in config, do we still need this? Seems dead code. 

If this works, however, I would lean into using swizzling for these tests instead of adding a new config option

sdk_compliance_adapter/run_tests.sh
```diff
@@ -0,0 +1,71 @@
...
+fi
+
+if ! docker info &> /dev/null; then
+    echo "❌ Error: Docker not running. Please start Docker Desktop"
+    exit 1
+fi
+
+# Build the adapter with TESTING flag
+echo "📦 Building iOS adapter..."
+swift build -c release -Xswiftc -DTESTING
```
ioannisj: In another comment in this PR you mentioned building in debug to be much faster. Should we do this here as well?

ioannisj (approved):
Left a couple of comments. Really good work, in general this LG to me! My only comment/concern is to avoid adding the extra `urlSessionConfiguration` config. This could indeed be helpful for people and advanced usage but could potentially complicate support when people misconfigure?

PostHog/PostHogConfig.swift
```diff
@@ -185,6 +185,11 @@ public typealias BeforeSendBlock = (PostHogEvent) -> PostHogEvent?
         set { setSurveysConfig(newValue) }
     }
 
+    /// Optional custom URLSessionConfiguration for network requests
+    /// If not set, uses URLSessionConfiguration.default
+    /// Useful for testing, proxying, or custom network configurations
+    @objc public var urlSessionConfiguration: URLSessionConfiguration?
```
marandaneto, line 191: we have this for android as well (specifically for proxy)
https://github.com/PostHog/posthog-android/blob/98beb0bfe97a3b61c8732cce797c0f28eea278f6/posthog/src/main/java/com/posthog/PostHogConfig.kt#L195-L210
@ioannisj if we'd add a similar feat on iOS, is urlSessionConfiguration the way to go? (eg config.connectionProxyDictionary?)
then i'd add some examples

marandaneto: maybe we should add an extension to make this easier
eg

```
struct HTTPProxy {
    let host: String
    let port: Int
}

extension URLSessionConfiguration {
    func setHTTPProxy(_ proxy: HTTPProxy) {
        connectionProxyDictionary = [
            kCFNetworkProxiesHTTPEnable as String: true,
            kCFNetworkProxiesHTTPProxy as String: proxy.host,
            kCFNetworkProxiesHTTPPort as String: proxy.port,

            kCFNetworkProxiesHTTPSEnable as String: true,
            kCFNetworkProxiesHTTPSProxy as String: proxy.host,
            kCFNetworkProxiesHTTPSPort as String: proxy.port
        ]
    }
}
```

jose-sequeira: Ye this is definitely one of the pain points I felt in both android and iOS and I wasn't sure of how could I do it without tampering with the public API. 

@marandaneto @ioannisj Would one of you mind handling this specific part?

marandaneto: @jose-sequeira i think this is not a blocker (pr is approved already) but i'll let @ioannisj double check
you could do this automatically with swizzle but i dont recommend since we get the proxy feat for free with this public API

ioannisj: Yeah, discussed with @marandaneto and agreed this turns out to be okay since it will add the ability for a proxy, something that Android already has and people requested. I'll add some usage examples in the doc comments at least

Issue #436, opened 2026-01-16
setPersonProperties support ($set event)


marandaneto:
### Description

Relates to https://github.com/PostHog/posthog-android/issues/194

https://github.com/PostHog/posthog-js/blob/26adbc8bc9271090cbed667cf81c24c1639b4a1f/packages/browser/src/posthog-core.ts#L2316-L2373

https://posthog.com/docs/libraries/js/usage#setting-person-properties
https://posthog.com/docs/product-analytics/person-properties

- [x] Android
- [x] iOS
- [ ] Flutter
- [ ] RN

Check Android PR  as well

PR #441, opened 2026-01-21
feat: add setPersonProperties methods

ioannisj:

Implements #436 #442 

Unit and manual testing

## :pencil: Checklist


PostHog/PostHogSDK.swift
```diff
@@ -34,6 +34,8 @@ let maxRetryDelay = 30.0
     private let groupsLock = NSLock()
     private let flagCallReportedLock = NSLock()
     private let personPropsLock = NSLock()
+    private let cachedPersonPropertiesLock = NSLock()
```
ioannisj, line 37: This class body is getting way to large, we should start splitting this up soon

marandaneto: do it :)

PR #443, opened 2026-01-21
feat: add `$is_testflight` and `$is_sideloaded` properties

### LINKED ISSUE #117: Set `is_testflight` and `is_sideloaded` property
### Description

Relates to https://github.com/PostHog/posthog-android/issues/113
Sets a property if the running version was installed via testflight or its sideloaded.

- [ ] iOS
- [ ] Android
- [ ] RN
- [ ] Flutter will get it for free once Android and iOS are updated

ioannisj:

Closes #117 

Adds two new static context properties to detect app distribution method:

`$is_testflight` true if installed via TestFlight (detected via sandboxReceipt)
`$is_sideloaded` true if installed outside App Store/TestFlight (has embedded provisioning profile)

## ToDo
- Update backend taxonomy if needed

## :pencil: Checklist

### DIFF
--- a/PostHog/PostHogStorage.swift
+++ b/PostHog/PostHogStorage.swift
@@ -72,7 +72,9 @@ func getBundleIdentifier() -> String {
     #if TESTING // only visible to test targets
         return Bundle.main.bundleIdentifier ?? "com.posthog.test"
     #else
-        return Bundle.main.bundleIdentifier!
+        // Can be nil for command-line tools, XCTest hosts, Swift Playgrounds etc
+        // Should theoretically never be nil for a shipping app
+        return Bundle.main.bundleIdentifier ?? "com.posthog.unknown"
     #endif
 }
 


PostHog/PostHogContext.swift
```diff
@@ -460,4 +462,94 @@ class PostHogContext {
...
+        /// Running on iOS Simulator
+        case emulator
+        /// Detection failed or unknown distribution method
+        case unknown
+    }
+
+    /// Detects how the app was installed.
+    static let installSource: InstallSource = {
+        #if targetEnvironment(simulator)
+            return .simulator
```
marandaneto: for this we have $is_emulator as well right

PostHog/PostHogContext.swift
```diff
@@ -44,6 +44,7 @@ class PostHogContext {
                 properties["$app_build"] = appBuild
             }
         }
+        properties["$app_install_source"] = PostHogContext.installSource.rawValue
```
marandaneto: fine by me maybe just check if its also possible on android so we know we can add something similar
ideally we have common properties like this across the SDKs so they can be used in insights equally
if this type of info isnt possible at all, maybe it'd make more sense to have just booleans like the description suggests so we only added the necessary information eg is_side_loaded, etc

eg if i want to filter emulators on my insights, i should be able to use the very same flag instead of is_simulator on android and then $app_install_source = simulator on iOS, thats no good

PostHog/PostHogContext.swift
ioannisj: Yeah, mentioned in PR body `Already tracked with $is_emulator so we could potentially drop this`. I think this will just add confusion so I'll remove it

PostHog/PostHogContext.swift
ioannisj: yep, makes sense, we should aim for a cross sdk “install source” that’s semantically consistent, then optionally add platform-specific drilldowns (like we already have $is_mac_catalyst_app for iOS)

Asked Claude and this is the possible mapping for the enum on Android

```kotlin
fun getInstallSource(context: Context): InstallSource {
    // 1. Check if debug build
    if (BuildConfig.DEBUG || 
        (context.applicationInfo.flags and ApplicationInfo.FLAG_DEBUGGABLE) != 0) {
        return InstallSource.DEVELOPMENT
    }
    
    // 2. Check installer package name
    val installer = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
        context.packageManager.getInstallSourceInfo(context.packageName).installingPackageName
    } else {
        @Suppress("DEPRECATION")
        context.packageManager.getInstallerPackageName(context.packageName)
    }
    
    return when (installer) {
        "com.android.vending" -> InstallSource.PLAY_STORE
        "com.amazon.venezia" -> InstallSource.AMAZON_STORE
        "com.sec.android.app.samsungapps" -> InstallSource.SAMSUNG_STORE
        "com.huawei.appmarket" -> InstallSource.HUAWEI_STORE
        null -> InstallSource.SIDELOADED  // Manual APK install
        else -> InstallSource.ENTERPRISE  // Custom installer, possibly MDM
    }
}
```

So we could expose this shared property with some common values between the two platforms like `store`, `sideloaded`, `development`, maybe `enterprise` (or better mdm?). 

then, for android-only depth, add something like `$android_installer: "com.android.vending"`. Sounds good?

marandaneto: check this out https://github.com/getsentry/sentry-java/blob/b193867457267c1569d3bfa733c888e6c630f63c/sentry-android-core/src/main/java/io/sentry/android/core/ContextUtils.java#L353-L357
we cant use getInstallSourceInfo
getInstallerPackageName is ok

marandaneto: btw do not trust code before running it since this type of stuff usually is under OS access control

marandaneto: i do think having the store name is useful tho, so +1 on adding this

PostHog/PostHogContext.swift
```diff
@@ -44,10 +44,11 @@ class PostHogContext {
                 properties["$app_build"] = appBuild
             }
         }
+        properties["$is_testflight"] = PostHogContext.isTestFlight
+        properties["$is_sideloaded"] = PostHogContext.isSideloaded
+
+        properties["$app_namespace"] = getBundleIdentifier()
```
marandaneto, line 50: https://github.com/PostHog/posthog-ios/blob/1783865d79a1cabc472cf2d56a1fe3f797417b52/PostHog/PostHogStorage.swift#L75
we are forcing !, not sure if this is ever nil but i'll crash the app if so, probably better making getBundleIdentifier return nullable and check here

ioannisj: We are already doing that, so I would expect people to complain by now. It can be nil when running tests or command line tool. This is required for storage so I added a sensible fallback instead

marandaneto: true that, i'd still try to be defensive tho.
and forcing the ! as before was an oversight.
if the bundle identifier is nil and we need for storage, we'd just return early and do not have the storage feature even if the SDK is noop (ideally would do in-memory only)
same for $app_namespace, today isnt set, which would break insights if someone filters $app_namespace as non nil for eg just web events, now it'd be something unknown.
your call here, no strong opinions since its unlikely to be an issue as you mentioned

ioannisj: I'll merge this one as-is. Making storage optional feels like a big change that needs it's own PR. It would be a good chance to abstract this away as well, it will help with unit tests cause currently we run them in sequence because file operations bleed between test cases

PR #452, opened 2026-01-24
feat: disable or enable session replay plugins based on remote config

### LINKED ISSUE #351: Respect `consoleLogRecordingEnabled` remote config
### Description

```json
{
    "sessionRecording": {
        "consoleLogRecordingEnabled": true
        ...
    }
}
```

We only do it via code now, but we could also read from remote config.
Android needs an interceptor added specifically, but if the interceptor is already in place, we could always enable/disable it remotely as well.

- [ ] Android
- [x] iOS
- [ ] Flutter - (TODO: no logging impl. yet)
- [ ] RN

ioannisj:

Closes #351 

Adds the ability to enable/disable session replay plugins via PostHog project settings. 

The current configs options affected are 
- `sessionReplayConfig.captureLogs` (key sessionRecording.consoleLogRecordingEnabled)
- `sessionReplayConfig.captureNetworkTelemetry` (key capturePerformance)

Local config takes precedence. A plugin must first be **enabled** locally before remote config is evaluated.

Local Config | Remote Config | Result
-- | -- | --
false | true | Disabled (disabling locally takes precedence)
false | false | Disabled (disabling locally takes precedence)
true | true | Enabled (local and remote match)
true | false | Disabled (enabled locally but remotely disabled)

### Limitations / Future work
- Only works when `config.remoteConfig = true`. 
	- Future improvements could read and cache remote config from `/flags?config=true` response as a fallback when `config.remoteConfig = false`

- Unit tests and manual testing

## :pencil: Checklist

### DIFF
--- a/PostHog/Replay/PostHogReplayIntegration.swift
+++ b/PostHog/Replay/PostHogReplayIntegration.swift
@@ -29,6 +29,7 @@
 
         private let windowViewsLock = NSLock()
         private let windowViews = NSMapTable<UIWindow, ViewTreeSnapshotStatus>.weakToStrongObjects()
+        private let installedPluginsLock = NSLock()
         private var applicationEventToken: RegistrationToken?
         private var applicationBackgroundedToken: RegistrationToken?
         private var applicationForegroundedToken: RegistrationToken?
@@ -179,15 +180,21 @@
             // Install plugins
             let pluginTypes = postHog.config.sessionReplayConfig.getPluginTypes()
             let remoteConfig = postHog.remoteConfig?.getRemoteConfig()
-            installedPlugins = []
-            for pluginType in pluginTypes {
-                if !pluginType.isEnabledRemotely(remoteConfig: remoteConfig) {
-                    hedgeLog("[Session Replay] Plugin \(pluginType) skipped - disabled by cached remote config")
-                    continue
+            let pluginsToStart = installedPluginsLock.withLock {
+                installedPlugins = []
+                for pluginType in pluginTypes {
+                    if !pluginType.isEnabledRemotely(remoteConfig: remoteConfig) {
+                        hedgeLog("[Session Replay] Plugin \(pluginType) skipped - disabled by cached remote config")
+                        continue
+                    }
+                    let plugin = pluginType.init()
+                    installedPlugins.append(plugin)
                 }
-                let plugin = pluginType.init()
+                return installedPlugins
+            }
+
+            for plugin in pluginsToStart {
                 plugin.start(postHog: postHog)
-                installedPlugins.append(plugin)
             }
 
             // Start listening to application background events and pause all plugins
@@ -224,10 +231,14 @@
             remoteConfigLoadedToken = nil
 
             // stop plugins
-            for plugin in installedPlugins {
+            let pluginsToStop = installedPluginsLock.withLock {
+                defer { installedPlugins = [] }
+                return installedPlugins
+            }
+
+            for plugin in pluginsToStop {
                 plugin.stop()
             }
-            installedPlugins = []
         }
 
         func isActive() -> Bool {
@@ -242,13 +253,15 @@
         }
 
         private func pauseAllPlugins() {
-            for plugin in installedPlugins {
+            let pluginsToPause = installedPluginsLock.withLock { installedPlugins }
+            for plugin in pluginsToPause {
                 plugin.pause()
             }
         }
 
         private func resumeAllPlugins() {
-            for plugin in installedPlugins {
+            let pluginsToResume = installedPluginsLock.withLock { installedPlugins }
+            for plugin in pluginsToResume {
                 plugin.resume()
             }
         }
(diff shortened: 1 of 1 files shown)


PostHog/Utils/MulticastCallback.swift
```diff
@@ -0,0 +1,59 @@
...
+/// let token = onConfigLoaded.subscribe { config in
+///     print("Config loaded: \(config)")
+/// }
+///
+/// // Invoke all subscribers
+/// onConfigLoaded.invoke(someConfig)
+///
+/// // Token automatically unsubscribes when deallocated
+/// ```
+final class MulticastCallback<T> {
```
ioannisj: This is a generalized pattern already used in the codebase. Did not refactor existing code to keep the scope small for this PR, but will do with a future PR

PostHog/Replay/Plugins/Network/PostHogSessionReplayNetworkPlugin.swift
```diff
@@ -52,6 +54,18 @@
...
 
+        static func isEnabledRemotely(remoteConfig: [String: Any]?) -> Bool {
+            guard let capturePerformanceValue = remoteConfig?["capturePerformance"] else {
+                return true
+            }
+            // When disabled, capturePerformance is `false` (boolean)
+            if let isEnabled = capturePerformanceValue as? Bool {
+                return isEnabled
+            }
+            // When enabled, capturePerformance is an object (we ignore the keys for now)
```
marandaneto, line 68: which configs are within the capturePerformance config?

PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -239,6 +253,30 @@
...
+
+            let allPluginTypes = postHog.config.sessionReplayConfig.getPluginTypes()
+
+            for pluginType in allPluginTypes {
+                let isEnabled = pluginType.isEnabledRemotely(remoteConfig: remoteConfig)
+                let installedIndex = installedPlugins.firstIndex { type(of: $0) == pluginType }
+
+                if let index = installedIndex, !isEnabled {
+                    // Installed, but disabled in remote
+                    installedPlugins[index].stop()
```
marandaneto: Should we check that installedPlugins has the index (index) and the same for remove to avoid out-of-bounds issues?

PostHog/Utils/MulticastCallback.swift
marandaneto: should we call it PostHogMulticastCallback to avoid naming clashes with other things?

PostHog/PostHogApi.swift
```diff
@@ -20,6 +20,10 @@ class PostHogApi {
     func sessionConfig() -> URLSessionConfiguration {
         let config = URLSessionConfiguration.default
 
+        #if DEBUG
+            config.requestCachePolicy = .reloadIgnoringLocalCacheData
```
marandaneto: whats that used for?

PostHog/Replay/Plugins/Network/PostHogSessionReplayNetworkPlugin.swift
ioannisj: These are the configs inside [capturePerformance](https://github.com/PostHog/posthog-js/blob/bf96f5fd2f0309a5a8b69ceafc155d0497832e11/packages/types/src/posthog-config.ts#L145)

`network_timing` seems to be the one relevant to us but from my understanding js sdk seems to be ignoring it as well [here](https://github.com/PostHog/posthog-js/blob/bb1f8722295ec56ca589f8e7e4e950683a06e3be/packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts#L480-L482). @PostHog/team-replay can confirm?

marandaneto: looks like it does? 
https://github.com/PostHog/posthog-js/blob/bb1f8722295ec56ca589f8e7e4e950683a06e3be/packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts#L477C15-L479
otherwise web_vitals might be enabled, network_timing disabled, and you'll wrongly enable network_timing just because capturePerformance is an object

PostHog/PostHogApi.swift
ioannisj: Since config is a static resource served from CDN, the response includes cache-control headers. The default URLSession cache policy is `useProtocolCachePolicy` which respects `max-age` value served from CDN

When we change settings in the PostHog UI, the edge is invalidated, but iOS keeps working off the local cache until cache expires. 

Setting cache policy to `.reloadIgnoringLocalCacheData` will ensure that we always hit the network. I added it a development convenience since while testing, the config values I was getting were stale. 

I see that max-age:300 (5 mins) so for release builds this is short enough to be eventually consistent (and avoid extra network calls). We could use `.reloadRevalidatingCacheData` in release builds which I think will do a cheap net call to validate cache, then load from remote or use local value accordingly. wdyt?

PostHog/Replay/Plugins/Network/PostHogSessionReplayNetworkPlugin.swift
ioannisj: That's checking the local config I think?

PostHog/Replay/PostHogReplayIntegration.swift
ioannisj: Good point, but all changes to `installedPlugins` happen on main thread. Since we are reading the index right above in the same sync block I thin the check is redundant?

PostHog/Replay/Plugins/Network/PostHogSessionReplayNetworkPlugin.swift
ioannisj: pushed a change to use false as a fallback which matches js

PostHog/Utils/MulticastCallback.swift
ioannisj: Good idea, done

PostHog/Replay/PostHogReplayIntegration.swift
ioannisj: made `installedPlugins` thread safe

PostHog/PostHogApi.swift
ioannisj: Ended up using `reloadRevalidatingCacheData` for both debug and release builds. It's an extra quick net call checking if resource is expired or not which ensures that we don't work with stale remote config.

PostHog/Replay/Plugins/Console Logs/PostHogSessionReplayConsoleLogsPlugin.swift
```diff
@@ -44,6 +46,14 @@
             hedgeLog("[Session Replay] Console logs plugin paused")
         }
 
+        // see: https://github.com/PostHog/posthog-js/blob/main/packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts#L457-L461
```
marandaneto, line 49: @ioannisj better to use permalink, this will be outdated soon, it is already

PostHog/Replay/Plugins/Network/PostHogSessionReplayNetworkPlugin.swift
```diff
@@ -52,6 +54,21 @@
...
+                // No remote config means disabled
+                return false
+            }
+
+            // When capturePerformance is a boolean, use it directly
+            if let isEnabled = capturePerformanceValue as? Bool {
+                return isEnabled
+            }
+            // When enabled, capturePerformance is an object (we ignore the keys for now)
+            return true
```
marandaneto, line 69: we should have used `network_timing` here

marandaneto: mentioned here https://github.com/PostHog/posthog-ios/pull/452#discussion_r2740579253

PR #455, opened 2026-01-29
feat: Add `getFeatureFlagResult` method to client

dustinbyrne:
Adds a new `getFeatureFlagResult()` method that returns all feature flag information (value, variant, and payload) in a single call, aligning the iOS SDK with other SDKs.

Previously, retrieving a flag's payload required a separate `getFeatureFlagPayload()` call, and there was no unified way to get all flag data at once. The new method:

- Returns a `PostHogFeatureFlagResult` with `key`, `enabled`, `variant`, `payload`, and `payloadAs<T>()`
- Becomes the canonical source for flag data
- Tracks `$feature_flag_called` events

Existing methods are refactored to delegate to `getFeatureFlagResult()`, without affecting behavior.

Related to https://github.com/PostHog/posthog/issues/43520

Additional unit tests, existing tests (`getFeatureFlag`, `isFeatureEnabled`, `getFeatureFlagPayload`) passing without change.

## :pencil: Checklist

### DIFF
--- a/PostHog/PostHogRemoteConfig.swift
+++ b/PostHog/PostHogRemoteConfig.swift
@@ -572,10 +572,28 @@ class PostHogRemoteConfig {
     }
 
     func getFeatureFlagResult(_ key: String) -> PostHogFeatureFlagResult? {
-        let flagValue = getFeatureFlag(key)
+        var flagValue: Any?
+        var payloadValue: Any?
+
+        featureFlagsLock.withLock {
+            flagValue = getCachedFeatureFlags()?[key]
+            payloadValue = getCachedFeatureFlagPayload()?[key]
+        }
+
         guard flagValue != nil else { return nil }
 
-        let payload = getFeatureFlagPayload(key)
+        let payload: Any?
+        if let stringValue = payloadValue as? String {
+            do {
+                payload = try JSONSerialization.jsonObject(with: stringValue.data(using: .utf8)!, options: .fragmentsAllowed)
+            } catch {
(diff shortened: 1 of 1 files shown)


PostHog/Models/PostHogFeatureFlagResult.swift
```diff
@@ -0,0 +1,82 @@
...
+@objc(PostHogFeatureFlagResult)
+public class PostHogFeatureFlagResult: NSObject {
+    /// The key of the feature flag.
+    @objc public let key: String
+
+    /// Whether the feature flag is enabled.
+    ///
+    /// This is `true` if the flag value is:
+    /// - A boolean `true`
+    /// - A non-empty string (variant)
```
ioannisj: // non-empty string (variant)

We don't check for this in `getFeatureFlagResult`, should we?

PostHog/PostHogRemoteConfig.swift
```diff
@@ -571,6 +571,33 @@ class PostHogRemoteConfig {
         return value
     }
 
+    func getFeatureFlagResult(_ key: String) -> PostHogFeatureFlagResult? {
+        let flagValue = getFeatureFlag(key)
+        guard flagValue != nil else { return nil }
+
+        let payload = getFeatureFlagPayload(key)
```
ioannisj: nit: both `getFeatureFlag` and `getFeatureFlagPayload` acquire the featureFlagsLock separately. For performance we could consider locking just once in this function, but I also see that this is more readable

PR #466, opened 2026-02-09
fix: prevent crashes from non JSON-serializable property types

### LINKED ISSUE #460: Crash in getPersonPropertiesHash when userProperties contains Date objects (v3.40.0)
## Description

After updating from PostHog iOS SDK 3.37.0 to 3.40.0, the app crashes immediately on launch with `NSInvalidArgumentException: Invalid type in JSON write (__NSTaggedDate)`.

## Environment

- **PostHog iOS SDK version**: 3.40.0 (works fine on 3.37.0)
- **iOS version**: iOS 18.x
- **Xcode version**: 16.x
- **Swift version**: 5.9+

## Crash Log

```
*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: 'Invalid type in JSON write (__NSTaggedDate)'
*** First throw call stack:
(0x19db94964 0x19aaa9814 0x19b14fc88 0x19b1d4628 0x19b1d7e3c 0x19b1d4628 0x19b1d7e3c 0x19bb4849c 0x19bb49448 ...)
```

## Root Cause

The crash occurs in the new `getPersonPropertiesHash()` function introduced in 3.40.0 (around line 630 in `PostHogSDK.swift`):

```swift
private func getPersonPropertiesHash(
    distinctId: String,
    userPropertiesToSet: [String: Any]?,
    userPropertiesToSetOnce: [String: Any]?
) -> String {
    var hashData: [String: Any] = ["distinct_id": distinctId]

    if let userPropertiesToSet {
        hashData["userPropertiesToSet"] = userPropertiesToSet
    }
    if let userPropertiesToSetOnce {
        hashData["userPropertiesToSetOnce"] = userPropertiesToSetOnce
    }

    // CRASH HERE: JSONSerialization cannot serialize Date objects
    if let jsonData = try? JSONSerialization.data(withJSONObject: hashData, options: [.sortedKeys]),
       let jsonString = String(data: jsonData, encoding: .utf8)
    {
        return jsonString
    }

    // fallback
    return hashData.description
}
```

`JSONSerialization.data(withJSONObject:)` throws an exception when the dictionary contains `Date` objects (or other non-JSON-serializable types like `Data`, custom objects, etc.).

## Steps to Reproduce

1. Update PostHog iOS SDK from 3.37.0 to 3.40.0
2. Call `identify()` or `setPersonProperties()` with a `Date` value in the properties dictionary:
   ```swift
   PostHogSDK.shared.identify("user123", userProperties: [
       "last_login": Date()  // This causes the crash
   ])
   ```
3. App crashes immediately

## Suggested Fix

The `getPersonPropertiesHash` function should sanitize the dictionary before attempting JSON serialization, similar to how other parts of the SDK handle this. Options:

**Option 1**: Use the existing `sanitizeDictionary()` function before serialization:
```swift
private func getPersonPropertiesHash(...) -> String {
    var hashData: [String: Any] = ["distinct_id": distinctId]

    if let userPropertiesToSet {
        hashData["userPropertiesToSet"] = sanitizeDictionary(userPropertiesToSet) ?? [:]
    }
    if let userPropertiesToSetOnce {
        hashData["userPropertiesToSetOnce"] = sanitizeDictionary(userPropertiesToSetOnce) ?? [:]
    }
    // ... rest of function
}
```

**Option 2**: Wrap the JSON serialization in a do-catch and handle the error gracefully:
```swift
do {
    let jsonData = try JSONSerialization.data(withJSONObject: hashData, options: [.sortedKeys])
    if let jsonString = String(data: jsonData, encoding: .utf8) {
        return jsonString
    }
} catch {
    hedgeLog("Failed to serialize person properties for hash: \(error)")
}
return hashData.description
```

## Workaround

Downgrade to PostHog iOS SDK 3.37.0 until this is fixed.

## Related

This was introduced in the `setPersonProperties` feature added in v3.40.0 (PR #455 or related).

ioannisj:
Closes #460

Add `toJSONData()` utilities that sanitizes dictionaries before JSON serialization.

Handles Date, URL, Data, UUID  and other non-JSON types by converting or dropping them gracefully instead of crashing.

Also fixed an existing issue with failing to sanitize infinity/NaN values which caused a crash discovered with the new unit tests added.
 
- Add toJSONData()
- Fix isValidObject() to reject Double.infinity, NaN, and CGFloat variants
- Update JSON serialization in PostHogSDK, PostHogQueue, PostHogApi
- Fixed some places I spotted where we force-unwrapped data!

- Added unit tests

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Next
 
 - fix: prevent crashes from non JSON-serializable property types (Date, URL, Data, infinity, NaN, etc.) ([#466](https://github.com/PostHog/posthog-ios/pull/466))
+- feat: add `$is_testflight` and `$is_sideloaded` event properties ([#443](https://github.com/PostHog/posthog-ios/pull/443))
 
 ## 3.40.0 - 2026-02-05
 
--- a/.github/workflows/ci-security.yml
+++ b/.github/workflows/ci-security.yml
@@ -22,7 +22,7 @@ jobs:
             - name: Checkout code
               uses: actions/checkout@v6
             - name: Ensure SHA pinned actions
-              uses: zgosalvez/github-actions-ensure-sha-pinned-actions@6124774845927d14c601359ab8138699fa5b70c3 # v4.0.1
+              uses: zgosalvez/github-actions-ensure-sha-pinned-actions@d5d20e15f2736816ee0e001ba8b24b54d9ffcff4 # v5.0.0
               with:
                   allowlist: |
                       actions/
--- a/.github/workflows/sdk-compliance.yml
+++ b/.github/workflows/sdk-compliance.yml
@@ -29,7 +29,7 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
 
       - name: Setup Xcode
         uses: maxim-lobanov/setup-xcode@60606e260d2fc5762a71e64e74b2174e8ea3c8bd # v1
@@ -53,7 +53,7 @@ jobs:
           docker-compose --version
 
       - name: Cache Swift build artifacts
-        uses: actions/cache@v4
+        uses: actions/cache@v5
         with:
           path: |
             sdk_compliance_adapter/.build
@@ -106,7 +106,7 @@ jobs:
 
       - name: Upload adapter logs
         if: always()
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v6
         with:
           name: adapter-logs
           path: sdk_compliance_adapter/adapter*.log
@@ -114,7 +114,7 @@ jobs:
 
       - name: Upload test report
         if: always()
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v6
         with:
           name: sdk-compliance-report
           path: report/sdk-compliance-report.md
@@ -134,7 +134,7 @@ jobs:
 
       - name: Comment on PR
         if: always() && github.event_name == 'pull_request'
-        uses: actions/github-script@v7
+        uses: actions/github-script@v8
         with:
           script: |
             const fs = require('fs');
--- a/PostHog/PostHogContext.swift
+++ b/PostHog/PostHogContext.swift
@@ -44,10 +44,11 @@ class PostHogContext {
                 properties["$app_build"] = appBuild
             }
         }
+        properties["$is_testflight"] = PostHogContext.isTestFlight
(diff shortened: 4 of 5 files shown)


PostHog/PostHogApi.swift
```diff
@@ -94,18 +94,14 @@ class PostHogApi {
...
         ]
 
-        var data: Data?
-
-        do {
-            data = try JSONSerialization.data(withJSONObject: toSend)
-        } catch {
-            hedgeLog("Error parsing the batch body: \(error)")
-            return completion(PostHogBatchUploadInfo(statusCode: nil, error: error))
+        guard let data = toJSONData(toSend) else {
```
marandaneto: arent the events props already normalized at this point? arent we doing this again?

ioannisj: Yeah correct, this would be redundant. My thinking is that we add some properties right above, but that's internal and something we control. I'll revert

ioannisj: Fixed

PostHogTests/PostHogPropertiesSerializationTest.swift
```diff
@@ -0,0 +1,346 @@
...
+        @Test("setGroupPropertiesForFlags with property type", arguments: propertyPayloads)
+        func setGroupPropertiesForFlagsWithPropertyType(_ payload: PropertyTestCase) async throws {
+            let sut = getSut()
+
+            sut.setGroupPropertiesForFlags("company", properties: payload.properties(), reloadFeatureFlags: false)
+
+            sut.reset()
+            sut.close()
+        }
+    }
```
marandaneto, line 268: what those 2 tests do? what do they assert?

CHANGELOG.md
```diff
@@ -1,7 +1,11 @@
 ## Next
 
+- fix: prevent crashes from non JSON-serializable property types (Date, URL, Data, infinity, NaN, etc.) ([#466](https://github.com/PostHog/posthog-ios/pull/466))
+
 ## 3.40.0 - 2026-02-05
 
+> ⚠️ **Warning**: This version contains a crash when using `setPersonProperties` with non-JSON-serializable types like `Date` in properties. Please upgrade to the next version.
```
marandaneto, line 8: also add this to the GH release notes

PostHogTests/PostHogPropertiesSerializationTest.swift
ioannisj: These just assert that calling these methods will not crash when the properties are serialized. Swift testing will automatically create a test case for each argument passed `propertyPayloads` in our case. I'll leave a comment here to clarify

PR #468, opened 2026-02-10
chore: add cocoapods keep alive workflow

ioannisj:

CocoaPods trunk sessions expire after 3 days of inactivity, which can cause our automated release workflow to fail if we don't publish a new version within that timeframe.

This adds a GH Actions workflow that runs daily to keep the session token alive by executing an authenticated call like `pod trunk me`

#skip-changelog

## :pencil: Checklist


marandaneto:
workaround for https://github.com/CocoaPods/trunk.cocoapods.org/pull/601

PR #470, opened 2026-02-10
fix: check network_timing key

ioannisj:

context: https://posthog.slack.com/archives/C0ACMS2BNTU/p1770734040094819

## :pencil: Checklist

### DIFF
--- a/PostHogTests/PostHogSessionReplayPluginTest.swift
+++ b/PostHogTests/PostHogSessionReplayPluginTest.swift
@@ -100,6 +100,28 @@
             #expect(PostHogSessionReplayNetworkPlugin.isEnabledRemotely(remoteConfig: config) == false)
         }
 
+        @Test("Network plugin disabled when capturePerformance object has null web_vitals_allowed_metrics")
+        func networkPluginDisabledWhenNetworkTimingIsNull() {
+            let config: [String: Any?] = [
+                "capturePerformance": [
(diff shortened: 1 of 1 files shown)


PostHog/Replay/Plugins/Network/PostHogSessionReplayNetworkPlugin.swift
```diff
@@ -65,8 +65,14 @@
             if let isEnabled = capturePerformanceValue as? Bool {
                 return isEnabled
             }
-            // When enabled, capturePerformance is an object (we ignore the keys for now)
-            return true
+            // When enabled, capturePerformance is an object, check network_timing key
+            if let perfConfig = capturePerformanceValue as? [String: Any],
```
marandaneto, line 69:
```suggestion
            if let perfConfig = capturePerformanceValue as? [String: Any?],
```

i see the API returning: "web_vitals_allowed_metrics": null
so the value can be nullable, not sure if [String: Any] would work

marandaneto (approved):

PostHog/Replay/Plugins/Network/PostHogSessionReplayNetworkPlugin.swift
ioannisj: should be handled already through NSNull but added a test case for web_vitals_allowed_metrics: null

PR #476, opened 2026-02-17
fix: replace ReadWriteLock with NSLock for consistent thread-safety

Kilo-Loco:

The `@ReadWriteLock` property wrapper provides a false sense of thread-safety for collection types because the lock is acquired and released between separate operations, allowing the underlying collection to change between a check and a mutation.

Added concurrency tests covering concurrent adds, deletes, and mixed operations.

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 ## Next
 
+## 3.48.0
+
+### Minor Changes
+
+- 80f39ca: feat: add support for session replay event triggers
+
 ## 3.47.0
 
 ### Minor Changes
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -82,7 +82,7 @@ jobs:
 
       - name: Get GitHub App token
         id: releaser
-        uses: actions/create-github-app-token@v2
+        uses: actions/create-github-app-token@v3
         with:
           app-id: ${{ secrets.GH_APP_POSTHOG_IOS_RELEASER_APP_ID }}
           private-key: ${{ secrets.GH_APP_POSTHOG_IOS_RELEASER_PRIVATE_KEY }}
--- a/PostHog.podspec
+++ b/PostHog.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name             = "PostHog"
-  s.version          = "3.47.0"
+  s.version          = "3.48.0"
   s.summary          = "The hassle-free way to add posthog to your iOS app."
 
   s.description      = <<-DESC
--- a/PostHog.xcodeproj/project.pbxproj
+++ b/PostHog.xcodeproj/project.pbxproj
@@ -84,7 +84,6 @@
 		69ED1A9F2C8F451B00FE7A91 /* PostHogPersonProfiles.swift in Sources */ = {isa = PBXBuildFile; fileRef = 69ED1A9E2C8F451B00FE7A91 /* PostHogPersonProfiles.swift */; };
 		69ED1AB62C90711D00FE7A91 /* PostHogSDKPersonProfilesTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = 69ED1AB52C90711D00FE7A91 /* PostHogSDKPersonProfilesTest.swift */; };
 		69EE82BA2BA9C50400EB9542 /* PostHogReplayIntegration.swift in Sources */ = {isa = PBXBuildFile; fileRef = 69EE82B92BA9C50400EB9542 /* PostHogReplayIntegration.swift */; };
-		DA0SAMP22E2302260001FF01 /* PostHogSampling.swift in Sources */ = {isa = PBXBuildFile; fileRef = DA0SAMP12E2302260001FF01 /* PostHogSampling.swift */; };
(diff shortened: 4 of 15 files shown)


PostHog/PostHogFileBackedQueue.swift
```diff
@@ -60,7 +62,7 @@ class PostHogFileBackedQueue {
         do {
             let filename = "\(Date().timeIntervalSince1970)"
             try contents.write(to: queue.appendingPathComponent(filename))
-            items.append(filename)
```
marandaneto, line 63: pop method is only called when PostHogQueue acquires isFlushingLock
PostHogQueue calls PostHogFileBackedQueue add and delete in the same method
we also call peek/clear from PostHogQueue
maybe a single lock for all those operations within the PostHogQueue would be better?
i'll let @ioannisj figure this out
but i do see some possible race conditions, good catch @Kilo-Loco

ioannisj:
Hey @Kilo-Loco, thank you for this. I think what we need to do is remove the dependance on `@ReadWriteLock` property wrapper since it's indeed problematic and provides a false sense of protection against collection types specifically. So a simple NSLock around `items` here could potentially be safer and communicate intent better

Having said that, I can only see this as problematic since the rest of the operations should be protected by `@ReadWriteLock` already: 

```
if items.isEmpty { return }      // lock acquired, check, lock released
let removed = items.remove(at: index)  // lock acquired again, but items may have changed!
```

Kilo-Loco:
- [x] Removed `@ReadWriteLock` property wrapper dependency
- [x] Replaced with explicit `NSLock` for clearer intent
- [x] Merged main
- [x] Tests passing

Let me know if anything needs adjusting.

PostHog/PostHogContext.swift
```diff
@@ -268,9 +268,13 @@ class PostHogContext {
     func dynamicContext() -> [String: Any] {
         var properties: [String: Any] = [:]
 
-        if let screenSize {
-            properties["$screen_width"] = Float(screenSize.width)
-            properties["$screen_height"] = Float(screenSize.height)
+        screenSizeLock.lock()
+        let currentScreenSize = screenSize
+        screenSizeLock.unlock()
```
ioannisj: nit: prefer `.withLock` which has a generic return type. Commenting just here but it can be replaced in multiple places 

```suggestion
        let currentScreenSize = screenSizeLock.withLock { screenSize }
```

CHANGELOG.md
ioannisj, line 1: We switched to changeset based releases. So merge main and `pnpm changeset` + label this PR with `release` instead (details in [RELEASING.md](https://github.com/PostHog/posthog-ios/blob/1534f43a491e55c7db815755646acb6d39de35a5/RELEASING.md?plain=1#L1))

Kilo-Loco:
- [x] Updated call sites to use `.withLock` block
- [x] Add changeset file

Ready to be labeled with release if everything else looks good 👍🏽

PostHog/PostHogFileBackedQueue.swift
ioannisj: Pushed additional concurrency tests that expose a race condition in the queue. when multiple threads call add() simultaneously, they can generate identical filenames since `Date().timeIntervalSince1970` and file operations are not atomic. 

To clarify, this was always there and not added with this PR, but since we are working on this we should address it.

ioannisj: So one operation can overwrite the file of another add operation which can lead to data loss, if my rationale is correct here

PostHog/PostHogFileBackedQueue.swift
```diff
@@ -58,9 +64,9 @@ class PostHogFileBackedQueue {
 
     func add(_ contents: Data) {
         do {
-            let filename = "\(Date().timeIntervalSince1970)"
+            let filename = "\(Date().timeIntervalSince1970)-\(UUID.v7().uuidString)"
```
marandaneto: can we use only `(UUID.v7().uuidString` instead?
uuid v7 already has a timestamp in its string
you can also list files from a directory and sort by timestamp from its metadata (we do that on android btw)

ioannisj: The idea here was to be compatible when mixed with old file names. v7 is sortable which is very helpful, but I'll have to think how to handle the mix of UUIDv7 and old timestamps. I'll think about it

ioannisj: >you can also list files from a directory and sort by timestamp from its metadata (we do that on android btw)

yeah probably this is the way

marandaneto: i just mentioned above how we do this for android
this should be compatible with whatever name the files are
see example https://github.com/PostHog/posthog-android/blob/65700f9159a789861b35dac1f936f76bb0683807/posthog/src/main/java/com/posthog/internal/PostHogQueue.kt#L417-L422

marandaneto: timeIntervalSince1970-v7 is also fine, not a blocker, just not needed because the approach above would solve the sorting

ioannisj: Okay implemented by modification date [c33a357](https://github.com/PostHog/posthog-ios/pull/476/commits/c33a357b6ef61be6d71b49707f22eec9019f18eb) even though API is not as elegant as Android

PR #482, opened 2026-02-19
fix: push git tag before CocoaPods publish

marandaneto:
## Problem

The release workflow was failing at the CocoaPods publish step with:

```
fatal: Remote branch 3.41.2 not found in upstream origin
```

`pod trunk push` validates the podspec by cloning the repo from GitHub at the specified tag. The workflow was:

1. Create tag locally
2. Run `pod trunk push` ← fails because tag doesn't exist on remote yet
3. Push tag to GitHub

## Fix

Push the tag to GitHub **before** running `pod trunk push`, so CocoaPods can find it during validation.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -36,7 +36,7 @@ jobs:
           queries: security-and-quality
       - name: Build
         if: ${{ matrix.language == 'swift' }}
-        run: make buildSdk
+        run: make buildIOS
       - name: Perform CodeQL Analysis
         uses: github/codeql-action/analyze@v4
         with:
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -200,17 +200,16 @@ jobs:
...
+          git push origin "${{ steps.get-version.outputs.version }}"
 
       - name: Publish to CocoaPods
         run: make releaseCocoaPods
 
-      - name: Push tag to GitHub
-        run: |
-          git push origin "${{ steps.get-version.outputs.version }}"
-
       - name: Create GitHub release
```
ioannisj, line 210: Let's move this up with the tag, so that when cocoapods fails we are not left with a stale tag?

marandaneto: that is the problem, the tag has to exist before cocoapods is triggered, so if it fails, we'll have a stale tag, i dont see another way

ioannisj: Was talking about grouping tag + gh release before cocoapods

ioannisj: So at least we have a healthy SPM release, even if cocoapods fails

marandaneto: good point, fixed

Issue #478, opened 2026-02-19
Support for Enriched flag analytics


marandaneto:
### Description
follow up https://github.com/PostHog/posthog-android/issues/191
https://posthog.com/docs/libraries/js/usage#enriched-flag-analytics

- [x] Android
- [ ] iOS
- [ ] Flutter
- [ ] RN

marandaneto:
not all SDKs have setPersonPropertiesForFlags or similar, I think, have to double check

PR #487, opened 2026-02-20
Feat/support for Enriched flag analytics

jeremiahseun:
This PR adds support for **Enriched Flag Analytics**, resolving #478. These changes allow developers to track more granular user interactions with feature flags.

Specifically, it adds:
- `captureFeatureView(_:)`: Captures a `$feature_view` event for the specified flag.
- `captureFeatureInteraction(_:)`: Captures a `$feature_interaction` event and sets a person property (`$feature_interaction/[feature_flag] = true`) to track that the interaction has occurred.

These events enable deeper insights into flag performance and user engagement within PostHog.

I implemented a comprehensive test suite in [PostHogEnrichedAnalyticsTest.swift](cci:7://file:///Users/mac/Documents/GitHub/posthog-ios/PostHogTests/PostHogEnrichedAnalyticsTest.swift:0:0-0:0) covering:
- Correct event names and properties for both view and interaction calls.
- Verification of person property updates during interaction capture.
- Proper handling of opt-out and disabled SDK states.

During testing, I identified and fixed a scenario where lifecycle events (like `Application Installed`) interfered with batch counts. This was resolved by disabling `captureApplicationLifecycleEvents` in the test configuration, ensuring stable and deterministic test results.

## :pencil: Checklist

## If releasing new changes

### DIFF
--- a/PostHog/PostHogSDK.swift
+++ b/PostHog/PostHogSDK.swift
@@ -1402,8 +1402,10 @@ let maxRetryDelay = 30.0
     ///
     /// - Parameter flag: The key of the feature flag being interacted with.
     /// - Parameter flagVariant: The variant of the feature flag being interacted with.
-    @objc public func captureFeatureInteraction(flag: String,
-        flagVariant: String?) {
+    @objc public func captureFeatureInteraction(
+        flag: String,
+        flagVariant: String?
+    ) {
         if !isEnabled() {
             return
         }
--- a/.changeset/fuzzy-experts-teach.md
+++ b/.changeset/fuzzy-experts-teach.md
@@ -1,0 +1,5 @@
+---
+"posthog-ios": minor
+---
+
+feat: Add `captureFeatureView` and `captureFeatureInteraction` methods for tracking feature flag analytics


PostHog/PostHogSDK.swift
```diff
@@ -1282,6 +1282,40 @@ let maxRetryDelay = 30.0
         }
     }
 
+    /// Captures a $feature_view event for the specified feature flag.
```
marandaneto, line 1362: check the differences for the android PR https://github.com/PostHog/posthog-android/pull/416
there are a few missing things such as optional variant, reading the variant if not provided, etc

jeremiahseun: Done

PostHogTests/PostHogEnrichedAnalyticsTest.swift
```diff
@@ -0,0 +1,125 @@
+import Foundation
+import Nimble
+@testable import PostHog
+import Quick
+
+class PostHogEnrichedAnalyticsTest: QuickSpec {
```
ioannisj: @jeremiahseun mind rewriting this using swift testing framework? I'm in the process of migrating all tests away from quick and nimble

jeremiahseun: Alright. Got it. I'll do that

PostHog/PostHogSDK.swift
```diff
@@ -1282,6 +1282,73 @@ let maxRetryDelay = 30.0
...
+
+        if isOptOutState() {
+            return
+        }
+
+        var props: [String: Any] = [
+            "feature_flag": flag,
+        ]
+
+        let variant: Any = flagVariant ?? getFeatureFlag(flag, sendEvent: false) ?? true
```
ioannisj: Should we assume a variant value here? I think it would be better to guard at the top, log and return no-op. Same for captureFeatureView()

@marandaneto since you reviewed the Android implementation, wdyt?

```
// Get the variant value
guard let variant = flagVariant ?? getFeatureFlag(flag, sendEvent: false) else {
    // No variant available, don't capture the event and log
    hedgeLog("...")
    return
}
```

marandaneto: nope, if theres no variant that means its just a normal flag so defaults to true similar to android impl

marandaneto: actually, i think we shouldve called getFeatureFlagPayload instead
https://posthog.com/docs/libraries/js/usage#multivariate-feature-flags
i think android is wrong as well then?

marandaneto: unresolved because i am unsure
the goal is to default the variant if not provided
but it should not default to true if getFeatureFlag returns nil

ioannisj: Looks like we default to true here as well
https://github.com/PostHog/posthog-js/blob/755c2d7c73250e130e7975050d4e628fa1b811b3/packages/react/src/components/PostHogFeature.tsx#L83

and we get variant key with https://github.com/PostHog/posthog-js/blob/755c2d7c73250e130e7975050d4e628fa1b811b3/packages/react/src/hooks/useFeatureFlagVariantKey.ts#L4

marandaneto: which also calls https://github.com/PostHog/posthog-js/blob/755c2d7c73250e130e7975050d4e628fa1b811b3/packages/react/src/hooks/useFeatureFlagVariantKey.ts#L13
so the current implementation meets the JS SDK
the only issue is that if somebody calls captureInteraction(x)/captureFeature(x) and x does not exist, its still gonna be captured instead of no-op, not sure if this is the correct implementation

marandaneto: https://posthog.slack.com/archives/C07Q2U4BH4L/p1771664991308639

marandaneto: the flag should exist locally otherwise we should bail out
the suggestion @ioannisj mentioned would work https://github.com/PostHog/posthog-ios/pull/487#discussion_r2833241589
@jeremiahseun mind fixing this so we can get this merged?
we prob need to fix this on android as well

jeremiahseun: Alright then

PR #489, opened 2026-02-23
feat: support session recording sampleRate from remote config

marandaneto:

Add support for session recording `sampleRate` from remote config, matching the [Android SDK implementation](https://github.com/PostHog/posthog-android/pull/437).

This allows controlling what percentage of sessions get recorded via the PostHog project settings. Sampling is deterministic based on the session ID using the same `simpleHash`/`sampleOnProperty` algorithm as the JS SDK, ensuring consistent behavior across platforms.

The sample rate can also be configured locally via `PostHogSessionReplayConfig.sessionSampleRate` (defaults to 1.0 = record all), with the remote config value taking precedence when available.

See: https://posthog.com/docs/session-replay/how-to-control-which-sessions-you-record#sampling

- Added unit tests for `simpleHash` (consistency, positivity, distinctness, empty string)
- Added unit tests for `sampleOnProperty` (rate 0.0, 0.5, 1.0, clamping, determinism, approximate distribution)
- All 262 tests pass (`make test`)
- Lint passes (`make lint`)

## :pencil: Checklist

## If releasing new changes

### DIFF
--- a/PostHog/Replay/PostHogReplayIntegration.swift
+++ b/PostHog/Replay/PostHogReplayIntegration.swift
@@ -168,7 +168,7 @@
             // reset views when session id changes (or is cleared) so we can re-send new metadata (or full snapshot in the future)
             postHog.sessionManager.onSessionIdChanged = { [weak self] in
                 self?.resetViews()
-                self?.stopIfNotSampled()
+                self?.reevaluateSampling()
             }
 
             // flutter captures snapshots, so we don't need to capture them here
@@ -275,14 +275,19 @@
             return sampleOnProperty(sessionId, sampleRate)
         }
 
-        private func stopIfNotSampled() {
+        private func reevaluateSampling() {
             guard let postHog else { return }
 
             guard let sessionId = postHog.sessionManager.getSessionId(readOnly: true) else {
                 return
             }
 
-            if !shouldRecordSession(postHog: postHog, sessionId: sessionId) {
+            let sampled = shouldRecordSession(postHog: postHog, sessionId: sessionId)
+
+            if sampled, !isEnabled {
+                hedgeLog("[Session Replay] Session \(sessionId) sampled for recording. Starting.")
+                start()
+            } else if !sampled, isEnabled {
                 hedgeLog("[Session Replay] Session \(sessionId) not sampled for recording. Stopping.")
                 stop()
             }


PostHog/Replay/PostHogSampling.swift
```diff
@@ -0,0 +1,34 @@
...
+//  PostHog
+//
+//  Created on 23.02.26.
+//
+
+import Foundation
+
+/// Deterministic hash function matching the JS SDK's `simpleHash`.
+/// Uses 32-bit arithmetic to match JavaScript's bitwise operation behavior.
+func simpleHash(_ str: String) -> Int {
```
ioannisj, line 13: Did we test cross-sdk that this will produce the same hash?

marandaneto: that is the whole point -> Deterministic hash function
so as long as the impl is the same, it will be deterministic if using the same rate for the same session id

PostHog/Replay/PostHogSampling.swift
```diff
@@ -0,0 +1,34 @@
...
+    var hash: Int32 = 0
+    for scalar in str.unicodeScalars {
+        let charValue = Int32(truncatingIfNeeded: scalar.value)
+        hash = (hash &<< 5) &- hash &+ charValue
+    }
+    return abs(Int(hash))
+}
```
ioannisj, line 23:
```suggestion
/// Deterministic hash function matching the JS SDK's `simpleHash`.
/// Uses 32-bit arithmetic and UTF-16 encoding to match JavaScript's
/// `charCodeAt` behavior (which returns UTF-16 code units).
func simpleHash(_ str: String) -> Int {
    var hash: Int32 = 0
    for codeUnit in str.utf16 {
        let charValue = Int32(codeUnit)
        hash = (hash &<< 5) &- hash &+ charValue
    }
    return abs(Int(hash))
}
```

Had Claude compare and test this against https://github.com/PostHog/posthog-js/blob/00740eb7eae09a196644631062b54eb14940e261/packages/browser/src/extensions/sampling.ts#L12. Found it would mismatch with emojis due to UTF-16 vs unicode scalars. Emojis are not a thing but I would update just so the two implementations are identical throughout

marandaneto: good point, session ids are auto-generated by the SDKs, so its an UUID, wont have emojis, so the output was the same, but better to align, fixed that on android as well https://github.com/PostHog/posthog-android/pull/437/changes/101857848788dce8cb19d9639b2b7a3f60ed9eb0

marandaneto:
```
func simpleHash(_ str: String) -> Int {
    var hash: Int32 = 0

    for codeUnit in str.utf16 {
        let v = Int32(codeUnit)
        hash = (hash &<< 5) &- hash &+ v   // hash = hash * 31 + v, wrapping
    }

    // Match JS: return hash & 0x7fffffff
    return Int(hash & 0x7fffffff)
}
```

would be the 100% matching the JS impl tho because:

> If hash ends up as Int32.min (−2147483648):
> 
> Swift: abs(Int(hash)) can overflow or trap (depending on build/settings), and even when it doesn’t trap, it cannot produce a positive 32 bit value.
> 
> JS: Math.abs(-2147483648) becomes 2147483648, then if you later mask, it becomes 0.
> 
> So you want to avoid abs and instead apply the same mask as JS.

PostHog/PostHogRemoteConfig.swift
```diff
@@ -215,6 +216,10 @@ class PostHogRemoteConfig {
             if let endpoint = sessionReplay["endpoint"] as? String {
                 config.snapshotEndpoint = endpoint
             }
+
+            #if os(iOS)
```
marandaneto: needed otherwise tests fail since replay config is ios only

PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -156,10 +156,19 @@
...
+            {
+                hedgeLog("[Session Replay] Session \(sessionId) not sampled for recording. Skipping start.")
+                return
+            }
+
             isEnabled = true
             // reset views when session id changes (or is cleared) so we can re-send new metadata (or full snapshot in the future)
             postHog.sessionManager.onSessionIdChanged = { [weak self] in
                 self?.resetViews()
+                self?.stopIfNotSampled()
```
ioannisj: Wondering if we should react to sessionId changes to also start recording? 
Previews sessionId not sampled -> sessionId rotates and is now sampled, config.sessionReplay is enabled => recording should start. wdyt?

PostHog/PostHogRemoteConfig.swift
```diff
@@ -389,10 +394,39 @@ class PostHogRemoteConfig {
...
+                if let value {
+                    hedgeLog("Remote config sampleRate must be between 0.0 and 1.0, got \(value). Ignoring.")
+                }
+                return nil
+            }
+            return value
+        }
+
+        func getRecordingSampleRate() -> Double? {
+            recordingSampleRate
```
ioannisj: we should probably synchronize reading/writing `recordingSampleRate`. We write in a api.remoteConfig callback which would be a background thread most likely, and it seems we can read from any thread as well

ioannisj (approved):
LG, left a couple of comments

PostHog/PostHogRemoteConfig.swift
marandaneto: sessionReplayFlagActive has the same issue, will add a lock for both of them
although the chances are minimal since we read during app start (main thread) and after remote config is called, right now the approach is similar to @volatile on android

PostHog/Replay/PostHogReplayIntegration.swift
marandaneto: very good point, fixed

PR #490, opened 2026-02-25
Fix: Libwebp symbol conflict with @shopify/react-native-skia

jeremiahseun:
`Session Replay causes app crash when used with @shopify/react-native-skia` [issue](https://github.com/PostHog/posthog-js/issues/3029) fixed since it's root cause is from the native side

I wrote a dead-simple macOS Swift Package executable that links both `posthog-ios` and `libwebp`, that achieves the exact same linker-level verification.
However I am expecting a test result from a real environment from @slshults who will provide feedback and whether this resolves the issue.

## :pencil: Checklist

## If releasing new changes


johannesparty:
Adding another report for this, since the closing comment asked for one.
                                                                                                                                                                                  
  **Environment:**                                                                                                                                                                
  - PostHog iOS SDK: 3.34.0 (via `posthog-react-native-session-replay` 1.2.3)                                                                                                     
  - React Native 0.81 / Expo SDK 54                                                                                                                                               
  - Collision with `libwebp` 1.5.0, pulled in via `SDWebImageWebPCoder` 0.14.6 (which `ExpoImageManipulator` depends on — not Skia, but same root cause)                          
                                                                                                                                                                                  
  **Linker warning on every build:**                                                                                                                                              
                                                                                                                                                                                  
```
  duplicate symbol '_VP8LPredictor11_C' in:                                                                                                                                       
      .../PostHog/libPostHog.a(lossless.o)                                                                                                                                        
      .../libwebp/libwebp.a(lossless.o)               
```                                                                                                                            
                                                                                                                                                                                  
  Build succeeds — linker silently picks one copy — so behavior is nondeterministic across configuration changes.                                                                 
                                                                                                                                                                                  
  **Root cause (confirmed by reading `Pods/PostHog/vendor/libwebp/`):**                                                                                                           
                                                                                                                                                                                  
  The `ph_` prefix rename is incomplete. 44 of 131 vendored files are prefixed, but several implementation files — including `lossless.c` — are not, and their C function bodies  
  still use the upstream names (`VP8LPredictor0_C` … `VP8LPredictor14_C`, etc.). Any app that links both this SDK and a public `libwebp` pod hits the collision, regardless of how
   the public `libwebp` got pulled in (Skia, SDWebImageWebPCoder, or direct).                                                                                                     
                                                                       
This PR targets exactly this — would be great to see it reopened.     
  
  Happy to test a build if that helps.

PR #496, opened 2026-03-03
feat: add support for session replay event triggers

ioannisj:

Adds support for event-based triggers that control when session recording starts. Recording will only begin when one of the specified trigger events is captured.

### Behavior

- When triggers are configured, the replay integration starts but _drops_ all snapshot events until a trigger is matched. 
- Plugins are started but _drop_ events while waiting for an event trigger
- Once a matching event is captured, recording activates for that session
- Each new session requires a new trigger activation
- Sampling currently take precedence over event triggers. If the session is not sampled, it will not be started, and therefore event triggers will not be checked

```
[PostHog] New session id created 019CB300-70E6-7A2A-B498-14DDF3A10712 (sessionStart)
[PostHog] [Session Replay] Event triggers configured. Dropping snapshots until any of these events are captured: ["start_replay_trigger_1", "start_replay_trigger_2"]
[PostHog] [Session Replay] Console logs plugin started
[PostHog] [Session Replay] Network telemetry plugin started
...
[PostHog] Queued event 'start_replay_trigger_1'. Depth: 3
[PostHog] [Session Replay] Event trigger matched: start_replay_trigger_1. Activating replay for session 019CB300-70E6-7A2A-B498-14DDF3A10712.
[PostHog] Queued event '$snapshot'. Depth: 1
[PostHog] Queued event '$snapshot'. Depth: 2
[PostHog] Queued event '$snapshot'. Depth: 3
...
[PostHog] Session replay recording started. Session id is 019CB303-33B0-7590-A012-D5B0515E4136
[PostHog] New session 045AD28A-7455-4D9D-A08A-AAA30156FC3A, waiting for event triggers
[PostHog] [Session Replay] New session id created 045AD28A-7455-4D9D-A08A-AAA30156FC3A (customSessionId)

```

### Why drop instead of not starting?
This prepares for future buffering functionality where we'll keep the last x snapshots in a buffer (mem or disk) and flush them when a trigger activates, providing some context window leading up to the triggered event, not just after. 

### Ingestion controls and manual start/stop

There's some inconsistency on how we handle manual start/stop on mobile vs browser sdk when it comes to ingestion controls. 

Currently, all ingestion controls take precedence during `startSessionRecording()` call

- **feature flags:** For manual start, if the flag is not enabled, recording won't start
- **sampling:** For manual start, if the session is not sampled recording won't start
- **event triggers:** For manual start, if remote config has event triggers, recording will wait for trigger activation. For manual stop, even triggers will not work for the current session

Web sdk, on the other hand, provides a way to pass override flags to [startSessionRecording()](https://posthog.com/docs/session-replay/how-to-control-which-sessions-you-record?tab=Web#programmatically-start-and-stop-recordings) call to bypass these controls.

This is the third ingestion control we are adding to the SDK, so it would make sense to address this now. However, this could complicate the implementation since we'd need to refactor the way we check flags mainly, so I'll leave it as an improvement in a future PR. Open to suggestions and pushback though. 

- Added unit tests
- Added an event trigger section to sample app and tested manually

## :pencil: Checklist

## If releasing new changes

### DIFF
--- a/PostHog/Surveys/PostHogSurveyIntegration.swift
+++ b/PostHog/Surveys/PostHogSurveyIntegration.swift
@@ -1021,6 +1021,10 @@
                 }
             }
 
+            func testOnEvent(event: PostHogEvent) {
+                onEvent(event: event)
+            }
+
             static func clearInstalls() {
                 integrationInstalledLock.withLock {
                     integrationInstalled = false
--- a/PostHogTests/PostHogSurveysTest.swift
+++ b/PostHogTests/PostHogSurveysTest.swift
@@ -600,7 +600,7 @@ enum PostHogSurveysTest {
             )
             integration.testSetEventsToSurveys(["purchase": [(surveyId: "survey-1", condition: condition)]])
 
-            integration.onEvent(event: "purchase", properties: ["amount": 150])
+            integration.testOnEvent(event: PostHogEvent(event: "purchase", distinctId: "test", properties: ["amount": 150], timestamp: Date()))
             #expect(integration.testIsEventActivated(surveyId: "survey-1") == true)
         }
 
@@ -609,7 +609,7 @@ enum PostHogSurveysTest {
             let condition = PostHogEventCondition(name: "purchase")
             integration.testSetEventsToSurveys(["purchase": [(surveyId: "survey-1", condition: condition)]])
 
-            integration.onEvent(event: "purchase", properties: [:])
+            integration.testOnEvent(event: PostHogEvent(event: "purchase", distinctId: "test", properties: [:], timestamp: Date()))
             #expect(integration.testIsEventActivated(surveyId: "survey-1") == true)
         }
 
@@ -624,7 +624,7 @@ enum PostHogSurveysTest {
             )
             integration.testSetEventsToSurveys(["purchase": [(surveyId: "survey-1", condition: condition)]])
 
-            integration.onEvent(event: "purchase", properties: ["amount": 150, "category": "clothing"])
+            integration.testOnEvent(event: PostHogEvent(event: "purchase", distinctId: "test", properties: ["amount": 150, "category": "clothing"], timestamp: Date()))
             #expect(integration.testIsEventActivated(surveyId: "survey-1") == false)
         }
     }


PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -164,11 +174,38 @@
...
+                // Check if this session has already been activated
+                return activatedSession != currentSessionId
+            }()
+
+            if shouldWaitForTriggers {
+                let triggers = eventTriggersLock.withLock {
+                    isWaitingForTriggerActivation = true
+                    return eventTriggers
+                } ?? []
+                hedgeLog("[Session Replay] Event triggers configured. Dropping snapshots until any of these events are captured: \(triggers)")
```
marandaneto: > This prepares for future buffering functionality where we'll keep the last x snapshots in a buffer (mem or disk) and flush them when a trigger activates, providing some context window leading up to the triggered event, not just after.

I read the description, but we still dont know if this will cause OOM issues or not, so i'd rather not do this here until we are sure about memory pressure/footprint

The other thing is that you are probably dropping the full and meta snapshots, so once this starts capturing (after the trigger), i believe the recording will be broken?

PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -35,8 +35,14 @@
         private var applicationForegroundedToken: RegistrationToken?
         private var viewLayoutToken: RegistrationToken?
         private var remoteConfigLoadedToken: RegistrationToken?
+        private var eventCapturedToken: RegistrationToken?
         private var installedPlugins: [PostHogSessionReplayPlugin] = []
 
+        private let eventTriggersLock = NSLock()
+        private var eventTriggers: [String]?
```
marandaneto, line 43: should this be a string? for example, if i want to trigger a $screen event but for a specific $screen_name, would that be possible? is that possible on the web?

PostHog/Replay/PostHogReplayIntegration.swift
marandaneto: right now we pay the performance price of taking the screenshot but not the memory cost of caching it, right?

PostHog/Replay/PostHogReplayIntegration.swift
ioannisj: From what I see in replay settings (and remote config), you can only add event names. So similar limitation to survey event triggers currently. We can't trigger on event properties

PostHog/Replay/PostHogReplayIntegration.swift
ioannisj: Yes that's correct. Okay fair point, I'll change so that integration does not start

PostHog/Replay/PostHogReplayIntegration.swift
marandaneto: > So similar limitation to survey event triggers currently. We can't trigger on event properties

afaik this is an implementation issue on our end, the surveys feat do allow to filter by event properties as well

marandaneto:
```
"eventTriggers": [
            "$pageview"
        ],
```

yeah looks lile you cant for session replay

ioannisj: Yes, true for surveys. For session replay, the control I see on the settings page is just for selecting events

The config payload is:

```json
"sessionRecording": {
    "canvasFps": 3,
    "canvasQuality": "0.4",
    "consoleLogRecordingEnabled": true,
    "endpoint": "/s/",
    "eventTriggers": [
      "start_replay_trigger_1",
      "start_replay_trigger_2"
    ],
    ...
```

And js [type](https://github.com/PostHog/posthog-js/blob/d13c850cbc9fde020bf348ae7f159e1cded0cd6a/packages/browser/src/types.ts#L225) is using string[]

PostHog/Replay/PostHogReplayIntegration.swift
ioannisj: @marandaneto in this case, how would you expect manual startSessionRecording() to behave? to be consistent, it should be a no-op if event triggers are in place (similar to how we handle sampling)

marandaneto: how does the web handle that? i'd match the behaviour

ioannisj: I think js removes trigger match hook on stopSessionRecording(), so event matching will not work once a recording is stopped. But I'll need to validate. On startSessionRecording(), if no overrides are provided, it will wait based on "triggerMatchType". Which I just realized we only support "triggerMatchType": "all" which was the default?

marandaneto: iirc if you call start session recording on the web, it will wait for matching the events
or you can override with posthog.startSessionRecording({ event_trigger: true }) or posthog.startSessionRecording(true) to start right away.
so i think it makes sense to add the override option to the start methods as well

marandaneto: yeah we dont have triggerMatchType yet, i think it makes sense having now since we support feat flag + sample rate and soon event triggers
this should have been added along sample rate, but i didnt see it at all, good point

marandaneto: or we can just document that we only support the `triggerMatchType=all` option for now when you have multiple ways of enabling/disabling and create another issue for adding triggerMatchType so we can get this out sooner

ioannisj: It will probably need some refactoring on how we handle session replay feature flag, but if we put the effort on supporting `triggerMatchType=any` I would also add the ingestion overrides when calling startSessionRecording() so we are even closer to web implementation. Then we'll have just min session duration + buffering to figure out.

I would personally lean towards adding event triggers to all sdks first, with current limitations. Then circle back for triggerMatchType + ingestion overrides

ioannisj: > this should have been added along sample rate, but i didnt see it at all, good point

yeap, completely missed this too while reviewing

marandaneto: > I would personally lean towards adding event triggers to all sdks first, with current limitations. Then circle back for triggerMatchType + ingestion overrides

fine by me
just make sure to document that calling start session recording will still respect all configurations (flags + sample rate + events) to match anyway

marandaneto: @ioannisj is this addressed and should i just approve now?

PostHog/Surveys/PostHogSurveyIntegration.swift
```diff
@@ -73,6 +74,11 @@
 
         func start() {
             #if os(iOS)
+                // Subscribe to event captures
+                eventCapturedToken = postHog?.onEventCaptured.subscribe { [weak self] event in
+                    self?.onEvent(event: event)
+                }
```
ioannisj, line 80: @PostHog/team-surveys fyi, minor refactor on event property filtering. Survey integration now uses onEventCapture hook from posthog instance. Had to adjust onEvent(event:) signature and tests

PostHog/Replay/PostHogReplayIntegration.swift
ioannisj: Refactored so that integration is not started at all. Logic now is: 

- integration start(), event triggers present skip start
- manual startSessionRecording() will *not* start the recording, waiting for an event trigger first 
- on an event trigger, store activated session id and start
- manual startSessionRecording() will now start the recording since session has been activated
- on session id changed, check for event triggers and stop integration if needed
- manual start will *not* start the recording, waiting for an event trigger for new session

PR #501, opened 2026-03-04
Fix/queue pending feature flags reload

### LINKED ISSUE #456: Feature flag experience continuity bug (iOS, Android, Flutter, likely React Native)
Ticket: https://posthoghelp.zendesk.com/agent/tickets/45321
 
> Users get re-bucketed after identify() despite "Persist flag across authentication" being enabled.
>
> There seems to be a race condition: when preloadFeatureFlags (default: true) is in flight and user calls identify(), the subsequent reloadFeatureFlags() is dropped instead of queued.
>
> iOS - [PostHogRemoteConfig.swift:266-271](https://github.com/PostHog/posthog-ios/blob/main/PostHog/PostHogRemoteConfig.swift#L266-L271)
> Android - [PostHogRemoteConfig.kt:341-344](https://github.com/PostHog/posthog-android/blob/main/posthog/src/main/java/com/posthog/internal/PostHogRemoteConfig.kt#L341-L344)
>
>Both have:
>
>``` 
>if loadingFeatureFlags {
>    return  // ← Dropped, never sends $anon_distinct_id
>}
>```
>
> First request completes without $anon_distinct_id → server never stores hash key override → user re-bucketed.
>
> A potential fix is to queue pending reload instead of dropping:
> 
> ``` 
> if loadingFeatureFlags {
>    pendingFlagsReload = true  // ← Queue it
>   return
>}
>// After request completes, check pendingFlagsReload and retry
>```

- [ ] iOS
- [ ] Android
- [ ] Flutter (bump min native versions when completed)
- [ ] React Native

ordehi:

When preloadFeatureFlags is enabled (the default) and identify() is called shortly after SDK init, the flag reload triggered by identify() races with the preload request. The NSLock.withLock guard in loadFeatureFlags() uses return inside the closure, which only exits the closure (not the enclosing function), so both requests fire concurrently. The anonymous-context response can then overwrite the correct identified-context response.

The same broken guard pattern also exists in reloadRemoteConfig().

This was fixed in posthog-android 3.31.0 (https://github.com/PostHog/posthog-android/pull/407). The iOS SDK remains unfixed.

Fixes #456

- Added 6 unit tests covering: guard prevents concurrent requests, pending request preserves correct identity context, latest pending replaces earlier ones (replaced callback fires with nil), callbacks fire for both initial and pending requests, single call has no pending queue, and reloadRemoteConfig guard works correctly
- All tests pass via make test filter=PostHogRemoteConfigTest (13/13)
- Added a debug-only "Test preload+identify race" button to the example app for manual verification. With debug = true, console shows two sequential /flags requests where the second uses the identified distinct_id

## :pencil: Checklist

## If releasing new changes

### DIFF
--- a/PostHog/PostHogRemoteConfig.swift
+++ b/PostHog/PostHogRemoteConfig.swift
@@ -268,15 +268,8 @@ class PostHogRemoteConfig {
         groups: [String: String],
         callback: @escaping ([String: Any]?) -> Void
     ) {
-        let alreadyLoading = loadingFeatureFlagsLock.withLock {
+        let (alreadyLoading, previousCallback): (Bool, (([String: Any]?) -> Void)?) = loadingFeatureFlagsLock.withLock {
             if self.loadingFeatureFlags {
-                return true
-            }
-            self.loadingFeatureFlags = true
-            return false
-        }
-        if alreadyLoading {
-            let previousCallback: (([String: Any]?) -> Void)? = loadingFeatureFlagsLock.withLock {
                 let prev = self.pendingFeatureFlagsRequest?.callback
                 self.pendingFeatureFlagsRequest = PendingFeatureFlagsRequest(
                     distinctId: distinctId,
@@ -284,8 +277,12 @@ class PostHogRemoteConfig {
                     groups: groups,
                     callback: callback
                 )
-                return prev
+                return (true, prev)
             }
+            self.loadingFeatureFlags = true
+            return (false, nil)
+        }
+        if alreadyLoading {
             let cached = featureFlagsLock.withLock { getCachedFeatureFlags() }
             previousCallback?(cached)
             return


PostHog/PostHogRemoteConfig.swift
```diff
@@ -263,11 +268,27 @@ class PostHogRemoteConfig {
...
+                    anonymousId: anonymousId,
+                    groups: groups,
+                    callback: callback
+                )
+                return prev
+            }
+            let cached = featureFlagsLock.withLock { getCachedFeatureFlags() }
+            previousCallback?(cached)
+            return
         }
```
dustinbyrne: The gap is quite narrow, but re-acquiring the lock here looks like it could introduce a race condition:

- Thread A acquires lock, sees `loadingFeatureFlags = true`, releases lock, `alreadyLoading = true`
- Thread B (the one that's loading) finishes, API callback runs, calls `notifyFeatureFlagsAndRelease` which atomically sets `loadingFeatureFlags = false` and reads `pendingFeatureFlagsRequest` as nil
- Thread A acquires lock, sets `pendingFeatureFlagsRequest`, and this request is now orphaned

```suggestion
        let (alreadyLoading, previousCallback): (Bool, (([String: Any]?) -> Void)?) = loadingFeatureFlagsLock.withLock {
            if self.loadingFeatureFlags {
                let prev = self.pendingFeatureFlagsRequest?.callback
                self.pendingFeatureFlagsRequest = PendingFeatureFlagsRequest(
                    distinctId: distinctId,
                    anonymousId: anonymousId,
                    groups: groups,
                    callback: callback
                )
                return (true, prev)
            }
            self.loadingFeatureFlags = true
            return (false, nil)
        }
        if alreadyLoading {
            let cached = featureFlagsLock.withLock { getCachedFeatureFlags() }
            previousCallback?(cached)
            return
        }
```

ordehi: Not letting me commit the suggestion, so will do this locally: "Applying suggestions on deleted lines is currently not supported."

PR #505, opened 2026-03-05
feat(surveys): support survey wait period

adboio:
## 💡 Motivation and Context

adds support for survey wait period on ios

why:

1. ios does not support survey wait period
2. we do now support survey schedule 'always'
3. given the ux of the new guided editor (which sets "Frequency" as a combination of schedule=always and wait period = N days), it's very easy for a customer to accidentally have surveys showing up repeatedly

closes https://github.com/PostHog/posthog-ios/issues/498

## 💚 How did you test it?

manual testing, unit tests

## :pencil: Checklist

## If releasing new changes

### DIFF
--- a/PostHog/Surveys/PostHogSurveyIntegration.swift
+++ b/PostHog/Surveys/PostHogSurveyIntegration.swift
@@ -114,6 +114,9 @@
                         let deviceTypeCheck = self.doesSurveyDeviceTypesMatch(survey: survey)
                         return deviceTypeCheck
                     }
+                    .filter { survey in // 3.5. wait period has passed
+                        self.hasWaitPeriodPassed(survey: survey)
+                    }
                     .filter { survey in // 4. and match linked flags
                         let allKeys: [String?] = [
                             [survey.linkedFlagKey],
@@ -323,6 +326,7 @@
             }
 
             storage?.setDictionary(forKey: .surveySeen, contents: seenKeys ?? [:])
+            setLastSeenSurveyDate(Date())
         }
 
         /// Returns survey seen list (and mem-cache from disk if needed)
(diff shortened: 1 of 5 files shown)


PostHog/Surveys/PostHogSurveyIntegration.swift
```diff
@@ -928,6 +955,14 @@
                 getNewResponseKey(for: questionId)
             }
 
+            func testSetLastSeenSurveyDate(_ date: Date) {
```
marandaneto: Why do we need those? I know its called by the tests but i think theres better ways to test this without having those

PostHog/Surveys/PostHogSurveyIntegration.swift
```diff
@@ -362,6 +366,29 @@
...
+        private func hasWaitPeriodPassed(survey: PostHogSurvey) -> Bool {
+            guard let waitPeriodInDays = survey.conditions?.seenSurveyWaitPeriodInDays else {
+                return true
+            }
+            guard let lastSeenDate = getLastSeenSurveyDate() else {
+                return true
+            }
+            let now = Date()
+            let diffSeconds = abs(now.timeIntervalSince(lastSeenDate))
+            let diffDays = Int(ceil(diffSeconds / (3600 * 24)))
```
marandaneto: Use a const for the 3600*24

PostHog/Surveys/PostHogSurveyIntegration.swift
```diff
@@ -114,6 +114,9 @@
                         let deviceTypeCheck = self.doesSurveyDeviceTypesMatch(survey: survey)
                         return deviceTypeCheck
                     }
+                    .filter { survey in // 3.5. wait period has passed
```
marandaneto, line 117: What does 3.5 mean here?

adboio: this is claude-ism

but it's just the "step" number 

above/below the filters are labeled as step 3 and step 4

PostHog/Surveys/PostHogSurveyIntegration.swift
adboio: goal was just to test the actual getter/setter fns but not really necessary

more important bit is testing getActiveMatchingSurveys which can just hit storage directly 

will update

PostHog/Surveys/PostHogSurveyIntegration.swift
```diff
@@ -362,6 +366,30 @@
...
+        /// Checks if the wait period has passed since the last seen survey date
+        private func hasWaitPeriodPassed(survey: PostHogSurvey) -> Bool {
+            guard let waitPeriodInDays = survey.conditions?.seenSurveyWaitPeriodInDays else {
+                return true
+            }
+            guard let lastSeenDate = getLastSeenSurveyDate() else {
+                return true
+            }
+            let now = Date()
+            let secondsPerDay: Double = 3600 * 24
```
marandaneto: Add global const otherwise every method calls has to calculate this.
Global const is calculated at compile time

adboio:
### Merge activity

* **Mar 5, 7:08 PM UTC**: @adboio merged this pull request with [Graphite](https://app.graphite.com/github/pr/PostHog/posthog-ios/505).

Issue #513, opened 2026-03-10
Capture $feature_flag_called when session replay is enabled behind a flag


marandaneto:
### Description

Relates to https://github.com/PostHog/posthog-android/issues/250

- [X] Android/Flutter
- [ ] iOS/Flutter
- [ ] RN
- [ ] Web

PR #514, opened 2026-03-10
feat: capture $feature_flag_called event for session replay linked flags

marandaneto:

Closes https://github.com/PostHog/posthog-ios/issues/513

When session replay is gated behind a linked feature flag, the SDK checks the flag value in `isRecordingActive()` but never reports it as "called". This means the `$feature_flag_called` event is missing for these flags, causing incomplete usage analytics on the dashboard.

This is the iOS equivalent of https://github.com/PostHog/posthog-android/pull/431.

- Verified the SDK builds successfully on iOS and macOS
- Ran the existing test suite (`make test`) — all tests pass (1 pre-existing unrelated failure)
- Verified formatting with `make format`

## :pencil: Checklist

## If releasing new changes

### DIFF
--- a/PostHog/PostHogRemoteConfig.swift
+++ b/PostHog/PostHogRemoteConfig.swift
@@ -276,7 +276,7 @@ class PostHogRemoteConfig {
         // we consider session recording is active
 
         // Report the feature flag as called so usage is tracked
-        if let flagKey, let flagValue {
+        if let flagKey, let flagValue, config.sendFeatureFlagEvent {
             featureFlagCalledCallback?(flagKey, flagValue)
         }
 
--- a/PostHogTests/PostHogRemoteConfigTest.swift
+++ b/PostHogTests/PostHogRemoteConfigTest.swift
@@ -699,6 +699,32 @@ enum PostHogRemoteConfigTest {
                 #expect(calledFlagValue as? String == "web")
             }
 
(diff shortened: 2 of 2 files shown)


PostHog/PostHogSDK.swift
```diff
@@ -1438,6 +1440,9 @@ let maxRetryDelay = 30.0
     }
 
     private func reportFeatureFlagCalled(flagKey: String, flagValue: Any?) {
+        if remoteConfig == nil {
```
MohitMandalia, line 1443: I hope this is fine as here previous behavior was to call capture even if remoteConfig was nil.

PostHog/PostHogSDK.swift
```diff
@@ -1438,6 +1440,9 @@ let maxRetryDelay = 30.0
...
+            return
+        }
         var shouldCapture = true
```
MohitMandalia, line 1446: shouldn't `shouldCapture ` depend on config.sendFeatureFlagEvent as well instead of directly assuming true?

update: I see its taken care of in isRecordingActive itself still check once

PostHog/PostHogSDK.swift
marandaneto: yes its fine, this should not be null or its a sdk bug

PR #520, opened 2026-03-10
feat(surveys): support survey popup delay

### LINKED ISSUE #499: Implement survey popup delay (surveyPopupDelaySeconds)
## 🚨 IMPORTANT

This issue is likely user-facing in the main PostHog app, see [`surveyVersionRequirements.ts`](https://github.com/PostHog/posthog/blob/master/frontend/src/scenes/surveys/surveyVersionRequirements.ts). If you delete or close this issue, be sure to update the version requirements list here.

## Summary

The posthog-ios SDK does not implement the survey popup delay (`surveyPopupDelaySeconds`). The field is parsed but never used.

## Current State

- The `surveyPopupDelaySeconds` field exists on `PostHogSurveyAppearance` model (`PostHog/Models/Surveys/PostHogSurveyAppearance.swift:35`) as `TimeInterval?`
- The `showNextSurvey()` method in `PostHogSurveyIntegration.swift` immediately renders the survey via `DispatchQueue.main.async` with no delay
- The field is correctly parsed from JSON (verified by test at `PostHogSurveysTest.swift:55`)

## Expected Behavior

When `survey.appearance.surveyPopupDelaySeconds` is set to a value > 0:
1. The survey should not be rendered immediately
2. Instead, it should be delayed by the specified number of seconds
3. Surveys should be sorted by their delay value (shorter delays first)

## Reference Implementation

See posthog-js browser: `packages/browser/src/extensions/surveys.tsx` — `_handlePopover()` method which uses `setTimeout` for the delay.

## Tracking

This is tracked in the survey SDK feature parity issue: https://github.com/PostHog/posthog/issues/45658

_This issue was generated by Claude using the `/survey-sdk-audit` skill._

adboio:

ios does not support survey popup delay

closes https://github.com/PostHog/posthog-ios/issues/499

unit test / manual testing

## :pencil: Checklist

## If releasing new changes


PostHogTests/PostHogSurveysTest.swift
```diff
@@ -1717,6 +1717,41 @@ enum PostHogSurveysTest {
...
+        func sortsByDelayAscending() {
+            let surveys = [
+                PostHogSurvey.testInstance(name: "delayed-10", appearance: makeAppearance(delay: 10)),
+                PostHogSurvey.testInstance(name: "no-delay", appearance: nil),
+                PostHogSurvey.testInstance(name: "delayed-3", appearance: makeAppearance(delay: 3)),
+                PostHogSurvey.testInstance(name: "delayed-0", appearance: makeAppearance(delay: 0)),
+            ]
+
+            let sorted = surveys.sorted {
+                ($0.appearance?.surveyPopupDelaySeconds ?? 0) < ($1.appearance?.surveyPopupDelaySeconds ?? 0)
```
ioannisj: nit: I think this could be flaky cause .sorted is not guaranteed to be stable for equal keys I think. Maybe add tie-breaker as well or remove the scenario where there are equal keys (no-delay and delayed-0)

Also, this just duplicates production code, so if we change sorting at some point, this will still pass. I think we just need to test what getActiveMatchingSurveys returns instead

PostHog/Surveys/PostHogSurveyIntegration.swift
```diff
@@ -284,20 +284,31 @@
...
-                            if let self {
-                                // render the survey
-                                self.postHog?.config.surveysConfig.surveysDelegate.renderSurvey(
-                                    survey.toDisplaySurvey(),
-                                    onSurveyShown: self.handleSurveyShown,
-                                    onSurveyResponse: self.handleSurveyResponse,
-                                    onSurveyClosed: self.handleSurveyClosed
-                                )
+                        let delay = survey.appearance?.surveyPopupDelaySeconds ?? 0
+                        let deadline: DispatchTime = delay > 0 ? .now() + delay : .now()
```
ioannisj: I think this may be  a bit unecessary and just adds noise. We could just do this? 

```suggestion
                        let delay = max(survey.appearance?.surveyPopupDelaySeconds ?? 0, 0)
                        let deadline: DispatchTime = .now() + delay
```

PostHog/Surveys/PostHogSurveyIntegration.swift
```diff
@@ -284,20 +284,31 @@
 
                 // Check if there is a new popover surveys to be displayed
                 getActiveMatchingSurveys { activeSurveys in
-                    if let survey = activeSurveys.first(where: self.canRenderSurvey) {
-                        // set survey as active
+                    // Sort by popup delay (shorter delays first)
+                    let sortedSurveys = activeSurveys.sorted {
+                        ($0.appearance?.surveyPopupDelaySeconds ?? 0) < ($1.appearance?.surveyPopupDelaySeconds ?? 0)
+                    }
```
ioannisj: Does web sdk behave similar to this? This feels like the wrong approach to me cause it changes the whole survey selection behavior, right? Now surveys with a smaller delay are given display priority, whereas before, selection was simply by `first(where:)` to a list returned by the server/config

I feel that the delay should actually be handled in the UI layer, in `PostHogSurveysDefaultDelegate` where the delegate will delay the presentation of a survey (and handle the dismissal scenario before delay like we do in L302) _after_ the survey has been selected by getActiveMatchingSurveys()

PostHogTests/PostHogSurveysTest.swift
ioannisj: On first point above, actually it seems that it's guaranteed, so never mind 

https://developer.apple.com/documentation/swift/array/sorted()
> The sorting algorithm is guaranteed to be stable. A stable sort preserves the relative order of elements that compare as equal.

PostHog/Surveys/PostHogSurveyIntegration.swift
ioannisj: Yes, looks like js SDK is performing this delay logic in display layer in [handlePopoverSurvey](https://github.com/PostHog/posthog-js/blob/64363a53f41e31b8487dce546492149efa1330a2/packages/browser/src/extensions/surveys.tsx#L182)

ioannisj:
@PostHog/team-client-libraries pushed some changes if you can please review?

marandaneto (approved):

Issue #321, opened 2025-03-23
PostHog's Session Replay feature degrades Google services performance


katcinskiy:
### Description

I noticed that Google Street View and Google Maps are lagging when PostHog's session replay is active. Every time a screenshot is captured, there's a noticeable drop in frame rate that decreases the overall smoothness of Street View navigation.

I've observed that the smaller the config.sessionReplayConfig.debouncerDelay value, the less smooth the panorama becomes, so probably this is the root cause. 

Here's my configuration:

`
let config = PostHogConfig(apiKey: "", host: "https://eu.i.posthog.com")

config.captureScreenViews = false
config.captureApplicationLifecycleEvents = true
config.sessionReplay = true
config.sessionReplayConfig.screenshotMode = true
config.sessionReplayConfig.captureNetworkTelemetry = true
config.sessionReplayConfig.maskAllImages = false
config.sessionReplayConfig.maskAllTextInputs = false
config.sessionReplayConfig.debouncerDelay = 1.0
PostHogSDK.shared.setup(config)
`

I didn't experience it with SmartLook session replay.

katcinskiy:
@ioannisj sorry for the delay. Sure I will provide you example with panorama view. Please replace `<gms_key>` and `<posthog_api_key>`

[google-streetview-test.zip](https://github.com/user-attachments/files/19929817/google-streetview-test.zip)

Here is the screencasts WITH post hog and WITHOUT, in PostHog version you can see some glitches:

https://github.com/user-attachments/assets/a95de077-dbf0-4a6b-811b-d07384d8ed15

https://github.com/user-attachments/assets/df5d69f1-8b30-4d7b-90c8-71dcb3584bdf

finder39:
@ioannisj this is on an iPhone 16 Pro running 18.4.1

finder39:
@ioannisj we are running the latest PostHog version, yes. I'd be happy to test out a version if you sent it off to me.

finder39:
@ioannisj that did seem to help. This test was on an iPhone 12 Pro running 18.3.1, so on older hardware as well, but different OS version too. I can test back on the original hardware/OS if you'd like.

A couple things I noticed:
1. It seems to be doing generation of the screenshot off the main thread now, which is good. Not sure if this is an actual change or if I was looking wrong before.
2. It seems to take less resources and be less stuttery

Options used:
```
    config.captureApplicationLifecycleEvents = true
    
    config.sessionReplay = true
[...]

PR #536, opened 2026-03-25
perf(replay): optimize screenshot mode — 93% faster (14.7x)

marandaneto:

Relates to https://github.com/PostHog/posthog-ios/issues/321

Session replay's screenshot capture was consuming **3,319µs per snapshot on the main thread** — nearly 20% of a 60fps frame budget. Users on older devices reported frame drops and jank caused by the view hierarchy traversal and image rendering.

This PR reduces main-thread time to **~230µs (93.2% reduction, 14.7× faster)**, bringing frame budget impact down to **1.4% at 60fps** — negligible.

### Performance Results

| Metric | Before | After | Improvement |
|--------|--------|-------|-------------|
| **Total** | 3,319µs | ~230µs | **−93.2%** |
| findMaskableWidgets | 1,760µs | ~175µs | −90.0% |
| toImage (drawHierarchy) | 653µs | ~50µs | −92.3% |
| Frame budget (60fps) | 19.9% | **1.4%** | — |

### Key Optimizations

**Main-thread performance:**
1. **Plain UIView + UIWindow fast path** — Skip all type casts for container views that can never contain sensitive content. Biggest win (~66%).
2. **Custom CGContext renderer** (`PostHogGraphicsImageRenderer`) — Based on [Sentry's approach](https://blog.sentry.io/boosting-session-replay-performance-on-ios-with-view-renderer-v2/). Bypasses `UIGraphicsImageRenderer` overhead by directly allocating a CGContext with `malloc`. Uses native screen scale which avoids internal rescaling in `drawHierarchy`.
3. **Config caching** (`ReplayMaskConfig`) — Cache `maskAllTextInputs`, `maskAllImages`, `maskAllSandboxedViews` once per snapshot instead of traversing weak ref + optional chain per view.
4. **If-else type chains** — Skip unnecessary type casts after a match in both `findMaskableWidgets` and `toWireframe`.
5. **Touch phase filtering** — Skip `location(in:)` for non-began/ended phases (avoids coordinate conversion during drag/scroll).
6. **Move lock + meta event off main thread** — `windowViewsLock` access and meta event dict creation moved to dispatch queue.

**Off-main-thread + memory:**
- Skip `maskImage` re-render when `maskableWidgets` is empty (saves ~1.3MB)
- Release `UIImage` early in `toDict()` before base64 encoding (reduces peak memory)
- Pre-size dictionaries in `toDict()` (`minimumCapacity: 16`)
- Hex lookup table in `toRGBString()` — replaces `String(format:)` with direct char array
- `malloc` instead of `calloc` for renderer buffer (skip zero-init of ~12MB at 3x scale)

### Files Changed
- `PostHogReplayIntegration.swift` — Fast paths, config caching, if-else chains, touch handler, async refactor
- `PostHogGraphicsImageRenderer.swift` — **NEW** custom CGContext renderer
- `UIView+Util.swift` — Custom renderer integration, `isNoCapture()` optimization
- `CGColor+Util.swift` — Hex lookup table for color string formatting
- `RRWireframe.swift` — `maskImage` skip, early image release, dict pre-sizing
- `RRStyle.swift` — Dict pre-sizing

- Benchmark test on iOS Simulator (iPhone 17 Pro) with 200-view synthetic hierarchy (484 actual views including UIKit internals)
- 80 experiments across 9 optimization sessions with automated measurement
- `make buildIOS` and `swift build` both pass
- `make format` clean (SwiftLint + SwiftFormat)

## :pencil: Checklist

### DIFF
--- a/PostHog.xcodeproj/project.pbxproj
+++ b/PostHog.xcodeproj/project.pbxproj
@@ -188,6 +188,7 @@
 		DA4AF6292D119FCD0053EA38 /* PostHog.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 3AC745B5296D6FE60025C109 /* PostHog.framework */; };
 		DA4AF62A2D119FCD0053EA38 /* PostHog.framework in Embed Frameworks */ = {isa = PBXBuildFile; fileRef = 3AC745B5296D6FE60025C109 /* PostHog.framework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
 		DA4FFB152DA93C78006BAEEA /* PostHogSessionReplayTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = DA4FFB142DA93C78006BAEEA /* PostHogSessionReplayTest.swift */; };
+		ARPERF0004000000000004 /* PostHogGraphicsImageRenderer.swift in Sources */ = {isa = PBXBuildFile; fileRef = ARPERF0003000000000003 /* PostHogGraphicsImageRenderer.swift */; };
 		DA4FFBB52DAD5B01006BAEEA /* PostHogIdentityTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = DA4FFBB42DAD5AF9006BAEEA /* PostHogIdentityTests.swift */; };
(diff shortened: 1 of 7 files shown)


ioannisj:
> @ioannisj autoresearch result, we'd need to test everything but the results are good

Oh wow! Will take a look and test. Looks promising from PR description

PR #545, opened 2026-03-31
feat(workflows): add push subscription functionality

havenbarnes:

This will need to be merged after the posthog PR stack https://github.com/PostHog/posthog/pull/53276 and documentation updates: [<pr here>](https://github.com/PostHog/posthog.com/pull/16177)

https://github.com/PostHog/posthog/issues/44444

We're adding push notification channel to workflows, and SDKs need to send us their device tokens to fulfill the push subscription on our end

Tested everything end to end locally, hitting APNS sandbox on physical device

image

Added tests for new logic

## :pencil: Checklist

- [x] I updated the docs if needed https://github.com/PostHog/posthog.com/pull/16177

## If releasing new changes


PostHog/PostHogSDK.swift
```diff
@@ -2075,6 +2083,95 @@ let maxRetryDelay = 30.0
...
+        /// When the user grants permission, the SDK automatically registers for remote notifications.
+        /// After receiving the device token from the system, call ``handlePushNotificationDeviceToken(_:)``
+        /// from your `AppDelegate`'s `application(_:didRegisterForRemoteNotificationsWithDeviceToken:)` method
+        /// to send the token to PostHog.
+        ///
+        /// - Parameter options: The notification authorization options to request. Defaults to `[.alert, .sound, .badge]`.
+        /// - Parameter completion: A closure called with the result of the authorization request.
+        ///   The `Bool` indicates whether permission was granted, and the `Error` contains any error that occurred.
+        @available(iOS 14.0, macOS 11.0, *)
+        public func requestPushNotificationPermission(
```
ioannisj: Not sure if requesting push notifications should belong to the SDK's scope tbh. Feels like notification permission flow is very nuanced and there are other libs that deal with this better. How and when a request for notifications should be made is host app responsibility, we just need to intercept or accept the notification token for registration?

PostHog/PostHogSDK.swift
```diff
@@ -2075,6 +2083,95 @@ let maxRetryDelay = 30.0
...
+    @objc public func handlePushNotificationDeviceToken(_ deviceToken: Data) {
+        if !isEnabled() {
+            return
+        }
+
+        let tokenString = deviceToken.map { String(format: "%02x", $0) }.joined()
+        sendPushNotificationDeviceToken(tokenString)
+    }
+
+    private func sendPushNotificationDeviceToken(_ deviceToken: String) {
```
ioannisj: This fires a one-shot network request, with no retry or persistence. For offline situations the token will never be registered. Should we go through normal event capture here which already has all these mechanisms in place?

PostHog/PostHogConfig.swift
```diff
@@ -72,6 +72,18 @@ public typealias BeforeSendBlock = (PostHogEvent) -> PostHogEvent?
...
+    #if os(iOS) || os(macOS)
+        /// Automatically capture push notification engagement events when users interact with notifications.
+        ///
+        /// When enabled, the SDK will swizzle `UNUserNotificationCenterDelegate` methods to automatically
+        /// capture a "Push Notification Opened" event whenever a user taps a push notification.
+        ///
+        /// Requires `enableSwizzling` to be `true`.
+        ///
+        /// Default: true
+        @objc public var capturePushNotificationSubscriptions: Bool = true
```
ioannisj: This also silently handles registering push notification token through app delegate swizzling. Should we split these into two different config options?

I also feel we should give users who opt out of this config a manual way to handle this by adding public APIs so they can just forward delegate calls to this SDK (from UNUserNotificationCenterDelegate events and UIApplicationDelegate)

PostHog/PushNotifications/PostHogPushNotificationIntegration.swift
```diff
@@ -0,0 +1,426 @@
...
+            }
+
+            method_exchangeImplementations(originalMethod, swizzledMethod)
+
+            Self.swizzledDelegateClassesLock.withLock {
+                Self.swizzledDelegateClasses.removeAll()
+            }
+        }
+
+        // MARK: - UNUserNotificationCenterDelegate Method Swizzling
```
ioannisj, line 105: We already have some methods for swizzling in the SDK that we could reuse or extend instead?

ioannisj (commented):
A few comments from a first pass

PostHog/PostHogSDK.swift
havenbarnes: > How and when a request for notifications should be made is host app responsibility

I agree - this is totally optional to use but felt like a nice convenience. We are intercepting the token registration regardless of how they request for notif permissions. Do you think we should just remove this entirely instead?

PostHog/PostHogSDK.swift
havenbarnes: Yeah that's fair. AFAIK, normal event capture is not an option because we should encrypt these device tokens

havenbarnes: Ok I added offline retry/persistence for this, was a pretty small change

PostHog/PostHogConfig.swift
havenbarnes: > This also silently handles registering push notification token through app delegate swizzling. Should we split these into two different config options?

You know best here - sounds like there should be an explicit / non-swizzled version? I only opted for this as it felt the most frictionless and I noticed there were other swizzling-required SDK features.

> I also feel we should give users who opt out of this config a manual way to handle this by adding public APIs so they can just forward delegate calls to this SDK (from UNUserNotificationCenterDelegate events and UIApplicationDelegate)

Certainly a good idea. I can add this on our side

PostHog/PushNotifications/PostHogPushNotificationIntegration.swift
havenbarnes: Ok I refactored this a bit to reuse PostHogSwizzler and it did require a little extension for add-or-exchange swizzling

PostHog/PostHogSDK.swift
ioannisj: Yes, I would drop it since I feel this should not be a responsibility for this sdk (yet?)

PostHog/PostHogConfig.swift
ioannisj: >You know best here - sounds like there should be an explicit / non-swizzled version? I only opted for this as it felt the most frictionless and I noticed there were other swizzling-required SDK features.

I would have a non-swizzled path as well and maybe even flip the default to `false` here. App delegate is the entry point for apps, and if something goes wrong with swizzling here (esp with custom classes) we'd be breaking the whole app. @turnipdabeets wdyt?

ioannisj (commented):
We recently refactored PostHogIntegration protocol, so probably that's why CI is failing, should be a quick fix. Todo here I think is exposing manual methods for token registration and have token autodetection with swizzling as an opt-in?

PostHog/PostHogSDK.swift
```diff
@@ -2120,6 +2127,124 @@ let maxRetryDelay = 30.0
...
+        api.pushSubscription(distinctId: distinctId, deviceToken: deviceToken, appId: appId) { [weak self] success in
+            if success {
+                hedgeLog("Sent push subscription to PostHog (retry).")
+                self?.storage?.remove(key: .pushSubscription)
+            } else {
+                hedgeLog("Retry of push subscription failed. Will retry on next flush.")
+            }
+        }
+    }
 }
```
marandaneto: lets move this to a dedicated class that gets all the dependencies as ctor eg storage, api, etc like we do for the other stuff keeping this PostHogSDK cleaner

PostHog/PostHogSDK.swift
```diff
@@ -2120,6 +2127,124 @@ let maxRetryDelay = 30.0
...
+            }
+        }
+    #endif
+
+    /// Sends the device token to PostHog for push notification delivery.
+    /// Push notifications can be configured using PostHog Workflows to target users based on their behavior and properties.
+    /// When a push notification is sent from PostHog, the device token is used to deliver the notification to the correct device.
+    ///
+    /// - Parameter deviceToken: The device token received from the system.
+    @objc public func handlePushNotificationDeviceToken(_ deviceToken: Data) {
```
marandaneto: is the deviceToken really a `Data`? Shouldn't it be a String or something? where do people read this from?

PostHog/PostHogConfig.swift
```diff
@@ -283,6 +295,14 @@ public typealias BeforeSendBlock = (PostHogEvent) -> PostHogEvent?
             }
         #endif
 
+        #if os(iOS) || os(macOS)
+            if #available(iOS 14.0, macOS 11.0, *) {
+                if capturePushNotificationSubscriptions {
```
marandaneto, line 351: i think we should also guard this against enableSwizzling right

PostHog/PostHogConfig.swift
turnipdabeets: Agreed, if we offer this it should be opt-in and maybe 2 separate versions/configs for users who can't swizzle.

PostHog/PostHogConfig.swift
ioannisj: PostHogSDK checks `integration.requiresSwizzling` and skips installation with a log, but maybe this was a better place for that check. Either way, should be guarded by requiresSwizzling already

PostHog/PostHogSDK.swift
havenbarnes: Depends where it comes from, if they use the swizzled codepath it will be Data. I'll converge this to string to make it more straightforward

PostHog/PostHogConfig.swift
havenbarnes: Currently it's only filtered at install time in PostHogSDK.swift:2039 via requiresSwizzling, but I'll add another guard here to be safe yeah

PostHog/PostHogSDK.swift
marandaneto: yeah then the swizzling path should do the casting
the public API is gonna be user-friendlier with a String instead

PostHog/PostHogConfig.swift
marandaneto: i'd lean to make it as automagically as possible but also give a manual API (opt out by default)
thes less users have to do, the more automatic stuff works the better
swizzle is dangerous, but if done right its super useful
session replay and autocapture only works with swizzle and people use it just fine, except a few folks that prefer to disable everything, we should just respect the config if people disable this config or disable swizzle for everything

PostHog/PostHogConfig.swift
marandaneto: its done here its good https://github.com/PostHog/posthog-ios/blob/c165ead50ecf66368685a8467fa625fc0960bace/PostHog/PostHogSDK.swift#L2032-L2033
no need to dupe the check, i just missed that

PostHog/PostHogSDK.swift
ioannisj: Not sure I completely agree here since the delegate method is called with [Data](https://developer.apple.com/documentation/uikit/uiapplicationdelegate/application(_:didregisterforremotenotificationswithdevicetoken:)). So simplest api would be people just forward this call to the sdk?

marandaneto: https://firebase.google.com/docs/cloud-messaging/ios/get-started#disable-swizzling
i think it depends on what we are collecting here, APN tokens or fcmToken (firebase cloud messaging)?
i understood that `Data` [was only for the swizzling use case](https://github.com/PostHog/posthog-ios/pull/545#discussion_r3125495912), so its automatic, type does not matter
the manual API should have the type that the provider is using, if that is Data, then Data it should be

https://github.com/firebase/firebase-ios-sdk/blob/90c17718b5126b6d95e24c019f10a9fc6a62c54f/FirebaseMessaging/Sources/FIRMessaging.m#L427-L429 here is Data
the firebase link has fcmToken as String

so we just need to figure out what we need and what is its type and that should be the type in the public API

PostHog/PostHogConfig.swift
ioannisj: I'm okay with that. Only reason I pushed back a bit here is that this is not using the regular swizzling path which we know works and used by other integrations, but we now handle and need to worry about optional methods from the concrete delegate class (with swizzleAddingIfNeeded())

marandaneto: yep fair point, so it has to be tested that it does not break, whats the common use cases for swizzling breaking apps?
iOS version? OS (eg ipad vs ios vs macos)? runtime config or something?
figure out and test those use cases
i also think that its quite common to swizzle this code path to make notifications setup automatic, i've seen a few examples on the internet, so i think its common and it should be safe enough, even firebase mentions swizzle a few times which is the main notifications provider for apple if not using APN directly

ioannisj: We had a similar discussion when deeplinks were implemented on UIKit. I'm okay testing this out as true by default, then we can come back and make deeplinks automagic as well since swizzling app delegate will be there

PostHog/PostHogSDK.swift
ioannisj: Ah good point on fcm vs apns tokens. Logic says that we'll need a way to distinguish BE side on this since delivery method will differ? So probably two different public api methods I would say. Swizzling here seems to cover the apns token. For FCM I assume we'll need to stick to manual wiring. Either way, we'd need to know if this is FCM or APNS when delivering the token

havenbarnes: AFAIK, the device tokens format should be the same regardless of delivery platform; FCM ultimately delivers via APNS anyway.

We're not bringing in any FCM-related logic into the SDK - all of the platform differences for delivery happen [on the backend here](https://github.com/PostHog/posthog/pull/51554/changes#diff-a8467d2cc3ef2aafffc39e939931fb2087dd0888a526a671222d2c0ee92a8faeR61-R64)

So I don't think we need to care about anything on the SDK side other than delivering this device token

PostHog/PostHogSDK.swift
havenbarnes: done

PostHog/PostHogSDK.swift
havenbarnes: done

PostHog/PostHogConfig.swift
havenbarnes: Ok, following this thread and (i think) the consensus i've made two changes:
- Set `capturePushNotificationSubscriptions: Bool = false` to default, will make sure this is called out in my documentation PR
- Added a non-swizzling codepath for setting up push subscriptions.

LMK if I misinterpreted or anything else is needed here

PostHog/PostHogSDK.swift
marandaneto: the question was more about the datatype really

PostHog/PostHogSDK.swift
```diff
@@ -2120,6 +2132,86 @@ let maxRetryDelay = 30.0
...
+
+    /// Sends the device token to PostHog for push notification delivery.
+    /// Push notifications can be configured using PostHog Workflows to target users based on their behavior and properties.
+    /// When a push notification is sent from PostHog, the device token is used to deliver the notification to the correct device.
+    ///
+    /// - Parameter deviceToken: The device token as a hexadecimal string. If you receive a `Data` token from
+    ///   `application(_:didRegisterForRemoteNotificationsWithDeviceToken:)`, convert it first via
+    ///   `token.map { String(format: "%02x", $0) }.joined()`.
+    @objc public func handlePushNotificationDeviceToken(_ deviceToken: String) {
+        if !isEnabled() {
```
marandaneto, line 2145: do we want to guard against `isOptOut` as well?

marandaneto (approved):

PostHog/PostHogSDK.swift
havenbarnes: Yep added here and in capturePushNotificationOpened

ioannisj (approved):

turnipdabeets (approved):

marandaneto:
@ioannisj @turnipdabeets i think this is done, but just need to test/fix conflicts
i think @havenbarnes already left so should we pick this up? can you check with his former team? also for android btw

ioannisj:
Need to rebase this since they are unsigned commits from main (before signing requirements were added probably) that I have to get rid of. Turns out I'm refactoring this quite a bit so it'll beed a fresh review anw

ioannisj:
Rebasing did not work since this tree already had unverified commits from a previous main merge, and I could not push or force-push even.  Closing in favor of https://github.com/PostHog/posthog-ios/pull/608

PR #548, opened 2026-04-03
feat: session replay minimum recording duration

ioannisj:

Adds support for the `minimumDurationMilliseconds` remote config setting for session replay. When configured, session replay snapshots are buffered locally until the session reaches the minimum duration threshold, at which point they are migrated to the replay queue and sent to PostHog. Sessions that don't meet the minimum duration are discarded on next session start

This implementation is similar to Web SDK's [strict mode](https://posthog.com/docs/session-replay/how-to-control-which-sessions-you-record#strict-mode-available-in-12910) but adapted for mobile:

- Duration is measured from buffered recording data (first to last snapshot timestamp), not session age. 
- The buffer is not reset on navigation. On mobile apps, screen transitions are far more frequent than web page refreshes, so clearing the buffer on every navigation would make the feature impractical imho.

### Implementation overview

- Snapshots are written to a temporary disk-based buffer instead of the main replay queue
- The buffer tracks duration using timestamps in UUID v7 filenames
- ~A sliding window prunes snapshots older than the minimum duration threshold~
- Once the time span between oldest and newest buffered snapshot meets the threshold, all buffered snapshots are migrated to the replay queue for upload (moved)
- Buffer is always cleared on new session start. So If the session ends before meeting the threshold, the buffer is discarded on next session
- Once the minimum duration threshold is met for a session, subsequent snapshots go directly to the replay queue (no more buffering for that session).
- When minimum duration is not configured or is zero, snapshots go directly to the replay queue (current behavior)
- Flushing will be ignored while replay queue is writing to buffer

### Performance considerations

- There should be no additional memory footprint for this feature, since what really changes is _where_ we write snapshots on disk. No memory buffer involved

### Other considerations

- Buffer length, used to decide when to drain, is calculated from the first to last snapshot timestamps. This ignores user inactivity. For example, if we snapshot at launch, the user is idle for ~29s, then interacts for 1s, the minimum session duration check still passes and we record the replay. Ideally, we’d require ~30s of actual user activity before capturing a session.

- Manual testing on sample projects mainly
- Added a few tests to unit test queue changes

## :pencil: Checklist

## If releasing new changes


PostHog/Replay/PostHogReplayBufferQueue.swift
```diff
@@ -0,0 +1,147 @@
...
+        // they didn't meet the minimum duration threshold and should be discarded
+        deleteSafely(queue)
+
+        do {
+            try FileManager.default.createDirectory(atPath: queue.path, withIntermediateDirectories: true)
+        } catch {
+            hedgeLog("Error trying to create replay buffer folder \(error)")
+        }
+
+        items = []
```
dustinbyrne: Should this be locked?
```suggestion
        itemsLock.withLock { items = [] }
```

dustinbyrne (approved):
Looks good, one minor comment.

ioannisj:
> > A sliding window prunes snapshots older than the minimum duration threshold
> 
> how do we know that we are not dropping important snapshots? eg a recording without a metadata snapshot is a broken/not playable recording so i am not sure this is the right approach, most likely we want either all or nothing here

That's actually a very good and valid point. Did not think about that, but meta is the first snapshot type we capture, so essentially if any pruning would happen it would destroy the recording you are right. Anyway buffer size was set to min session duration, so in the normal use case it would never prune. Just added it as a safety mechanism for disk usage. 

I'll remove and keep it all or nothing like you suggested

marandaneto:
@ioannisj is this ready for review after the changes mentioned [here](https://github.com/PostHog/posthog-ios/pull/548#issuecomment-4201912053)?

ioannisj:
> @ioannisj is this ready for review after the changes mentioned [here](https://github.com/PostHog/posthog-ios/pull/548#issuecomment-4201912053)?

Yes 🙏 I've run some manual tests and it seems to be behaving as expected

PostHog/Replay/PostHogReplayBufferQueue.swift
```diff
@@ -0,0 +1,109 @@
...
+    /// Returns seconds since epoch as `TimeInterval`, or nil if parsing fails.
+    static func timestampFromUUIDv7(_ uuidString: String) -> TimeInterval? {
+        // UUID v7 format: XXXXXXXX-XXXX-7XXX-XXXX-XXXXXXXXXXXX
+        // First 48 bits (12 hex chars across first two groups) = milliseconds since epoch
+        let cleaned = uuidString.replacingOccurrences(of: "-", with: "")
+        guard cleaned.count >= 12 else { return nil }
+        let hexTimestamp = String(cleaned.prefix(12))
+        guard let milliseconds = UInt64(hexTimestamp, radix: 16) else { return nil }
+        return TimeInterval(milliseconds) / 1_000.0
+    }
```
marandaneto, line 108: should we just read the timestamp from the file metadata instead? not sure if any cheaper but at least maybe more reliable?

marandaneto: something like

```
let url = URL(fileURLWithPath: "/path/to/file")
let values = try url.resourceValues(forKeys: [.creationDateKey, .contentModificationDateKey])

let created = values.creationDate
let modified = values.contentModificationDate
```

PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -1173,6 +1214,43 @@
...
+        }
+
+        func replayQueueDidBufferSnapshot(_ replayQueue: PostHogReplayQueue) {
+            guard let postHog else { return }
+
+            let minimumDuration: TimeInterval? = bufferingLock.withLock { cachedMinimumDuration }
+            guard let minimumDuration, minimumDuration > 0 else {
+                // No minimum duration configured: should not be buffering, migrate immediately
+                bufferingLock.withLock { hasPassedMinimumDuration = true }
+                replayQueue.migrateBufferToQueue()
```
marandaneto, line 1238: are those heavy migrateBufferToQueue calls and similar IO operations happening NOT in the main thread?
they should not, since those operations can drop a lot of frames

marandaneto (approved):
left 2 comments otherwise LGTM
the only blocker one is if we are doing too much in the main thread otherwise all good

PostHog/Replay/PostHogReplayBufferQueue.swift
ioannisj: Parsing filename is definitely faster and I think I'd keep that as-is. Should be safe since filename is immutable and added as a UUIDv7 when queue.add() is called. Since this is a utility func I'd lean towards keeping light

PostHog/Replay/PostHogReplayIntegration.swift
ioannisj: Normal replay activity should be off main thread yes. However, I'll explicitly move this to a dedicated background queue cause I think there's an edge case of someone calling capture("$snapshot") on main q that would trigger the migration.

ioannisj: Now switch to a background queue during migration if needed

PR #552, opened 2026-04-08
feat: add device bucketing support for stable flag assignment

dmarticus:
## Summary

- Adds `$device_id` support to the iOS SDK so feature flags can bucket on a stable device identifier instead of `distinct_id`
- The device ID persists across `identify()` and `reset()` calls, preventing variant flipping during identity transitions (anonymous → authenticated, logout → re-login)
- Sends `$device_id` in `/flags` request payload, matching the JS SDK's behavior from PostHog/posthog-js#3340

## Problem

Users running experiments targeting anonymous users see variant assignments flip when `identify()` or `reset()` changes the `distinct_id`. The hash changes → different bucket → different variant for the same physical device.

## Changes

- **`PostHog/PostHogStorage.swift`**: Add `deviceId` to `StorageKey` enum (intentionally preserved across `reset()`)
- **`PostHog/PostHogStorageManager.swift`**: Add cached `deviceId` with thread-safe access, `getDeviceId()` with lazy-init for upgrades, and `initializeDeviceId()` for setup-time seeding
- **`PostHog/PostHogApi.swift`**: Include `$device_id` in `/flags` request payload when present
- **`PostHog/PostHogRemoteConfig.swift`**: Thread `deviceId` through `reloadFeatureFlags()` → `loadFeatureFlags()` → `api.flags()`
- **`PostHog/PostHogSDK.swift`**: Initialize `device_id` during `setup()`, expose public `getDeviceId()` method
- **`PostHogTests/PostHogDeviceBucketingTests.swift`**: 7 new tests covering init, persistence, identify/reset preservation, multiple cycles, flag request payload, and SDK restart

## Test plan

- [x] SDK builds cleanly (`swift build`)
- [x] SwiftLint passes
- [x] 7 new device bucketing tests written (test target has pre-existing compilation errors in unrelated files on `main`)
- [ ] Manual verification with a flag configured for device bucketing on the server side

### DIFF
--- a/PostHog/PostHogSDK.swift
+++ b/PostHog/PostHogSDK.swift
@@ -108,6 +108,12 @@ let maxRetryDelay = 30.0
             let api = PostHogApi(config)
 
             config.storageManager = config.storageManager ?? PostHogStorageManager(config)
+
+            // Initialize device_id before remote config loads, so it's available for flag requests.
+            // This provides a stable identifier for device-level feature flag bucketing that
+            // survives identify() and reset(). Seeded from the anonymous ID on first init.
+            config.storageManager?.initializeDeviceId()
+
             remoteConfig = PostHogRemoteConfig(config, theStorage, api, { [weak self] in
                 self?.getDefaultPersonProperties() ?? [:]
             }, { [weak self] flagKey, flagValue in
@@ -188,6 +194,17 @@ let maxRetryDelay = 30.0
         }
 
         return config.storageManager?.getAnonymousId() ?? ""
+    }
+
+    /// Returns the stable device identifier used for device-level feature flag bucketing.
+    /// This ID persists across `identify()` and `reset()` calls, only changing on a fresh
+    /// app install, manual cache clearing, or OS-initiated storage cleanup.
+    @objc public func getDeviceId() -> String {
+        if !isEnabled() {
+            return ""
+        }
+
+        return config.storageManager?.getDeviceId() ?? ""
     }
 
     @objc public func getSessionId() -> String? {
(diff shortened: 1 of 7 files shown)


PostHog/PostHogStorageManager.swift
```diff
@@ -58,6 +60,41 @@ public class PostHogStorageManager {
...
+                    storage.setString(forKey: .deviceId, contents: anonId)
+                }
+            }
+        }
+        return deviceId ?? ""
+    }
+
+    /// Initializes device_id if not already persisted. Called during SDK setup to ensure
+    /// the device ID is seeded from the anonymous ID before any flag requests.
+    func initializeDeviceId() {
```
marandaneto: this method looks like the very same as `getDeviceId`, should we just kill this one and call `getDeviceId` instead?

marandaneto:
missing changesets @dmarticus

marandaneto (approved):

PostHog/PostHogSDK.swift
```diff
@@ -108,6 +108,11 @@ let maxRetryDelay = 30.0
             let api = PostHogApi(config)
 
             config.storageManager = config.storageManager ?? PostHogStorageManager(config)
+
+            // Ensure device_id is initialized before remote config loads, so it's
+            // available for flag requests. Seeded from the anonymous ID on first init.
+            _ = config.storageManager?.getDeviceId()
```
marandaneto, line 114:
```suggestion
            _ = getDeviceId()
```

unless the SDK isnt ready yet?

dmarticus: Good catch — you're right that it would be cleaner, but the SDK isn't ready yet at this point. This runs inside `setup()` before `enabled = true` is set (line 104), so `getDeviceId()` would hit the `isEnabled()` guard and return `""` without initializing anything. That's why we go through `config.storageManager` directly here.

marandaneto (approved):

PR #554, opened 2026-04-10
feat: add rageclick support for iOS/macCatalyst

ioannisj:

Docs PR: https://github.com/PostHog/posthog.com/pull/16309

Adds `$rageclick` support for iOS and Mac Catalyst. Touch events are observed via `ApplicationEventPublisher`, rage clicks are detected using the reusable `RageClickDetector` utility, and events are enriched with `$elements_chain` when available before being captured.

Also refactored to remove code smell of `PostHogIntegration` throwing during install

- Manual test on device
- Added unit tests

## :pencil: Checklist

## If releasing new changes


PostHog/Autocapture/PostHogRageClickIntegration.swift
```diff
@@ -0,0 +1,159 @@
...
+            }
+
+            let touchCoordinates = touch.location(in: window)
+            let eventData = touch.view?.eventData(touchCoordinates: touchCoordinates)
+            let elementsChain = eventData?.getElementChain() ?? ""
+
+            captureRageClickIfNeeded(
+                touchCoordinates: touchCoordinates,
+                screenName: eventData?.screenName ?? currentScreenName(),
+                elementsChain: elementsChain
```
marandaneto: we probably dont want to capture anything if elementsChain and/or currentScreenName is empty otherwise how else would the user know where that happened?

PostHog/Autocapture/PostHogRageClickIntegration.swift
```diff
@@ -0,0 +1,159 @@
...
+                  let touch = touches.first,
+                  touch.phase == .ended,
+                  touch.tapCount > 0
+            else {
+                return
+            }
+
+            let touchCoordinates = touch.location(in: window)
+            let eventData = touch.view?.eventData(touchCoordinates: touchCoordinates)
+            let elementsChain = eventData?.getElementChain() ?? ""
```
marandaneto, line 87: are elements identifiable? like, swiftui/uikit widgets that arent named/with an id, how users would know where this comes from?

PostHog/Autocapture/PostHogRageClickIntegration.swift
```diff
@@ -0,0 +1,159 @@
...
+            var properties: [String: Any] = [
+                "$touch_x": touchCoordinates.x,
+                "$touch_y": touchCoordinates.y,
+            ]
+
+            if let screenName {
+                properties["$screen_name"] = screenName
+            }
+
+            postHog.autocapture(
```
marandaneto: this method is for `$autocapture` events but now we pass a different event name? its a bit confusing

PostHog/PostHogConfig.swift
```diff
@@ -76,6 +76,18 @@ public typealias BeforeSendBlock = (PostHogEvent) -> PostHogEvent?
...
+        ///
+        /// When enabled, rapid taps in the same area will be captured as `$rageclick` events.
+        /// Works independently of `captureElementInteractions` — you can enable rage click
+        /// detection without enabling full autocapture.
+        ///
+        /// Default: true
+        @objc public var captureRageClicks: Bool = true
+
+        /// Rage click detection configuration
+        @objc public let rageClickConfig: PostHogRageClickConfig = .init()
```
marandaneto, line 81: should we move captureRageClicks inside of `PostHogRageClickConfig` instead? feels like its different than the rest but its more organized and how we should have done before i guess

PostHog/PostHogSDK.swift
```diff
@@ -1009,6 +1005,7 @@ let maxRetryDelay = 30.0
     }
 
     func autocapture(
+        eventName: String = "$autocapture",
```
marandaneto: would that be a breaking change for objc methods?

marandaneto:
left a couple of comments but interested in @turnipdabeets review here as well

PostHog/Autocapture/PostHogRageClickIntegration.swift
ioannisj: Right, it'd keep just either one to be present. People may still rageclick on what seems to be an element (i.e background images) so I think we'd want to capture that as well. I think screenName OR elementsChain should be enough context, since there's also $touch_x,$touch_y coords passed along with the event

PostHog/Autocapture/PostHogRageClickIntegration.swift
ioannisj: element_chain is best effort here I think. We run the same detection logic we have for $autocapture, but people can rageclick wherever on the screen. Screen name + $touch_x,$touch_y should give context in those cases

PostHog/Autocapture/PostHogRageClickIntegration.swift
ioannisj: True, $rageclick has the same event structure as $autocapture. I'll refactor this

PostHog/PostHogConfig.swift
ioannisj: Let's keep as is just to be consistent with the API and add this as an item in https://github.com/PostHog/posthog-ios/issues/557. I agree thought that we should have been doing this before. wdyt?

PostHog/Autocapture/PostHogRageClickIntegration.swift
marandaneto: well... x and y positions, on an image element, with an app with a few different screens, is just noisy to me, and people will pay for those events, so the event should be actionable in a way
if you think you can reliably track where the clicks come from within an app with multiple screens, then we good, if not, lets rethink this

eg the widget has an identifier, then we good, then we dont need a screen name

PostHog/PostHogSDK.swift
ioannisj: This is an internal method, so I think we're okay

PostHog/PostHogConfig.swift
marandaneto: mm i think keeping adding just for the consistency isnt the best answer here
its just more migration for us and customers later in the next major version
thats my 2 cents, but feel free to ignore if not

PostHog/Autocapture/PostHogRageClickIntegration.swift
ioannisj: Hm good point. So is it okay to say:  
- if missing screen name, we required an element id. 
- In other cases, just an element chain (no id) + x,y + screen name feels enough info to be valuable

marandaneto: autocapture events would need something similar i think? not a blocker for this pr since its the way how it works, but maybe a follow up if you think this would capture more actionable events

ioannisj: Yeah, valid concern. I'll revisit autocapture on iOS and the implementation I'm working on for Android

PostHog/PostHogConfig.swift
ioannisj: done as `rageClickConfig.enabled`, updated docs PR as well

ioannisj:
> left a couple of comments but interested in @turnipdabeets review here as well

@turnipdabeets can you have a look when you get a chance pls?

PostHog/Autocapture/PostHogRageClickConfig.swift
```diff
@@ -0,0 +1,32 @@
...
+
+@objc(PostHogRageClickConfig) public class PostHogRageClickConfig: NSObject {
+    /// Enable rage click detection for iOS/macCatalyst.
+    ///
+    /// When enabled, rapid taps in the same area will be captured as `$rageclick` events.
+    /// Works independently of `captureElementInteractions` — you can enable rage click
+    /// detection without enabling full autocapture.
+    ///
+    /// Default: true
+    @objc public var enabled: Bool = true
```
turnipdabeets, line 11: Might be worth adding in docs or changelog that this is opt-out

PostHog/Autocapture/PostHogRageClickIntegration.swift
```diff
@@ -0,0 +1,170 @@
...
+                      timestamp: ProcessInfo.processInfo.systemUptime
+                  )
+            else {
+                return
+            }
+
+            let normalizedScreenName = screenName?.trimmingCharacters(in: .whitespacesAndNewlines)
+            let normalizedElementsChain = elementsChain.trimmingCharacters(in: .whitespacesAndNewlines)
+            let hasScreenName = normalizedScreenName?.isEmpty == false
+            let hasElementId = normalizedElementsChain.contains("attr_id=")
```
turnipdabeets: What do you think of doing this instead?

```suggestion
            let hasElementId = eventData?.viewHierarchy.contains { $0.label?.isEmpty == false } ?? false
```

PostHog/Autocapture/PostHogRageClickIntegration.swift
```diff
@@ -0,0 +1,170 @@
...
+            else {
+                return
+            }
+
+            // Rage clicks represent repeated single-pointer frustration taps.
+            // Ignore multi-touch interactions (e.g. pinch/zoom) to avoid false positives.
+            guard touches.count == 1,
+                  let touch = touches.first,
+                  touch.phase == .ended,
+                  touch.tapCount > 0
```
turnipdabeets, line 80: I think on iPad (Split View / Stage Manager), getCurrentWindow() only returns one scene's key window, so taps in other foreground scenes get silently dropped. Letting the touch carry its own window fixes that. 
```suggestion
            guard touches.count == .touches,
                  let touches = event.allTouches,                           
                  touches.count == 1,
                   let touch = touches.first,
                   let window = touch.window,
                    touch.phase == .ended,                                                                                                                                                                                           
                   touch.tapCount > 0
```

PostHog/Autocapture/PostHogRageClickConfig.swift
ioannisj: Already mentioned in the docs PR [here](https://github.com/PostHog/posthog.com/pull/16309). Should be enough I think?

PostHog/Autocapture/PostHogRageClickIntegration.swift
ioannisj: Yup looks cleaner

PostHog/Autocapture/PostHogRageClickIntegration.swift
ioannisj: Yeah, not sure this will compile but letting touch carry its own window makes sense. Will refactor

turnipdabeets (approved):

PR #564, opened 2026-04-16
fix: swift toolchain 5.3 compatibility issue with Xcode 15.4

ioannisj:

This PR fixes a Swift 5.3 compatibility regression introduced in error tracking code (added in `3.45.0`). Error happens only when building with swift 5.3 toolchain (i.e Xcode 15.4 or older)

The fix was replacing swift key-path shorthand method references (e.g. `map(\.toDictionary)`) with closure syntax and added a CI job to avoid future regressions

Added CI build step for swift 5.3 which [failed](https://github.com/PostHog/posthog-ios/actions/runs/24519241872/job/71671998235?pr=564) before the fix here

## :pencil: Checklist

## If releasing new changes


PostHog/ErrorTracking/PostHogExceptionProcessor.swift
```diff
@@ -386,7 +386,7 @@ enum PostHogExceptionProcessor {
 
         // Reverse to bottom-up order (outermost/main first, innermost/crash site last)
         // to match the Sentry event format. callStackReturnAddresses is top-down.
-        let frameDicts = frames.reversed().compactMap(\.toDictionary)
+        let frameDicts = frames.reversed().map { frame in frame.toDictionary }
```
turnipdabeets: Should this and the one below preserve compactMap since those filter out nils?

PostHog/ErrorTracking/PostHogExceptionProcessor.swift
```diff
@@ -411,7 +411,7 @@ enum PostHogExceptionProcessor {
 
         // Reverse to bottom-up order (outermost/main first, innermost/throw site last)
         // to match the Sentry event format. callStackReturnAddresses is top-down.
-        let frameDicts = frames.reversed().compactMap(\.toDictionary)
+        let frameDicts = frames.reversed().map { frame in frame.toDictionary }
```
turnipdabeets: this one too

ioannisj: Oh good catch, yes thats true!

ioannisj: Let's see if CI will be happy now

PostHog/ErrorTracking/PostHogCrashReportProcessor.swift
```diff
@@ -121,7 +121,7 @@ import Foundation
             // the Sentry event format. PLCrashReport delivers them top-down, so reverse.
             if !stackFrames.isEmpty {
                 exception["stacktrace"] = [
-                    "frames": stackFrames.reversed().map(\.toDictionary),
+                    "frames": stackFrames.reversed().compactMap { frame in frame.toDictionary },
```
turnipdabeets, line 124: haha this one got flipped the other way 😄

turnipdabeets (approved):

PostHog/ErrorTracking/PostHogCrashReportProcessor.swift
ioannisj: Fixed, need to actually write code more :)

PR #572, opened 2026-04-21
feat: Add tracing headers for URLSession requests

dustinbyrne:

Add support for `addTracingHeaders` on iOS so apps can automatically attach PostHog tracing headers to selected outbound `URLSession` requests.

This is aimed at teams using `URLSession` directly across both classic completion-handler APIs and async/await APIs who want consistent outbound tracing headers on iOS.

- Added `PostHogTracingHeadersTest` coverage for:
  - classic `URLSession` data/upload/download task APIs
  - async/await `URLSession` data/upload/download/bytes APIs
  - custom `URLSession` configurations
  - exact-host matching and hostname normalization
  - unlisted-host omission
  - ended-session omission of session headers
  - single-application of request modifiers for URL overloads
- Ran:
  - `make test filter=PostHogConfigTest`
  - `xcrun xcodebuild test -scheme PostHog -destination 'id=043AFB16-0063-4585-9666-03FEC5FF694E' -only-testing:PostHogTests/PostHogTracingHeadersTest`

## :pencil: Checklist

## If releasing new changes

### DIFF
--- a/PostHog/PostHogConfig.swift
+++ b/PostHog/PostHogConfig.swift
@@ -188,7 +188,7 @@ public typealias BeforeSendBlock = (PostHogEvent) -> PostHogEvent?
         /// - Hostname matching is exact and does not include ports or subdomain wildcards
         /// - iOS does not send `X-POSTHOG-WINDOW-ID` because mobile apps do not have a per-window/tab concept
         /// - Existing values for these headers will be overwritten
-        @objc public var addTracingHeaders: [String]?
+        @objc public var tracingHeaders: [String]?
 
         /// Enable Recording of Session Replays for iOS
         ///
@@ -310,7 +310,7 @@ public typealias BeforeSendBlock = (PostHogEvent) -> PostHogEvent?
         }
 
         #if os(iOS)
-            if addTracingHeaders?.isEmpty == false {
+            if tracingHeaders?.isEmpty == false {
                 integrations.append(PostHogTracingHeadersIntegration())
             }
 
--- a/PostHog/PostHogConfig.swift
+++ b/PostHog/PostHogConfig.swift
@@ -176,6 +176,20 @@ public typealias BeforeSendBlock = (PostHogEvent) -> PostHogEvent?
     public static let defaultHost: String = "https://us.i.posthog.com"
 
     #if os(iOS)
+        /// When set, PostHog injects tracing headers into `URLSession` requests whose
+        /// destination hostname exactly matches one of the configured hostnames.
+        ///
+        /// Injected headers on iOS:
+        /// - `X-POSTHOG-DISTINCT-ID`
+        /// - `X-POSTHOG-SESSION-ID`
+        ///
+        /// Notes:
+        /// - Requires `enableSwizzling = true`
+        /// - Hostname matching is exact and does not include ports or subdomain wildcards
+        /// - iOS does not send `X-POSTHOG-WINDOW-ID` because mobile apps do not have a per-window/tab concept
+        /// - Existing values for these headers will be overwritten
+        @objc public var addTracingHeaders: [String]?
+
         /// Enable Recording of Session Replays for iOS
         ///
         /// Note: Ingestion controls (sampling, feature flags, and event triggers) are currently applied using AND logic.
@@ -296,6 +310,10 @@ public typealias BeforeSendBlock = (PostHogEvent) -> PostHogEvent?
         }
 
         #if os(iOS)
+            if addTracingHeaders?.isEmpty == false {
+                integrations.append(PostHogTracingHeadersIntegration())
+            }
+
             if sessionReplay {
                 integrations.append(PostHogReplayIntegration())
             }
--- a/PostHog/Replay/Plugins/Network/PostHogSessionReplayNetworkPlugin.swift
+++ b/PostHog/Replay/Plugins/Network/PostHogSessionReplayNetworkPlugin.swift
@@ -22,6 +22,9 @@
 
             let interceptor = URLSessionInterceptor(
                 shouldCapture: shouldCaptureNetworkSample,
+                shouldCaptureRequest: { [weak self] request in
+                    self?.shouldCaptureNetworkRequest(request) ?? false
+                },
                 onCapture: handleNetworkSample,
                 getSessionId: { [weak self] date in
                     self?.postHog?.sessionManager.getSessionId(at: date)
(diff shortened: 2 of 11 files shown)


PostHog/Replay/Plugins/Network/URLSessionSwizzler.swift
```diff
@@ -10,240 +10,968 @@
 
     import Foundation
 
+    final class URLSessionInstrumentation {
```
dustinbyrne, line 13: This has been generalized such that tracing headers and session replay network capture can coexist via a registration pattern using the same URLSession swizzles

PostHog/Replay/Plugins/Network/URLSessionSwizzler.swift
```diff
@@ -10,240 +10,968 @@
...
+            )
+            downloadTaskWithURL = try DownloadTaskWithURL.build(modifyRequest: modifyRequest)
+
+            // Async/await URLSession convenience APIs are iOS 15+, so only install the
+            // task-level resume fallback on those OS versions.
+            if #available(iOS 15, *) {
+                taskResume = try TaskResume.build(modifyRequest: modifyRequest)
+            } else {
+                taskResume = nil
+            }
```
dustinbyrne, line 222: This is what gives us support for async URLSession requests. It's currently scoped only to request mutation, meaning it won't (yet) add async request support to session replay network capture.

I'll add that in a follow up.

PostHog/Replay/Plugins/Network/URLSessionSwizzler.swift
```diff
@@ -10,240 +10,968 @@
...
+            uploadTaskWithRequestAndFile.swizzle()
+
+            downloadTaskWithRequestAndCompletion.swizzle()
+            downloadTaskWithURLAndCompletion.swizzle()
+            downloadTaskWithRequest.swizzle()
+            downloadTaskWithURL.swizzle()
+            taskResume?.swizzle()
+
             hasSwizzled = true
         }
```
dustinbyrne, line 244: Worth noting that this introduces additional URLSession API swizzles. Because these swizzles are shared, this might increase session replay network capture coverage.

Any considerations here?

turnipdabeets (approved):

PostHog/Replay/Plugins/Network/URLSessionSwizzler.swift
ioannisj: Did we test if this doesn't break network capture plugin for session replay?

PostHog/Tracing/PostHogTracingHeadersIntegration.swift
```diff
@@ -0,0 +1,151 @@
+#if os(iOS)
```
ioannisj, line 1: Is there a reason this is only iOS? Previously network instrumentation was used only for session replay which is iOS only, but feels like we can now expand this?

PostHog/Tracing/PostHogTracingHeadersIntegration.swift
```diff
@@ -0,0 +1,151 @@
...
+        private func getNormalizedHostnames(for postHog: PostHogSDK) -> Set<String> {
+            let hostnames = postHog.config.addTracingHeaders ?? []
+
+            return normalizedHostnamesLock.withLock {
+                if hostnames != cachedHostnames {
+                    cachedHostnames = hostnames
+                    normalizedHostnames = PostHogTracingHeaders.normalizeHostnames(hostnames)
+                }
+
+                return normalizedHostnames
```
ioannisj: Config should be immutable once the SDK is setup, so I think we can just compute this once when the integration starts?

PostHog/PostHogConfig.swift
```diff
@@ -162,6 +162,20 @@ public typealias BeforeSendBlock = (PostHogEvent) -> PostHogEvent?
...
+        /// Injected headers on iOS:
+        /// - `X-POSTHOG-DISTINCT-ID`
+        /// - `X-POSTHOG-SESSION-ID`
+        ///
+        /// Notes:
+        /// - Requires `enableSwizzling = true`
+        /// - Hostname matching is exact and does not include ports or subdomain wildcards
+        /// - iOS does not send `X-POSTHOG-WINDOW-ID` because mobile apps do not have a per-window/tab concept
+        /// - Existing values for these headers will be overwritten
+        @objc public var addTracingHeaders: [String]?
```
marandaneto: similar to https://github.com/PostHog/posthog-android/pull/495#discussion_r3122587561
and https://github.com/PostHog/posthog-ios/blob/c165ead50ecf66368685a8467fa625fc0960bace/PostHog/PostHogConfig.swift#L307-L330
not sure if the concurrentmodification ios an issue here tho

PostHog/Replay/Plugins/Network/URLSessionSwizzler.swift
marandaneto: the more the better? :)

i'm just not sure whether eg dataTaskWithURL and dataTaskWithURLAndCompletion are the same thing, since they call each other internally? do they?

Also consider this https://github.com/PostHog/posthog-ios/blob/main/PostHog/Replay/Plugins/Network/URLSessionExtension.swift

We can't swizzle Swift methods so users have to use this wrapper
i believe we'd need to set the headers for those as well

PostHog/Replay/Plugins/Network/URLSessionSwizzler.swift
dustinbyrne: I added a couple more tests + verified it manually

PostHog/Tracing/PostHogTracingHeadersIntegration.swift
dustinbyrne: Agree we can likely expand this. I've kept it as iOS to be consistent with session replay, but we should do this as a follow up item.

PostHog/Tracing/PostHogTracingHeadersIntegration.swift
dustinbyrne: Good point, I've changed it

PostHog/PostHogConfig.swift
dustinbyrne: I've changed this such that we snapshot and normalize `tracingHeaders` on install instead of re-reading it on every request. That should remove any concurrent modification risk.

also renamed to `tracingHeaders`

PostHog/Replay/Plugins/Network/URLSessionSwizzler.swift
dustinbyrne: I added explicit test coverage for `postHogData`, `postHogUpload`, and `postHogDownload`. They do get the tracing headers as well, [and the tests verify that](https://github.com/PostHog/posthog-ios/blob/60f9b60cdfb2243e8f5cee795d36e0cf6c324ab6/PostHogTests/PostHogTracingHeadersTest.swift#L152-L168).

re: `dataTaskWithURL` and `dataTaskWithURLAndCompletion` potentially calling each other internally, I've added a test to ensure that for the overlapping URL overloads [we only apply the request modifier once](https://github.com/PostHog/posthog-ios/blob/60f9b60cdfb2243e8f5cee795d36e0cf6c324ab6/PostHogTests/PostHogTracingHeadersTest.swift#L170), so we don't double-inject headers

PostHog/Replay/Plugins/Network/PostHogSessionReplayNetworkPlugin.swift
```diff
@@ -75,11 +93,105 @@
...
+                return false
+            }
+
+            let path = strippingAPIHostPathPrefix(from: url.path, apiHostPath: apiHost.path)
+            let ignoredPaths = [snapshotEndpoint, "/batch", "/e/", "/i/"]
+
+            return ignoredPaths.contains { ignoredPath in
+                pathMatches(path, ignoredPath: ignoredPath)
+            }
+        }
```
dustinbyrne, line 109: A bit of scope creep, but I noticed session replay network capture was now picking up POST requests to `posthog.com/s/`, so I've ported this behavior from the browser SDK

ioannisj (approved):
I think should be good to go after resolving the merge conflict?

PR #574, opened 2026-04-23
fix: vendor PLCrashReporter

### LINKED ISSUE #570: CocoaPods: CrashReporter.xcframework collision with DatadogCrashReporting (PostHog ≥ 3.45)
## Summary

Starting with `PostHog` iOS SDK **3.45.0**, the pod vendors `vendor/CrashReporter.xcframework` directly. This collides with the `PLCrashReporter` pod pulled in transitively by **`DatadogCrashReporting`** (via `@datadog/mobile-react-native`), which ships the same framework by the same name. CocoaPods refuses to install both and aborts `pod install`.

The two frameworks are copies of the same underlying library (Microsoft's [`plcrashreporter`](https://github.com/microsoft/plcrashreporter)). `DatadogCrashReporting` follows the conventional CocoaPods distribution (`pod 'PLCrashReporter'`), which is what most Apple crash-reporting integrations — Firebase, Sentry, New Relic — depend on. PostHog 3.45+ diverged by bundling a private copy, which makes coexistence impossible under CocoaPods.

## Error

```
[!] The 'Pods-<AppTarget>' target has frameworks with conflicting names: crashreporter.xcframework.
```

## Environment

- `posthog-ios` **3.45.0** through **3.55.0** (latest) — all affected
- `@datadog/mobile-react-native` 3.0.2 → `DatadogCrashReporting` 3.4.0 → `PLCrashReporter` 1.12.0
- Reproduces via `posthog-react-native` + `posthog-react-native-session-replay` on React Native / Expo, but the root cause is iOS-layer only
- CocoaPods 1.15+, Xcode 16+

## Reproduction (minimal Podfile)

```ruby
target 'MyApp' do
  pod 'PostHog', '3.55.0'           # or any version >= 3.45.0
  pod 'DatadogCrashReporting', '~> 3.4'
end
```

`pod install` → the collision error above.

## Current workaround

For apps that use `posthog-react-native-session-replay` (which declares `PostHog '~> 3.21'`, accepting anything ≥ 3.21 < 4.0), adding an explicit app-target constraint to the Podfile forces CocoaPods to pick a pre-3.45 version that doesn't vendor the framework:

```ruby
target 'MyApp' do
  # ...
  pod 'PostHog', '< 3.45'
end
```

CocoaPods resolves to **PostHog 3.44.x**, which still includes session replay but not the vendored CrashReporter. Session-replay consumers who've bumped to `posthog-react-native-session-replay@1.5.0+` (PostHog `~> 3.53.1`) cannot use this workaround and are fully blocked.

## Related

- #407 — "Apple error tracking support" — the roadmap issue where @marandaneto flagged the "error tracking as separate module" plan.
- #472 — "Deprecate CocoaPods support" — PostHog's longer-term direction, but CocoaPods users still need this collision resolved before that migration completes.

ioannisj:

Closes #570 

Refactors PostHog error tracking integration to avoid risk of `PLCrashReporter` symbol/linking conflicts in mixed SDK setups (notably with Datadog/New Relic), while keeping  behavior intact across SPM and CocoaPods.

 - Customers reported integration issues caused by duplicate PLCrashReporter symbols when PostHog and another crash reporting SDK were both present.
 - Existing setup had mixed:
     - SPM dependency on `microsoft/plcrashreporter`
     - Pods vendored `CrashReporter.xcframework`

### What changed

 - Vendor `PLCrashReporter` source into PostHog (vendor/PHPLCrashReporter) and wire it as an internal module for SPM to isolate compiler flags.
 - Remove external SPM dependency on `microsoft/plcrashreporter` from `Package.swift` and `Package@swift-5.9.swift`.
 - Replace CrashReporter product usage with internal PHPLCrashReporter in spm target dependencies.
 - Update crash integration code/tests to use prefixed PH types (PHPLCrashReport, PHPLCrashReporter, etc.).
 - Update CocoaPods integration (PostHog.podspec) to compile vendored PLCrashReporter sources directly, including:
     - required preprocessor flags (PLCR_PRIVATE, PLCRASHREPORTER_PREFIX=PH, SWIFT_PACKAGE)
     - protobuf-c headers/sources
     - exclusions for watchOS/visionOS where crash reporting is unsupported
 - Fix repository ignore behavior (.gitignore) so vendored dependency sources are included correctly.
 - Internalization of shared_handler_context() linkage to remove duplicate-symbol collisions.

- [x] Verified builds with sample project matrix and verified previously failing builds are now okay
    - [x] pods-posthog
	- [x] pods-posthog-static
 	- [x] pods-posthog-static-framework
 	- [x] pods-posthog-datadog (plcrashreporter)
 	- [x] pods-posthog-newrelic (plcrashreporter)
 	- [x] pods-posthog-sdwebimage-webpcoder (libwebp)
 	- [x] pods-posthog-path
 	- [x] pods-posthog-watchos
 	- [x] pods-posthog-visionos
 	- [x] spm-posthog
 	- [x] spm-posthog-datadog (plcrashreporter)
 	- [x] spm-posthog-newrelic (plcrashreporter)
 	- [x] som-posthog-sdwebimage-webpcoder (libwebp)
 	- [x] spm-posthog-watchos 
 	- [x] spm-posthog-maccatalyst
 	- [x] spm-posthog-platforms
 	- [x] xcframework-external-sdk 
- [x] PH-prefixed types in unit tests and ensured unit tests pass
- [x] Validate React-Native 
- [x] Validate Flutter
- [x] Validate no functional regression

## :pencil: Checklist

## If releasing new changes


ioannisj:
@PostHog/team-client-libraries This should be somewhat ready to review. Just need to validate RN and Flutter builds and that error tracking functionality is intact (I don't see why it wouldn't). If you have any other local samples, would appreciate if you can point the dependency here for a quick test as well

PostHog.podspec
```diff
@@ -25,20 +25,44 @@ Pod::Spec.new do |s|
...
+    'HEADER_SEARCH_PATHS' => '$(inherited) "${PODS_TARGET_SRCROOT}/vendor/PHPLCrashReporter/Dependencies/protobuf-c" "${PODS_TARGET_SRCROOT}/vendor/PHPLCrashReporter/Dependencies/protobuf-c/protobuf-c" "${PODS_TARGET_SRCROOT}/vendor/PHPLCrashReporter/Source"'
+  }
 
   s.source_files = [
     'PostHog/**/*.{swift,h,hpp,m,mm,c,cpp}',
-    'vendor/libwebp/**/*.{h,c}'
+    'vendor/libwebp/**/*.{h,c}',
+    'vendor/PHPLCrashReporter/Source/**/*.{h,hpp,m,mm,c,cpp,S}',
+    'vendor/PHPLCrashReporter/Dependencies/protobuf-c/**/*.h',
+    'vendor/PHPLCrashReporter/Dependencies/protobuf-c/**/*.c'
```
dustinbyrne, line 42: Should the vendored header files be `private_header_files`? Otherwise, if I understand correctly, CocoaPods re-exports them through the generated `PostHog` umbrella header, and we don't expose these types through our public API

PostHog/ErrorTracking/PostHogCrashReportProcessor.swift
```diff
@@ -8,16 +8,19 @@
 import Foundation
 
 #if os(iOS) || os(macOS) || os(tvOS)
-    import CrashReporter
+    #if canImport(PHPLCrashReporter)
+        // SPM-only helper module. CocoaPods/Xcodeproj builds compile vendored sources directly into PostHog.
+        @_implementationOnly import PHPLCrashReporter
+    #endif
```
dustinbyrne: I noticed this raises a warning during build, is this expected?

>warning: using '@_implementationOnly' without enabling library evolution for 'PostHog' may lead to instability during execution

PostHog.podspec
ioannisj: Sounds like we should yes. Let me check what gets generated and come back

PostHog/ErrorTracking/PostHogCrashReportProcessor.swift
ioannisj: Yes I think that's because we are using an unofficial attribute here @_implementationOnly. The proper way to do this is `internal import PHPLCrashReporter`, but that was introduced with Swift 5.9. I'll see if there's a way to use that instead and fallback to @_implementationOnly for people using older toolchains

ioannisj: iirc we do the same for libweb for people bundling PostHog in their .xcframework with ABI stability turned on

dustinbyrne (approved):
Approving to unblock

PostHog/ErrorTracking/PostHogCrashReportProcessor.swift
ioannisj: Was planning to replace with `internal import PHPLCrashReporter` using `#if swift(>=5.9)` but decided to skip because:
- Seems to have slightly different behavior than @ _implementationOnly
- Pods refactor require @_implementationOnly
- Other like Sentry use the same
- Did not cause any known issues so far since it was used with phlibwebp

turnipdabeets:
It might be worth dropping a tiny UPSTREAM.md or VENDORING.md recording the upstream commit/tag this was forked from. Once we own the source, Package.resolved won't tell us what version we're on so the next sync and any CVE response will be much easier with a paper trail.

turnipdabeets (approved):

PostHog.podspec
ioannisj: This was indeed the case.  Refactored so that now Pods use private module maps instead. Public umbrella header now exposes only PostHog/PostHog.h, so vendored headers stay private. Run the integration test matrix in PR body and everything is green.

ioannisj:
> It might be worth dropping a tiny UPSTREAM.md or VENDORING.md recording the upstream commit/tag this was forked from. Once we own the source, Package.resolved won't tell us what version we're on so the next sync and any CVE response will be much easier with a paper trail.

Great point! Added

PR #577, opened 2026-04-24
fix: capture Swift runtime crash messages from __crash_info

ioannisj:

Closes #522 

Swift runtime crashes like **fatalError**, **assert**, **precondition**, and **force unwraps** are currently captured as generic `SIGTRAP` without the original Swift message. This change reads the runtime message from `__crash_info` and propagates it to `$exception`'s payload.

- Added swift_crash_info_message to `PLCrashReport.proto` and regenerated protobuf models
- Added __crash_info extraction in `PLCrashLogWriter.m`
- Exposed parsed value on PLCrashReportSignalInfo as `swiftCrashInfoMessage`
- Updated `PostHogCrashReportProcessor` to prefer `swiftCrashInfoMessage` as an exception value for signal exceptions.
- Updated `PostHogCrashReportProcessor` to extract error type from Swift message (e.g Fatal error)

For the lookup strategy we scan the image containing the crash-site address first, then falls back to scanning loaded images. This improves coverage and was required for force-unwrap crashes, where the message source isn’t always tied to `libswiftCore` in the same way as fatalError/assert and precondition are.

**Note:** For for force-unwraps in optimized builds, I could not find a Swift crash message available to extract in any of the images. I think it may not be available

- Manual testing with local sample app

## :pencil: Checklist

## If releasing new changes


marandaneto:
thats cool!

PostHog/ErrorTracking/PostHogCrashReportProcessor.swift
```diff
@@ -76,14 +76,11 @@ import Foundation
...
-                // See: https://github.com/getsentry/sentry-cocoa/pull/1596
-                //      https://github.com/bugsnag/bugsnag-cocoa/pull/948
-                // Future enhancement: implement __crash_info parsing in PHPLCrashReporter for richer Swift crash messages.
-                exception["type"] = signalInfo.name
-                exception["value"] = signalMessage(signalInfo)
+                // For Swift runtime traps (fatalError/assert/precondition), use __crash_info
+                // to override the generic signal type while preserving the original message.
+                let swiftCrashInfoMessage = sanitizeSwiftCrashInfoMessage(signalInfo.swiftCrashInfoMessage)
+                exception["type"] = swiftFatalErrorType(from: swiftCrashInfoMessage) ?? signalInfo.name
+                exception["value"] = swiftCrashInfoMessage ?? signalMessage(signalInfo)
```
dustinbyrne: If we don't recognize the error type, we fall back to `signalInfo.name` as the type but we still use the `__crash_info` message as the exception value. Is this right, or should both be gated on the error type?

It looks like `__crash_info` isn't guaranteed to be associated with the current signal

ioannisj: Yes good point since __crash_info can be stale or unrelated to current crash (since it could be read from any loaded image). However, since its purpose is to carry diagnostics, I added it to "meta" payload so we don't discard it all together. It's now only promoted to exception value when we can parse its type

dustinbyrne (approved):

Issue #571, opened 2026-04-21
Session replay screenshot pipeline still appears to cause iOS app hangs after #532


Jaltaire:
I want to report what looks like a remaining session replay issue in the iOS screenshot pipeline.

I previously commented on [#553](https://github.com/PostHog/posthog-ios/issues/553#issuecomment-4256427263) saying that patching locally seemed to help and asking for new React Native releases. That now looks premature.

I am intentionally removing app-identifying details below, but this is from a production React Native iOS app on Expo SDK 55 using:

- `posthog-react-native 4.42.1`
- `posthog-react-native-session-replay 1.5.4`

**Observed behavior**

- App hangs for roughly 2.4 to 3.6 seconds on iOS (per Sentry). In practice, we are seeing the same crashing issue as before.
- The strongest trace goes through `PostHogReplayIntegration.toScreenshotWireframe` and `UIView.toImage`.
- This is happening on an actively updating React Native screen, which may be relevant because the UI is changing frequently while replay is enabled.
- We are seeing this in production, not only in local profiling.

**Sanitized stack trace from the clearest hang**

```text
closure in PostHogReplayIntegration.start
PostHogReplayIntegration.snapshot
PostHogReplayIntegration.generateSnapshot
closure in PostHogReplayIntegration.generateSnapshot
PostHogReplayIntegration.toScreenshotWireframe
UIView.toImage
closure in UIView.toImage
-[UIGraphicsImageRenderer imageWithActions:]
-[UIGraphicsRenderer runDrawingActions:completionActions:error:]
-[UIGraphicsRenderer runDrawingActions:completionActions:format:error:]
__44-[UIGraphicsImageRenderer imageWithActions:]_block_invoke
-[UIGraphicsImageRendererContext currentImage]
rip_auto_context_create_image
rip_auto_context_rasterization_loop
CGDisplayListDrawInContextDelegate
ripc_DrawImage
ripc_AcquireRIPImageData
CGSImageDataLock
img_raw_read
CGDataProviderDirectGetBytesAtPositionInternal
CGColorTransformConvertUsingCMSConverter
convert_using_vImageConverter
vImageConverterConvert
vImageConvert_AnyToAny
```

**Second sanitized hang from the same production build**

This one is less conclusive, but I am including it because it happened under the same general conditions:

```text
AppDelegate.main
UIApplicationMain
-[UIApplication _run]
GSEventRunModal
CFRunLoopRunSpecific
__CFRunLoopRun
[...]

PR #578, opened 2026-04-25
fix: improve session replay screenshot performance

### LINKED ISSUE #321: PostHog's Session Replay feature degrades Google services performance

ioannisj:

Closes #321 and possibly #571

This PR improves screenshot performance by replacing `UIGraphicsImageRenderer` with a custom and more lightweight CGContext-backed renderer.

Physical-device benchmarking and profiling on an old iPhone XR showed the UIView.toImage() path was the dominant source of UI glitches:

 - Baseline screenshot snapshots cost ~135ms on the main thread.
 - Each snapshot correlated with a >33ms frame hitch.
 - At higher snapshot rates, replay could saturate the app and drop effective frame pacing below 60fps.

### Changes
 - Adds `PostHogGraphicsImageRenderer`, a direct CoreGraphics bitmap renderer.
 - Uses it for `UIView.toImage()` screenshot capture.
 - Uses it for `RRWireframe.maskImage()` masked screenshot redraw.
 - Does not change masking traversal logic, masking decisions, serialization, or replay event structure. 

### Findings

Physical device benchmark on an iPhone XR (`iPhone11,8`), running screenshots on a complex idle view with re-layout ticks (to trigger snapshots)

Judging from the metrics collected, snapshot main-thread cost dropped by ~92%:

 | Throttle | Baseline snapshot avg | This PR snapshot avg | Delta |
 |---:|---:|---:|---:|
 | 0.50s | 134.22ms | 11.14ms | -91.7% |
 | 0.25s | 134.58ms | 10.69ms | -92.1% |
 | 0.10s | 138.92ms | 10.66ms | -92.3% |

There was also a big improvement on frame pacing and app hangs. This was also prominent when running the app, frame rate stayed at 60fps at all throttle levels (not recommended if you are reading this):

 | Throttle | Baseline frames >33ms | This PR frames >33ms |
 |---:|---:|---:|
 | 0.50s | 29 | 0 |
 | 0.25s | 57 | 0 |
 | 0.10s | 93 | 0 |
 
**Before**

https://github.com/user-attachments/assets/adc13e41-1a83-4f54-9295-c79bbfab4b65 

**After**

https://github.com/user-attachments/assets/c283d247-1b58-4eb9-9818-5326b411ae9c

### Post-release

- [ ] Bump RN posthog-ios min version
- [ ] Bump Flutter posthog-ios min version

### Note

We previously experimented with using a custom renderer to improve performance. Back then, the timing results were not that convincing and I decided not to ship a fix, which was the wrong call. This could be because: 
- I had the wrong approach on profiling the improvement and baseline
- I was not profiling/testing on the correct device 

This one is on me because we could have shipped this improvement much earlier!

- Profiling with Xcode Instruments and custom benchmarking
- Running on a lower-end physical device
- Verified that session recordings are scaled and masked correctly in the PostHog UI
- This 👇 
 

## :pencil: Checklist

## If releasing new changes

### DIFF
--- a/PostHog/Replay/UIView+Util.swift
+++ b/PostHog/Replay/UIView+Util.swift
@@ -36,22 +36,17 @@
         }
 
         func toImage() -> UIImage? {
-            // Avoid Rendering Offscreen Views
-            let bounds = superview?.bounds ?? bounds
-            let size = bounds.intersection(bounds).size
+            let bounds = self.bounds
+            let size = bounds.size
 
             if !size.hasSize() {
                 return nil
             }
 
-            let rendererFormat = UIGraphicsImageRendererFormat.default()
-
-            // This can significantly improve rendering performance because the renderer won't need to
-            // process transparency.
-            rendererFormat.opaque = isOpaque
-            // Another way to improve rendering performance is to scale the renderer's content.
-            // rendererFormat.scale = 0.5
-            let renderer = UIGraphicsImageRenderer(size: size, format: rendererFormat)
+            // Use native screen scale for best drawHierarchy performance.
+            // Using a non-native scale can trigger internal rescaling overhead.
+            let nativeScale = (self as? UIWindow ?? window)?.screen.scale ?? 1
+            let renderer = PostHogGraphicsImageRenderer(size: size, scale: nativeScale)
 
             return autoreleasepool {
                 renderer.image { _ in
--- a/.changeset/tasty-hands-like.md
+++ b/.changeset/tasty-hands-like.md
@@ -1,0 +1,5 @@
+---
+"posthog-ios": patch
+---
+
+fix: improve session replay screenshot performance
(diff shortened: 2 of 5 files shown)


PostHog/Replay/UIView+Util.swift
```diff
@@ -36,22 +36,17 @@
         }
 
         func toImage() -> UIImage? {
-            // Avoid Rendering Offscreen Views
-            let bounds = superview?.bounds ?? bounds
-            let size = bounds.intersection(bounds).size
```
ioannisj, line 41: Unrelated change but:
1. this is always called on a UIWindow, and superview here will always be null
2. bounds.intersection(bounds).size will always be bounds.size

marandaneto (approved):

PR #582, opened 2026-04-30
fix: no-op setup for empty project token

marandaneto:

`PostHogConfig.normalizeProjectToken` already handles token normalization, so setup should be the single place that explains why initialization stops when the normalized project token/API key is empty.

This updates the setup log to clarify that the PostHog SDK will remain disabled when no project token or API key is provided.

- `make format`
- `make test` (latest run was interrupted by follow-up request; an earlier run had one `PostHogQueueTest` failure unrelated to this logging change)

## :pencil: Checklist

## If releasing new changes

- [x] Added a changeset file


sdk_compliance_adapter/Package.resolved
```diff
@@ -36,15 +36,6 @@
         "version" : "4.7.1"
       }
     },
-    {
```
marandaneto, line 39: @ioannisj leftover from vendoring?

ioannisj: Probably yes, I don't see how else this got here

ioannisj (approved):

PR #584, opened 2026-05-04
fix(reachability): broadcast network changes to multiple subscribers

turnipdabeets:

Noticed this while trying to create a new logQueue. We'll need this for the new queue as well as fixes for the other two. 

`Reachability.whenReachable` / `whenUnreachable` are single-value `var`s. Each subscriber overwrites the previous one. Since v3.0 the replay queue's subscription has silently disconnected the events queue from network notifications: `dataMode = .wifi` is ignored, auto-flush on WiFi reconnect doesn't fire, and the offline pause is reactive (after a wasted request) instead of proactive. No existing test exercises reachability transitions — every test sets `disableReachabilityForTesting = true` — so the regression went undetected.

This PR adds `onReachable` / `onUnreachable` `PostHogMulticastCallback`s on `Reachability` and migrates `PostHogQueue` (events) to subscribe via the multicast. Legacy single-callback fields are kept for external compat and marked `@available(*, deprecated)`. The events queue now correctly responds to network changes again.

**Unit tests** — two new tests in `PostHogReachabilityTest.swift`:
- `multicastNoStomp` — 4 independent subscribers on the same `Reachability`, every transition fires every subscriber. Would have shown 0 for the first-registered subscriber under the old single-slot behaviour.
- `tokenDeallocUnsubscribes` — dropping a `RegistrationToken` removes the subscriber.

**Local verification** — beyond the unit tests, I ran four extra checks:

1. **Build with deprecation visible**: `xcrun swift build --target PostHog --arch arm64` emits two deprecation warnings at the internal firing sites — confirming the `@available(*, deprecated)` markers compile and apply.

2. **External consumer check**: wrote `/tmp/posthog_consumer_check.swift` that imports PostHog without `@testable` and uses both APIs. Type-checked against the built `.swiftmodule`:
   ```
   xcrun swiftc -typecheck -I .build/arm64-apple-macosx/debug/Modules /tmp/posthog_consumer_check.swift
   ```
   Output: ✓ deprecation warnings emit when setting `whenReachable` / `whenUnreachable` — the migration nudge surfaces correctly to external users.

3. **Before/after demonstration**: wrote `/tmp/reachability_verification.swift` that simulates the old single-`var`-slot pattern alongside the new multicast pattern, side by side. Run output:
   ```
   === Old (single-slot) behaviour ===
     subscriber A (events) fired:  0  ← BUG: silently overwritten
     subscriber B (replay) fired:  1

   === New (multicast) behaviour ===
     subscriber A (events) fired:  1
     subscriber B (replay) fired:  1
     subscriber C (logs)   fired:  1   ← all three coexist

   === Token deinit unsubscribes cleanly ===
     while token alive, fired:    1
     after token dropped, fired:  1   ← unchanged
   ```
   The first block reproduces the actual production bug; events would fire 0 times once replay subscribed.

4. **Full SDK pipeline**:
   - `make test` → 463 tests pass.
   - `make lint` → 0 serious violations.
   - `xcodebuild -scheme PostHog -destination generic/platform=ios` → BUILD SUCCEEDED.

## :pencil: Checklist

## If releasing new changes

### DIFF
--- a/PostHog/Utils/Reachability.swift
+++ b/PostHog/Utils/Reachability.swift
@@ -45,14 +45,11 @@ import Foundation
         static let reachabilityChanged = Notification.Name("reachabilityChanged")
     }
 
-    public class Reachability {
-        public typealias NetworkReachable = (Reachability) -> Void
-        public typealias NetworkUnreachable = (Reachability) -> Void
-
+    class Reachability {
         @available(*, unavailable, renamed: "Connection")
-        public enum NetworkStatus: CustomStringConvertible {
+        enum NetworkStatus: CustomStringConvertible {
             case notReachable, reachableViaWiFi, reachableViaWWAN
-            public var description: String {
+            var description: String {
                 switch self {
                 case .reachableViaWWAN: return "Cellular"
                 case .reachableViaWiFi: return "WiFi"
@@ -61,9 +58,9 @@ import Foundation
             }
         }
 
-        public enum Connection: CustomStringConvertible {
+        enum Connection: CustomStringConvertible {
             case unavailable, wifi, cellular
-            public var description: String {
+            var description: String {
                 switch self {
                 case .cellular: return "Cellular"
                 case .wifi: return "WiFi"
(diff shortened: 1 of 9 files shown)


PostHog/Utils/Reachability.swift
```diff
@@ -75,9 +75,18 @@ import Foundation
             public static let none: Connection = .unavailable
         }
 
+        @available(*, deprecated, message: "Subscribe via onReachable for multi-subscriber safety. Setting this overwrites any prior setter.")
```
turnipdabeets: @ioannisj do you think we could delete these or should we leave as deprecated?

marandaneto: this file is internal so i dont see a reason to add a deprecated note

ioannisj: Yeah I think this was public when vendored in but it's used only internally. Worth changing visibility to Internal as well maybe?

turnipdabeets: removed those two and changed visibility to Internal

marandaneto (approved):

marandaneto:
>  Test "captures Application Backgrounded event" recorded an issue at PostHogAppLifeCycleIntegrationTest.swift:280:9: Expectation failed: (events.count → 1) == 3
↳ // Verify Application Installed event
Swift/ContiguousArrayBuffer.swift:690: Fatal error: Index out of range

flaky test?

turnipdabeets:
> > Test "captures Application Backgrounded event" recorded an issue at PostHogAppLifeCycleIntegrationTest.swift:280:9: Expectation failed: (events.count → 1) == 3
> > ↳ // Verify Application Installed event
> > Swift/ContiguousArrayBuffer.swift:690: Fatal error: Index out of range
> 
> flaky test?

@marandaneto just pushed a fix:

The lifecycle integration tests implicitly relied on the reachability handler to never actually ran.

  Now that the multicast fix lets the events queue actually receive its callbacks: on a .wifi connection (which CI presents as), the events queue auto-flushes
  shortly after start(). With flushAt: 3 and only 1 event in the queue at that moment, an early single-event batch hits the mock server, fulfilling its
  batchExpectationCount = 1 and returning before the remaining 2 events arrive.

PR #585, opened 2026-05-05
fix(queue): handle HTTP 413 with adaptive batch sizing

turnipdabeets:

This was added to RN for logs but iOS is missing this. There was a TODO pointing to match [posthog-android#130](https://github.com/PostHog/posthog-android/pull/130). 

On 413, halve both the batch cap and the flush threshold and retry; drop the batch once the cap reaches 1. Reductions stay reduced for the SDK lifetime (no ramp-up, matching Android/RN). Retriable set widened to `{-1, 3xx, 429, 500, 502, 503, 504}` — narrow 5xx subset matches Android's `RETRYABLE_STATUS_CODES` exactly.

The cap and flushAt reductions are kept in private state rather than mutating user-supplied `config.maxBatchSize` / `config.flushAt`.

`make test` — 9 new `PostHogQueueTest` cases covering: single 413 halve, multi-step halve→halve→drop sequence, drop at cap=1, 500/429/network-error retain, 501 (outside narrow set) pop, non-retriable 4xx pop, 2xx pop without ramp-up.

## :pencil: Checklist

### DIFF
--- a/PostHog/PostHogConfig.swift
+++ b/PostHog/PostHogConfig.swift
@@ -28,6 +28,7 @@ public typealias BeforeSendBlock = (PostHogEvent) -> PostHogEvent?
         #endif
         static let maxBatchSize: Int = 50
         static let flushIntervalSeconds: TimeInterval = 30
+        static let maxRetries: Int = 3
     }
 
     @objc(PostHogDataMode) public enum PostHogDataMode: Int {
@@ -56,6 +57,12 @@ public typealias BeforeSendBlock = (PostHogEvent) -> PostHogEvent?
     @objc public var maxQueueSize: Int = Defaults.maxQueueSize
     @objc public var maxBatchSize: Int = Defaults.maxBatchSize
     @objc public var flushIntervalSeconds: TimeInterval = Defaults.flushIntervalSeconds
+
+    /// Maximum number of consecutive flush attempts before the entire queue is
+    /// dropped to avoid infinite retries against a permanently-broken backend.
+    /// Increments on every retriable failure including HTTP 413 cap halving;
+    /// resets on a successful 2xx response. Default 3.
+    @objc public var maxRetries: Int = Defaults.maxRetries
     @objc public var dataMode: PostHogDataMode = .any
     @objc public var sendFeatureFlagEvent: Bool = true
     @objc public var preloadFeatureFlags: Bool = true
--- a/PostHog/PostHogQueue.swift
+++ b/PostHog/PostHogQueue.swift
@@ -12,9 +12,31 @@ import Foundation
 private let retriableStatusCodes: Set<Int> = [429, 500, 502, 503, 504]
 
 /// Clamps `value` to a minimum of 1. Used wherever we initialise / halve
-/// `currentBatchCap` / `currentFlushAt` so we never store a value below 1.
+/// the adaptive batch limits so we never store a value below 1.
 private func clamped(_ value: Int) -> Int {
     max(1, value)
+}
+
+/// Adaptive limits for the events queue. Both `cap` and `flushAt` are halved
+/// together on HTTP 413 and stay reduced for the SDK's lifetime — matching
+/// posthog-android. Stored privately rather than mutating
+/// `config.maxBatchSize` / `config.flushAt` so user-supplied config fields
+/// aren't silently changed.
+private struct BatchLimits {
+    var cap: Int
+    var flushAt: Int
+
+    static func initial(from config: PostHogConfig) -> BatchLimits {
+        BatchLimits(cap: clamped(config.maxBatchSize), flushAt: clamped(config.flushAt))
+    }
+
+    /// Halves both `cap` and `flushAt`, returning the new cap.
+    @discardableResult
+    mutating func halve() -> Int {
+        cap = clamped(cap / 2)
+        flushAt = clamped(flushAt / 2)
+        return cap
+    }
 }
 
 /**
@@ -52,15 +74,8 @@ class PostHogQueue {
     private let endpoint: PostHogApiEndpoint
     private let dispatchQueue: DispatchQueue
 
-    /// Both halved on HTTP 413; the batch is dropped once the cap reaches 1.
-    /// Once reduced, both stay reduced for the SDK's lifetime — matches
-    /// posthog-android. There is no ramp-back-up on healthy sends and no
-    /// reset on the poison drop. Stored privately rather than mutating
-    /// `config.maxBatchSize` / `config.flushAt` so user-supplied config
-    /// fields aren't silently changed.
-    private var currentBatchCap: Int
-    private var currentFlushAt: Int
-    private let batchSizeLock = NSLock()
+    private var batchLimits: BatchLimits
+    private let batchLimitsLock = NSLock()
 
     /// Internal, used for testing
     var depth: Int {
@@ -75,8 +90,7 @@ class PostHogQueue {
             self.api = api
             self.reachability = reachability
             self.endpoint = endpoint
-            currentBatchCap = clamped(config.maxBatchSize)
-            currentFlushAt = clamped(config.flushAt)
+            batchLimits = .initial(from: config)
 
             switch endpoint {
             case .batch:
@@ -92,8 +106,7 @@ class PostHogQueue {
             self.config = config
             self.api = api
             self.endpoint = endpoint
-            currentBatchCap = clamped(config.maxBatchSize)
-            currentFlushAt = clamped(config.flushAt)
+            batchLimits = .initial(from: config)
 
             switch endpoint {
             case .batch:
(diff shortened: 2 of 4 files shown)


PostHog/PostHogQueue.swift
```diff
@@ -40,11 +40,33 @@ class PostHogQueue {
     private let endpoint: PostHogApiEndpoint
     private let dispatchQueue: DispatchQueue
 
+    /// Both halved on HTTP 413; the batch is dropped once the cap reaches 1.
+    /// Once reduced, both stay reduced for the SDK's lifetime — matches
+    /// posthog-android. There is no ramp-back-up on healthy sends and no
+    /// reset on the poison drop. Stored privately rather than mutating
+    /// `config.maxBatchSize` / `config.flushAt` so user-supplied config
+    /// fields aren't silently changed.
```
marandaneto: i like that, i think we should do the same on android instead of mutating config.values

PostHog/PostHogQueue.swift
```diff
@@ -53,6 +75,8 @@ class PostHogQueue {
             self.api = api
             self.reachability = reachability
             self.endpoint = endpoint
+            currentBatchCap = max(1, config.maxBatchSize)
```
marandaneto: we can probably extract max(1, config.maxBatchSize) to a function that gets the 2nd value as parameter, its repeated a few times

PostHog/PostHogQueue.swift
```diff
@@ -99,24 +125,54 @@ class PostHogQueue {
...
+        // Network error (-1), 3xx redirect, or transient server error: pause and
+        // retry the same batch. The 5xx subset is intentionally narrow to match
+        // posthog-android's RETRYABLE_STATUS_CODES.
+        let retriable = statusCode == -1
+            || (300 ... 399 ~= statusCode)
+            || statusCode == 429
+            || statusCode == 500
+            || statusCode == 502
+            || statusCode == 503
+            || statusCode == 504
```
marandaneto: we could extract this to a method or create a RETRYABLE_STATUS_CODES like android so its easier to read this method

PostHog/PostHogQueue.swift
```diff
@@ -40,11 +40,33 @@ class PostHogQueue {
...
+    private let batchSizeLock = NSLock()
+
     /// Internal, used for testing
     var depth: Int {
         fileQueue.depth
     }
 
+    /// Internal, used for testing — exposes the adaptive batch cap so 413
+    /// halving and poison-drop tests can assert the cap value.
+    var currentBatchCapForTesting: Int {
```
ioannisj: would rather keep these in #if TESTING in an extension to be consistent

PostHog/PostHogQueue.swift
```diff
@@ -40,11 +40,33 @@ class PostHogQueue {
...
+        batchSizeLock.withLock { currentBatchCap }
+    }
+
+    /// Internal, used for testing — exposes the adaptive flush threshold so
+    /// 413 tests can assert it was halved alongside the batch cap.
+    var currentFlushAtForTesting: Int {
```
ioannisj: Same as L60

PostHog/PostHogQueue.swift
```diff
@@ -99,24 +125,54 @@ class PostHogQueue {
...
+        }
+
+        // 413 Payload Too Large: halve both the batch cap and the flush
+        // threshold and retry without popping. Once the cap reaches 1, drop
+        // the batch — we can't shrink any further, so retrying is futile.
+        // Matches posthog-android's `deleteFilesIfAPIError`.
+        if statusCode == 413 {
+            let halvedCap: Int? = batchSizeLock.withLock {
+                guard currentBatchCap > 1 else { return nil }
+                currentBatchCap = max(1, currentBatchCap / 2)
```
ioannisj: Consider an empty queue and the defaults (flushAt: 20, batchSize: 50). After capturing 20 events we flush, and we hit a 413. We'll half the currentBatchCap to 25, and on next run we take(25) which will return the same 20 events that have already failed? This will fail again and we'll half again so eventually things will sort out, but we could avoid these failed calls? 

Just wondering if we should be considering `payload.events.count` in this logic. Maybe with `currentBatchCap = max(1, min(currentBatchCap, payload.events.count) / 2)`. So in the scenario above, this will take(10) next time which seems like better progress. I think Android implementation is the same btw

PostHog/PostHogQueue.swift
```diff
@@ -99,24 +125,54 @@ class PostHogQueue {
...
+                currentFlushAt = max(1, currentFlushAt / 2)
+                return currentBatchCap
+            }
+            if let halvedCap {
+                hedgeLog("Queue: HTTP 413, halved batch cap to \(halvedCap)")
+                retryCount = 0
+                payload.completion(false)
```
ioannisj, line 185: I think Android implementation pauses/backs off here? Tbh it makes more sense to shrink and retry immediately

PostHog/PostHogQueue.swift
```diff
@@ -99,24 +125,54 @@ class PostHogQueue {
...
+            || statusCode == 500
+            || statusCode == 502
+            || statusCode == 503
+            || statusCode == 504
+
+        if retriable {
             retryCount += 1
             let delay = min(retryCount * retryDelay, maxRetryDelay)
             pauseFor(seconds: delay)
             hedgeLog("Pausing queue consumption for \(delay) seconds due to \(retryCount) API failure(s).")
```
ioannisj, line 162: Android has a check on maxRetries as well https://github.com/PostHog/posthog-android/blob/097c3fdb645358fad7bc64548d73619491b7b35e/posthog/src/main/java/com/posthog/internal/PostHogQueue.kt#L208-L212 which drops events. It seems we are missing this on iOS? Could be a good PR to add or document a TODO?

PostHog/PostHogQueue.swift
```diff
@@ -40,11 +40,33 @@ class PostHogQueue {
...
     private let dispatchQueue: DispatchQueue
 
+    /// Both halved on HTTP 413; the batch is dropped once the cap reaches 1.
+    /// Once reduced, both stay reduced for the SDK's lifetime — matches
+    /// posthog-android. There is no ramp-back-up on healthy sends and no
+    /// reset on the poison drop. Stored privately rather than mutating
+    /// `config.maxBatchSize` / `config.flushAt` so user-supplied config
+    /// fields aren't silently changed.
+    private var currentBatchCap: Int
+    private var currentFlushAt: Int
```
ioannisj: nit: I would move these into a small struct (i.e private struct BatchLimits) just to communicate better that they are moving together

PostHog/PostHogQueue.swift
turnipdabeets: extracted

PostHog/PostHogQueue.swift
turnipdabeets: also extracted

PostHog/PostHogQueue.swift
turnipdabeets: moved to #if TESTING

PostHog/PostHogQueue.swift
turnipdabeets: moved

PostHog/PostHogQueue.swift
turnipdabeets: updated! It looks like Android might not actually do this so I'll add a fix for Android in another PR when I address mutating config.values for @marandaneto too.

PostHog/PostHogQueue.swift
turnipdabeets: moved together in a struct

PostHog/PostHogQueue.swift
turnipdabeets: added a max 3 retry

PostHog/PostHogConfig.swift
```diff
@@ -56,6 +57,12 @@ public typealias BeforeSendBlock = (PostHogEvent) -> PostHogEvent?
     @objc public var maxQueueSize: Int = Defaults.maxQueueSize
     @objc public var maxBatchSize: Int = Defaults.maxBatchSize
     @objc public var flushIntervalSeconds: TimeInterval = Defaults.flushIntervalSeconds
+
+    /// Maximum number of consecutive flush attempts before the entire queue is
+    /// dropped to avoid infinite retries against a permanently-broken backend.
+    /// Increments on every retriable failure including HTTP 413 cap halving;
+    /// resets on a successful 2xx response. Default 3.
+    @objc public var maxRetries: Int = Defaults.maxRetries
```
turnipdabeets, line 65: Android makes this public too, but I could change that if we think we don't want this configurable.

ioannisj: I think we should keep it and also update the [docs](https://posthog.com/docs/libraries/ios/configuration#all-configuration-options)

ioannisj (approved):

PostHog/PostHogConfig.swift
turnipdabeets: Added https://github.com/PostHog/posthog.com/pull/16740

PR #586, opened 2026-05-05
fix: clean up CocoaPods validation warnings

marandaneto:

Clean up the CocoaPods validation warnings we can address directly in SDK code/config:

- Keep the deprecated `propertiesSanitizer` public API, but read it internally through a non-deprecated backing accessor so SDK builds do not warn on their own deprecated API.
- Explicitly discard the `removeValue(forKey:)` result inside the URLSession swizzler lock closure.
- Remove the duplicate `-lc++` linker flag from the podspec while keeping `s.libraries = 'c++'`.

Possible follow-ups from the same CI log:

- Change the private PHPLCrashReporter module map from an `umbrella header` to a regular `header` to avoid umbrella-header warnings without exposing vendored internals.
- Suppress or otherwise handle `-Wshorten-64-to-32` warnings from vendored `protobuf-c` separately.

- `make format`
- `make test`
- `make lint`

## :pencil: Checklist

## If releasing new changes

### DIFF
--- a/PostHog/PostHogConfig.swift
+++ b/PostHog/PostHogConfig.swift
@@ -110,10 +110,18 @@ public typealias BeforeSendBlock = (PostHogEvent) -> PostHogEvent?
     /// Defaults to false.
     @objc public var reuseAnonymousId: Bool = false
 
+    private var _propertiesSanitizer: PostHogPropertiesSanitizer?
+    var legacyPropertiesSanitizer: PostHogPropertiesSanitizer? {
+        _propertiesSanitizer
+    }
+
     /// Hook that allows to sanitize the event properties
     /// The hook is called before the event is cached or sent over the wire
     @available(*, deprecated, message: "Use beforeSend instead")
-    @objc public var propertiesSanitizer: PostHogPropertiesSanitizer?
+    @objc public var propertiesSanitizer: PostHogPropertiesSanitizer? {
+        get { _propertiesSanitizer }
+        set { _propertiesSanitizer = newValue }
+    }
     /// Determines the behavior for processing user profiles.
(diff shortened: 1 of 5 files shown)


PostHog/PostHogConfig.swift
```diff
@@ -110,10 +110,18 @@ public typealias BeforeSendBlock = (PostHogEvent) -> PostHogEvent?
...
+    private var _propertiesSanitizer: PostHogPropertiesSanitizer?
+    var legacyPropertiesSanitizer: PostHogPropertiesSanitizer? {
+        _propertiesSanitizer
+    }
+
     /// Hook that allows to sanitize the event properties
     /// The hook is called before the event is cached or sent over the wire
     @available(*, deprecated, message: "Use beforeSend instead")
-    @objc public var propertiesSanitizer: PostHogPropertiesSanitizer?
+    @objc public var propertiesSanitizer: PostHogPropertiesSanitizer? {
```
marandaneto, line 121: otherwise we always trigger the deprecated warning ourselves

marandaneto:
@ioannisj about the 'Possible follow-ups from the same CI log:'
your call here, its a long list of warnings, not sure if they should be fixed or silenced

marandaneto:
found by https://posthog.slack.com/archives/C0A3UEVDDNF/p1777991448692979

dustinbyrne (approved):

ioannisj:
> @ioannisj about the 'Possible follow-ups from the same CI log:' your call here, its a long list of warnings, not sure if they should be fixed or silenced

Not sure they can be easily fixed. I'll try the first suggestion above otherwise I'll see how we can silence these

PR #588, opened 2026-05-05
ci: check pod version

ioannisj:

CocoaPods trunk can successfully publish a version but still return a transient/internal server error to CI. When the release job is rerun, pod trunk push then fails with a
 duplicate version error, even though the pod is already available.

 This updates the release workflow to make CocoaPods publishing idempotent:

 - Checks whether the release version already exists on CocoaPods before publishing.
 - Skips pod trunk push and succeeds if the version is already present.
 - If make releaseCocoaPods fails, captures the exit code, logs a warning, then checks CocoaPods again.
 - Treats the publish as successful if the version appeared despite the failure.
 - Still fails the job if the version is not available after the failed publish.

- Validated the CocoaPods version check locally with:
     - 3.57.4 → detected as existing
     - fake version → detected as missing

## :pencil: Checklist

## If releasing new changes

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -280,8 +280,54 @@ jobs:
         with:
           xcode-version: '16.2'
 
+      - name: Check if CocoaPods version already exists
+        id: check-cocoapods-version
+        env:
+          RELEASE_VERSION: ${{ needs.create-github-release.outputs.version }}
+        run: |
+          set -euo pipefail
+
+          version_exists_on_cocoapods() {
+            pod trunk info PostHog --no-ansi | grep -F -- "- $1 (" > /dev/null
+          }
+
+          if version_exists_on_cocoapods "$RELEASE_VERSION"; then
+            echo "PostHog ($RELEASE_VERSION) already exists on CocoaPods. Skipping publish."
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -280,8 +280,54 @@ jobs:
...
           xcode-version: '16.2'
 
+      - name: Check if CocoaPods version already exists
+        id: check-cocoapods-version
+        env:
+          RELEASE_VERSION: ${{ needs.create-github-release.outputs.version }}
+        run: |
+          set -euo pipefail
+
+          version_exists_on_cocoapods() {
```
turnipdabeets, line 290: I think this is defined twice, we could possibly DRY

.github/workflows/release.yml
```diff
@@ -280,8 +280,54 @@ jobs:
...
+            pod trunk info PostHog --no-ansi | grep -F -- "- $1 (" > /dev/null
```
turnipdabeets, line 291: nit: might match both 13.57.4 and 3.57.4
```suggestion
            pod trunk info PostHog --no-ansi | grep -E -- "^[[:space:]]*- $(printf '%s' "$1" 
```

PR #590, opened 2026-05-05
feat(logs): add storage + queue infrastructure for the logs feature

turnipdabeets:

Storage + queue infrastructure for the new logs feature. Wire format and shape match the JS / RN SDKs so the backend treats records identically across SDKs.

Two PRs:

- **PR1 (this)** — disk persistence, OTLP/JSON serialization, `/i/v1/logs` send, reachability, retry/backoff, HTTP 413 adaptive batching, `maxRetries` drop, rate cap, `beforeSend`, **and a generic-queue refactor**. The public capture API (`captureLog`, `logger.info/...`, `flushLogs()`) is _not_ in this PR.
- **PR2** (will be merged into this branch before merge to `main`) — public capture surface, screen-name caching, lifecycle wiring.

**Why merge before shipping:** PR1 already exposes `@objc public` config types (`PostHogLogsConfig`, `PostHogLogRecord`, `PostHogLogSeverity`); shipping PR1 alone would expose autocomplete-visible knobs with no way to actually emit a log.

## Generic queue refactor (the big change)

Reviewer feedback: `PostHogQueue` already serves `/batch` (events) and `/snapshot` (replay), so a third dedicated `PostHogLogsQueue` would duplicate ~80% of the queue infrastructure. So:

- **`PostHogQueue<Record>`** — generic class owning all queue infrastructure (file persistence, reachability, flush timer, retry/backoff, 413 halving, `maxRetries` drop). Was concrete on `PostHogEvent`.
- **`QueueEndpoint<Record>`** — struct encapsulating per-wire spec: storage key, codec, sender, retriable status set, redirect-as-retriable flag.
- Three factories: `.batch(api:)`, `.snapshot(api:)`, `.logs(api:resourceAttributes:)`.
- **`PostHogReplayQueue`** keeps its wrapper shape — composes `PostHogQueue<PostHogEvent>` with `.snapshot`.
- **`PostHogLogsQueue`** becomes a thin (~200 LOC) wrapper around `PostHogQueue<PostHogLogRecord>` holding only rate cap + `beforeSend`.

Net: **+197 / −449 LOC** (unification removes more than it adds). All existing events / replay / logs tests pass without modification — refactor is behavior-preserving.

## Files

| File | Role |
| --- | --- |
| `PostHog/QueueEndpoint.swift` + `+Factories.swift` | Generic-queue spec + three concrete factories. |
| `PostHog/PostHogQueue.swift` | Now generic over `Record`. Hosts shared helpers `halveBatchCap` / `retryCountExceeded`. |
| `PostHog/Logs/PostHogLogRecord.swift` | `@objc final class` record. `traceFlags: NSNumber?` for ObjC bridging + `traceFlagsValue: Int?` Swift sugar. |
| `PostHog/Logs/PostHogLogSeverity.swift` | 6-level enum with OTLP `severityNumber` mapping. |
| `PostHog/Logs/PostHogLogsConfig.swift` | `@objc` config: flush interval, batch / buffer sizes, service metadata, rate cap, `beforeSend` chain. |
| `PostHog/Logs/PostHogLogsOTLP.swift` | Pure OTLP/JSON serialization. Nested dicts encoded as `kvlistValue` (spec-correct; JS uses `JSON.stringify` — both stringified to flat ClickHouse cells today, see in-file comment). |
| `PostHog/Logs/PostHogLogsQueue.swift` | Thin wrapper around `PostHogQueue<PostHogLogRecord>`. |
| `PostHog/PostHogApi.swift` | New `logs(payload:completion:)`. |
| `PostHog/PostHogStorage.swift` | New `.logsQueue` storage key. |
| `PostHog/PostHogConfig.swift` | New `@objc public var logs: PostHogLogsConfig`. |
| `PostHog/PostHogSDK.swift` | Lifecycle wiring (constructed, started, flushed on background, stopped, nilled). |

## :green_heart: Tests

`make test` — **492 tests in 104 suites pass.**

`PostHogTests/PostHogLogsQueueTest.swift` (Swift Testing) covers persistence + 413 (halve, poison drop, max-retries drop, no ramp on 2xx, `actualBatchSize`-bounded halving, multi-cycle drainage proves poison-drop isn't a retry), 5xx / 4xx, rate cap, `beforeSend`, concurrency, reachability, OTLP wire format incl. `traceFlags` → `flags`, full-record disk codec round-trip.

`PostHogTests/PostHogQueueTest.swift` (Quick/Nimble) extended to verify events cap stays put after `dropAll` — pins the policy wiring after the generic refactor.

Threading audit: no recursive locking, no force-unwraps in production code, `[weak self]` everywhere. Fixed: `flushIfOverThreshold` previously ran disk peek + JSON decode synchronously on the caller's thread; now dispatches to the queue's serial dispatch queue.

## :pencil: Checklist

## If releasing new changes

- [x] `.changeset/logs-feature.md` describes the combined feature; PR2 must merge in before merge to `main`.


PostHog/Logs/PostHogLogRecord.swift
```diff
@@ -0,0 +1,163 @@
...
+        self.observedTimeUnixNano = observedTimeUnixNano ?? now
+        self.distinctId = distinctId
+        self.sessionId = sessionId
+        self.screenName = screenName
+        self.appState = appState
+        self.featureFlagKeys = featureFlagKeys
+    }
+    /// Current wall clock as a uint64 nanosecond string. OTLP/JSON encodes
+    /// uint64 as a string because JSON numbers cannot represent uint64 precisely.
+    public static func nanosNow() -> String {
```
marandaneto: i think this could be part of the date utils if not yet

PostHog/Logs/PostHogLogRecord.swift
```diff
@@ -0,0 +1,163 @@
...
+        let secs = Date().timeIntervalSince1970
```
marandaneto: we should use the date utils `now` here to make this testable

PostHog/Logs/PostHogLogRecord.swift
```diff
@@ -0,0 +1,163 @@
...
+        if let traceFlags { json["traceFlags"] = traceFlags }
+        if let distinctId { json["distinctId"] = distinctId }
+        if let sessionId { json["sessionId"] = sessionId }
+        if let screenName { json["screenName"] = screenName }
+        if let appState { json["appState"] = appState }
+        return json
+    }
+
+    static func fromStorageJSON(_ data: Data) -> PostHogLogRecord? {
+        guard let json = try? JSONSerialization.jsonObject(with: data, options: []) as? [String: Any] else {
```
marandaneto: in some case we use `allowFragments` as options, not sure we need it here tho

PostHog/Logs/PostHogLogRecord.swift
```diff
@@ -0,0 +1,163 @@
...
+            return nil
+        }
```
marandaneto: this can be an utils function and reuse across the sdk, i think we do the very same thing many times over

PostHog/Logs/PostHogLogRecord.swift
```diff
@@ -0,0 +1,163 @@
...
+
+/// A single log entry queued for delivery to PostHog.
+///
+/// Records are produced by `captureLog` (or `logger.<level>`), persisted to disk
+/// via `PostHogLogsQueue`, and serialized as OpenTelemetry log records on the wire.
+///
+/// `PostHogLogRecord` is a value type so user-supplied `beforeSend` callbacks can
+/// mutate a local copy and return a transformed instance without affecting the
+/// queued record before it has been processed.
+public struct PostHogLogRecord {
```
marandaneto: do we need `@objc` here?

PostHog/Logs/PostHogLogsConfig.swift
```diff
@@ -0,0 +1,69 @@
...
+
+/// Configuration for the PostHog logs subsystem. Mutate fields on `config.logs`
+/// before calling `PostHogSDK.setup(_:)`.
+@objc(PostHogLogsConfig) public final class PostHogLogsConfig: NSObject {
+    enum Defaults {
+        static let flushIntervalSeconds: TimeInterval = 30
+        static let maxBufferSize: Int = 100
+        static let maxBatchSize: Int = 50
+        static let rateCapMaxLogs: Int = 500
+        static let rateCapWindowSeconds: TimeInterval = 10
```
marandaneto: we should share the default values (use the same consts) with the events queue unless theres a reason to divert

PostHog/Logs/PostHogLogSeverity.swift
```diff
@@ -0,0 +1,71 @@
...
+
+    /// OTLP `severityText` (uppercase identifier).
+    public var severityText: String {
+        switch self {
+        case .trace: return "TRACE"
+        case .debug: return "DEBUG"
+        case .info: return "INFO"
+        case .warn: return "WARN"
+        case .error: return "ERROR"
+        case .fatal: return "FATAL"
```
marandaneto: should we move the strings to consts and then just do .uppercase() | .lowercase() for them? so they dont drift at some point, we repeat them 3 times already

PostHog/Logs/PostHogLogsQueue.swift
```diff
@@ -0,0 +1,497 @@
...
+/// (the multicast hooks). Pauses on no network and on cellular when
+/// `dataMode == .wifi`; auto-flushes on WiFi reconnect.
+///
+/// **Thread safety**: `add(_:)` and `flush()` are callable from any thread and
+/// return immediately. In-memory state is guarded by `stateLock`; on-disk state
+/// is owned by `PostHogFileBackedQueue` whose own internal lock makes its
+/// operations safe to call from `dispatchQueue` (write path) and the URLSession
+/// completion queue (`pop` path) concurrently. `stateLock` and `timerLock` are
+/// never held simultaneously.
+class PostHogLogsQueue {
```
marandaneto: we already use PostHogQueue for 2 things, flushing events and recordings
is there a reason to have its own queue now? the logic is super similar, feels a DRY type of thing

PostHog/PostHogApi.swift
```diff
@@ -186,6 +186,68 @@ class PostHogApi {
...
+        guard let url else {
+            hedgeLog("Malformed logs URL error.")
+            return completion(PostHogBatchUploadInfo(statusCode: nil, error: nil))
+        }
+
+        let sessionConfig = sessionConfig()
+        var headers = sessionConfig.httpAdditionalHeaders ?? [:]
+        headers["Accept-Encoding"] = "gzip"
+        headers["Content-Encoding"] = "gzip"
+        sessionConfig.httpAdditionalHeaders = headers
```
marandaneto: i think we duplicate this for the 3 time, lets extract and reuse it

PostHog/PostHogApi.swift
```diff
@@ -186,6 +186,68 @@ class PostHogApi {
...
+
+        let request = getURLRequest(url)
+
+        guard let data = try? JSONSerialization.data(withJSONObject: payload) else {
+            hedgeLog("Error parsing the logs body")
+            return completion(PostHogBatchUploadInfo(statusCode: nil, error: nil))
```
marandaneto: time to rename PostHogBatchUploadInfo since its not only about batch anymore

marandaneto:
one point to consider is that now we have /batch, /s (recordings), /logs and /flags
and they all could happen at the same time i guess? is that ok? maybe more of a concern for all but flags, since they load enough data into memory from disk, flush to the network etc, maybe not a blocker for this PR but maybe we should consider making this "sequential" so we only use 1 connection pool
iirc ios limited this to 5 connections at once and just here is 3

PostHog/Logs/PostHogLogSeverity.swift
```diff
@@ -0,0 +1,71 @@
...
+//  PostHog
+//
+
+import Foundation
+
+/// Severity level for a captured log record. Maps to OpenTelemetry severity
+/// numbers (TRACE=1, DEBUG=5, INFO=9, WARN=13, ERROR=17, FATAL=21).
+///
+/// Named `PostHogLogSeverity` rather than `PostHogLogLevel` to avoid colliding
+/// with the iOS-only `PostHogLogLevel` used by the session-replay console
```
ioannisj: We could probably reuse so that we minimize the number of types we expose but I'll leave this up to you

PostHog/Logs/PostHogLogSeverity.swift
```diff
@@ -0,0 +1,71 @@
...
+        case .info: return "INFO"
+        case .warn: return "WARN"
+        case .error: return "ERROR"
+        case .fatal: return "FATAL"
+        }
+    }
+
+    /// Parse a severity from its lowercase string form. Returns `nil` for unknown values.
+    public static func from(name: String) -> PostHogLogSeverity? {
+        switch name.lowercased() {
```
ioannisj: probably trimming white space as well

PostHog/Logs/PostHogLogsConfig.swift
```diff
@@ -0,0 +1,69 @@
...
+    /// Maximum number of records accepted per `rateCapWindowSeconds`. Set to
+    /// `0` to disable.
+    @objc public var rateCapMaxLogs: Int = Defaults.rateCapMaxLogs
+
+    /// Tumbling window in seconds used by the rate cap.
+    @objc public var rateCapWindowSeconds: TimeInterval = Defaults.rateCapWindowSeconds
+
+    /// Optional callback invoked before the record is enqueued. Returning `nil`
+    /// drops the record; returning a record with an empty `body` also drops it.
+    public var beforeSend: PostHogBeforeSendLogBlock?
```
ioannisj: This should work for objc interop. We should probably handle it the same way we handle beforeSend in `PostHogConfig`

PostHog/PostHogQueue.swift
```diff
@@ -17,6 +17,26 @@ private func clamped(_ value: Int) -> Int {
...
 }
 
+// MARK: - Shared queue policy helpers (used by PostHogQueue and PostHogLogsQueue)
+
+/// Halves a batch cap, bounded by the actual batch size that triggered the
+/// halving. The `min(cap, actualBatchSize)` bound avoids wasting halvings on
+/// a partial batch — queue depth was below the cap — when the server only
+/// saw `actualBatchSize` records anyway. Result is clamped to `>= 1`.
+/// Mirrors posthog-android's `BatchLimits.halve` and posthog-js-lite.
+func halveBatchCap(_ cap: Int, actualBatchSize: Int) -> Int {
```
ioannisj: nit: Not super fun of free functions 

I think we could reuse PostHogQueue btw and extend [PostHogApiEndpoint](https://github.com/PostHog/posthog-ios/blob/98a1fa33e39f66c99de686bc8adc01f9bfa719eb/PostHog/PostHogQueue.swift#L59)? In that case I would move these back to PostHogQueue

.changeset/logs-feature.md
```diff
@@ -0,0 +1,13 @@
+---
+"posthog-ios": minor
```
ioannisj: this should probably come with next PR?

ioannisj (commented):
Did a first read. Biggest comment is on reusing PostHogQueue. I believe the main blocker was PostHogEvent vs PostHogLogRecord which I think we should probably make them both depend on an abstraction, then make PostHogQueue generic on that?

PostHog/Logs/PostHogLogRecord.swift
turnipdabeets: refactored

PostHog/Logs/PostHogLogsConfig.swift
turnipdabeets: sharing

PostHog/Logs/PostHogLogSeverity.swift
turnipdabeets: good idea, done

PostHog/PostHogApi.swift
turnipdabeets: renamed

PostHog/Logs/PostHogLogSeverity.swift
turnipdabeets: I thought about keeping as separate since they have different uses and didn't want to extent PostHogLogLevel but it might be useful to reuse on auto capture logs so will join.

PostHog/Logs/PostHogLogSeverity.swift
turnipdabeets: added

PostHog/Logs/PostHogLogsConfig.swift
turnipdabeets: updated and refactored some of the shared beforeSend between log and events

PostHog/PostHogQueue.swift
turnipdabeets: good call, moved

.changeset/logs-feature.md
turnipdabeets: I was thinking it shouldn't matter if I'm merging the other PR into this one, but I'll move it to the next one anyway

PostHog/Logs/PostHogLogRecord.swift
turnipdabeets: added

PostHog/Logs/PostHogLogsQueue.swift
turnipdabeets: refactored

PostHog/PostHogApi.swift
turnipdabeets: extracted!

turnipdabeets:
> one point to consider is that now we have /batch, /s (recordings), /logs and /flags
> and they all could happen at the same time i guess? is that ok? maybe more of a concern for all but flags, since they load enough data into memory from disk, flush to the network etc, maybe not a blocker for this PR but maybe we should consider making this "sequential" so we only use 1 connection pool
> iirc ios limited this to 5 connections at once and just here is 3

Yea, I think per-request URLSession is an anti-pattern so we should consider refactoring this in a separate PR

PostHog/Logs/PostHogLogRecord.swift
```diff
@@ -0,0 +1,172 @@
...
+    /// screen view tracking is enabled.
+    @objc public var screenName: String?
+
+    /// `"foreground"` or `"background"` at capture time.
+    @objc public var appState: String?
+
+    /// Keys of feature flags that were active at capture time. Useful for
+    /// correlating log records with experiment cohorts. Empty when no flags
+    /// are loaded.
+    @objc public var featureFlagKeys: [String]
```
dustinbyrne: Probably obvious, but just raising caution that all of these public fields will become part of the SDKs public API via the `beforeSend` hook.

`timeUnixNano`, `observedTimeUnixNano`, and `featureFlagKeys` in particular feel internal enough that we should decide intentionally whether to expose them

`traceFlagsValue` as well?

dustinbyrne: There's a lot going on here in this PR so I'm just going to submit this one comment for now. I'm going to need some more time to wrap my head around this.

PostHog/Replay/Plugins/Console Logs/PostHogLogLevel.swift
```diff
@@ -5,18 +5,69 @@
...
+/// Severity level for a captured log entry. Used by both the session-replay
+/// console-log subsystem (iOS-only) and the cross-platform logs feature.
+///
+/// Cases are ordered by severity so `level.rawValue` comparisons are
+/// meaningful: `entry.level.rawValue >= minLevel.rawValue` keeps records at
+/// or above the threshold.
+///
+/// Maps to OpenTelemetry severity numbers (`TRACE=1`, `DEBUG=5`, `INFO=9`,
+/// `WARN=13`, `ERROR=17`, `FATAL=21`) for the OTLP wire format.
+@objc(PostHogLogLevel) public enum PostHogLogLevel: Int, CaseIterable {
```
ioannisj: This was my suggestion but now that I think about it, raw values here will change from previous implementation (e.g error was implicitly assigned a value of 2). No idea if people from objc would depend on raw values or not, but I highly doubt it. 

We could, however, just implicitly assign values here to keep old info, warn, error unchanged

PostHog/Logs/PostHogLogsConfig.swift
```diff
@@ -0,0 +1,96 @@
...
+
+    public func setBeforeSend(_ blocks: [PostHogBeforeSendLogBlock]) {
+        beforeSend.set(blocks)
+    }
+
+    public func setBeforeSend(_ blocks: PostHogBeforeSendLogBlock...) {
+        setBeforeSend(blocks)
+    }
+
+    @available(*, unavailable, message: "Use setBeforeSend(_ blocks: PostHogBeforeSendLogBlock...) instead")
```
ioannisj: Are we marking unavailable here only for Swift users? Will ObjC users be able to see this (since it's the only API they can use)?

PostHog/PostHogQueue.swift
```diff
@@ -53,20 +47,34 @@ private struct BatchLimits {
...
-        case batch
-        case snapshot
+class PostHogQueue<Record> {
+    /// True when `retryCount` has exceeded `maxRetries` after just being
+    /// incremented for the current attempt. Comparison is `>` (not `>=`) so the
+    /// configured value is the count of *retries* allowed before the drop fires
+    /// — with the default of 3 you get attempts 1–3 retried, attempt 4 drops.
+    private static func retryCountExceeded(_ retryCount: Int, maxRetries: Int) -> Bool {
+        retryCount > maxRetries
     }
```
dustinbyrne: nit: this feels a little excessive for what looks like a trivial helper

PostHog/PostHogQueue.swift
```diff
@@ -7,39 +7,33 @@
...
+    /// the server only saw `actualBatchSize` records anyway. `flushAt` is
+    /// clamped to the new `cap` so we never buffer more events than a single
+    /// batch can drain. Mirrors posthog-android's `BatchLimits.halve`.
     @discardableResult
     mutating func halve(actualBatchSize: Int) -> Int {
-        cap = clamped(min(cap, actualBatchSize) / 2)
+        cap = max(1, min(cap, actualBatchSize) / 2)
         flushAt = clamped(min(flushAt / 2, cap))
         return cap
     }
```
dustinbyrne, line 37: nit: changes here are effectively a no-op

PostHog/Replay/Plugins/Console Logs/PostHogLogLevel.swift
turnipdabeets: good idea, reordered to preserve that

PostHog/Logs/PostHogLogRecord.swift
turnipdabeets: good call out, I downgraded a bunch to be internal

PostHog/QueueEndpoint.swift
```diff
@@ -0,0 +1,63 @@
...
+    /// passes its `handleResult` continuation as `completion`.
+    let send: ([Record], @escaping (PostHogUploadInfo) -> Void) -> Void
+
+    // MARK: Retry policy
+
+    /// Status codes that trigger an exponential-backoff retry of the same
+    /// batch. Events: `[429, 500, 502, 503, 504]`. Logs: `[408, 429, 500..599]`.
+    let retriableStatusCodes: Set<Int>
+    /// Whether 3xx redirects are retriable. Events: yes. Logs: no.
+    let redirectIsRetriable: Bool
```
dustinbyrne: nit: slightly awkward split of responsibilities on these properties, maybe consider something like

```swift
let isRetriableStatusCode: (Int) -> Bool
```
 or
```swift
let shouldRetry: (PostHogUploadInfo) -> Bool
```

PostHog/PostHogQueue.swift
```diff
@@ -236,17 +240,20 @@ class PostHogQueue {
...
-                    // Always trigger a flush when we are on wifi
+                    // Always trigger a flush when we are on wifi. The
+                    // `isFlushing` snapshot is only an optimisation —
+                    // `flush()` re-checks under the lock.
                     if reachability.connection == .wifi {
-                        if !self.isFlushing {
+                        let busy = self.isFlushingLock.withLock { self.isFlushing }
+                        if !busy {
                             self.flush()
                         }
```
dustinbyrne: looks like toctou, and we're doing this in a few places before calling `flush`

`flush` also checks this lock, so this seems unnecessary

PostHog/Logs/PostHogLogsQueue.swift
```diff
@@ -0,0 +1,189 @@
...
+
+    // MARK: Rate-cap state (guarded by stateLock)
+
+    private let stateLock = NSLock()
+    private var rateCapWindowStart: Date?
+    private var rateCapCount: Int = 0
+    /// Flips to `true` the first time we drop a record in the current window;
+    /// reset when the window rolls. Used to emit one warning per window
+    /// instead of one per dropped record.
+    private var rateCapDropWarned = false
```
dustinbyrne: It looks like we're wrapping the `PostHogQueue` specifically to add record volume / time window rate limiting on top?

If this is the case, might be worth moving this into the generic `PostHogQueue` and swapping this out for a raw `PostHogQueue<PostHogLogRecord>` in the future. If I'm remembering correctly, there are plans to implement similar volume limits to event ingestion as well.

Just raising this, not a blocking item.

dustinbyrne (commented):
Overall, looking pretty good. I read through everything here. There are a lot of changes in this PR, so I had to have an agent break this up into reviewable commits locally. This was probably pushing into 2 PR territory, but I understand there was a fair amount of feedback to get here.

I'd recommend giving your changes a once-over just to clean up any obvious/extraneous artifacts of the LLM. E.g., cases where we're re-shuffling method docs unnecessarily, or heavily documenting straightforward internal logic - maybe it was just me, but it felt like there was a fair amount of noise.

dustinbyrne:
re the squash commit, did anything notable change that's worth taking a look at?

PostHog/Logs/PostHogLogsQueue.swift
turnipdabeets: Will do this as a follow-up!

PostHog/Logs/PostHogLogsConfig.swift
turnipdabeets: Improved ObjC ergonomics and fixed this

PostHog/Replay/Plugins/Console Logs/PostHogLogLevel.swift
turnipdabeets: I ended up reverting to 2 separate types. There were other places that adding extra cases made it more complicated.

PostHog/Logs/PostHogLogRecord.swift
marandaneto: looks like not?

marandaneto: or is that because its inherited from `NSObject`?

turnipdabeets: This is now the strut with internal properties only. Check out `PostHogMutableLogRecord` since `PostHogLogRecord` is no longer public

turnipdabeets: I can probably drop the `NSObject`. It's an artifact from earlier.

PostHog/Logs/PostHogMutableLogRecord.swift
```diff
@@ -0,0 +1,94 @@
...
+//  PostHog
+//
+
+import Foundation
+
+/// Mutable view over a captured log record handed to `beforeSend` callbacks.
+///
+/// Exposes only the redaction-safe fields. Wire-format internals
+/// (`timeUnixNano`, `observedTimeUnixNano`, OTLP encoding) are deliberately
+/// kept off this surface so callbacks cannot corrupt them.
```
marandaneto: could we use a single class but the internal fields are internal and immutable?

marandaneto (approved):
approving to unblock but ideally we figure [this out](https://github.com/PostHog/posthog-ios/pull/590#discussion_r3204574658) first since after merged/published we cannot remove anymore

marandaneto:
> Overall, looking pretty good. I read through everything here. There are a lot of changes in this PR, so I had to have an agent break this up into reviewable commits locally. This was probably pushing into 2 PR territory, but I understand there was a fair amount of feedback to get here.
> 
> I'd recommend giving your changes a once-over just to clean up any obvious/extraneous artifacts of the LLM. E.g., cases where we're re-shuffling method docs unnecessarily, or heavily documenting straightforward internal logic - maybe it was just me, but it felt like there was a fair amount of noise.

yep, i agree, i had similar feedback [here](https://github.com/PostHog/posthog-js/pull/3480#issuecomment-4336257135)

PostHog/Logs/PostHogMutableLogRecord.swift
turnipdabeets: Updated with consolidated. Will hold off on merging this until https://github.com/PostHog/posthog-ios/pull/592 merges in this branch.

PostHog/Logs/PostHogLogsQueue.swift
turnipdabeets: @dustinbyrne merged into this PR as https://github.com/PostHog/posthog-ios/pull/596

PR #591, opened 2026-05-06
fix: prefix libwebp symbols

### LINKED ISSUE #589: Conflict with libwebp when integrating PostHog iOS via SPM alongside other WebP-based libraries
Hi team,

I’m running into a linker conflict when integrating the PostHog iOS SDK via Swift Package Manager together with other libraries that also depend on libwebp (e.g. Kingfisher + KingfisherWebP).

It appears that PostHog includes a vendored and modified version of libwebp (phlibwebp), which leads to duplicate symbol conflicts at link time when another dependency also links against libwebp.

⸻

Environment

* Integration method: Swift Package Manager (SPM)
* PostHog iOS SDK: 3.57.5
* Other dependency: Kingfisher + KingfisherWebP (via SPM)
* Platform: iOS / tvOS
* Xcode: 26.4.1 

⸻

Error

Example linker error:

duplicate symbol '_WebPDecodeRGBA' in:
    libwebp.a
    phlibwebp.a

⸻

Root Cause

From what I can tell:

* PostHog vendors a modified version of libwebp (phlibwebp)
* Other libraries (e.g. KingfisherWebP) depend on standard libwebp
* Both expose the same global C symbols (e.g. WebPDecodeRGBA)
* SwiftPM does not deduplicate or namespace these symbols

As a result, the linker fails due to duplicate symbol definitions.

⸻

Expected Behavior

PostHog should be able to coexist with other libraries that depend on libwebp without causing symbol conflicts.

⸻

Suggested Solutions

Would it be possible to consider one of the following approaches?

Option 1: Namespace / prefix symbols in phlibwebp

Rename exported symbols (e.g. WebPDecodeRGBA → PH_WebPDecodeRGBA) to avoid collisions.

This is a common approach for vendored C libraries.

⸻

Option 2: Make WebP dependency optional

If WebP is only used for specific features (e.g. session replay), consider:

* Making it an optional subspec / target
* Allowing consumers to disable it

⸻

Option 3: Avoid vendoring libwebp

If feasible, depend on a shared/system version instead of bundling a custom one.

(Though this may be difficult if modifications are required.)

⸻

Option 4: Provide a dynamic framework option

Encapsulating phlibwebp inside a dynamic framework could potentially isolate symbols.

⸻

Additional Notes

Because PostHog uses a modified version of libwebp, simply deduplicating dependencies on the client side is not feasible.

At the moment, the only workaround is to fork and manually rename symbols, which is not ideal for maintainability.

⸻

Question

Is there an official or recommended way to use PostHog iOS alongside other WebP-based libraries?

⸻

Thanks!

ioannisj:

Closes #589 

Fixes duplicate symbol linker errors when apps use this SDK together with other dependencies that also include libwebp, such as SDWebImageWebPCoder or KingfisherWebP.

This PR prefixes PostHog’s vendored phlibwebp symbols with <PH...> 

- adding `vendor/libwebp/phlibwebp_prefix.h`
- including that prefix header from the vendored WebP sources/headers
- updating PostHog’s Swift WebP calls to use the prefixed imported APIs

- make buildSdk
- Run local example and ensured WebP generation is okay
- Updated local examples and confirmed that previously failing builds pass: 
  - spm-posthog-kingfisher-webp (was failing, now green)
  - spm-posthog-sdwebimage-webpcoder (was failing, now green)
  - pods-posthog-sdwebimage-webpcoder (was green, remains green)

## :pencil: Checklist

## If releasing new changes


turnipdabeets (approved):
LGTM, just 2 questions: 
- Is it possible to add a test for CI?
- How did you generate the symbol list in phlibwebp_prefix.h? Might be good to add a quick note to this PR documenting in case of rare need to regenerate like a CVE

ioannisj:
> LGTM, just 2 questions:
> 
> * Is it possible to add a test for CI?
> * How did you generate the symbol list in phlibwebp_prefix.h? Might be good to add a quick note to this PR documenting in case of rare need to regenerate like a CVE

1. Yeah, I'll add this with another PR I think cause we'll need at least 2 more samples (1 spm, 1 pod with both vendored dependencies + 3rd party collisions). We an maybe think of having these samples in their own repo and checkout during CI as well?
2. Added

PR #592, opened 2026-05-06
feat(logs): add public captureLog API and logger facade

turnipdabeets:

Public API for the logs feature on top of [#590](https://github.com/PostHog/posthog-ios/pull/590) (storage + queue). **Stacked on `feat/logs-storage-queue`**, not `main`.

```swift
PostHogSDK.shared.captureLog("hello", level: .info, attributes: ["k": "v"])
PostHogSDK.shared.logger?.warn("careful")
PostHogSDK.shared.flush()  // drains events + replay + logs
```

Adds `PostHogSDK.captureLog(_:level:attributes:traceId:spanId:traceFlags:)` (`@objc` overloads + Swift default-arg variant) and `PostHogSDK.logger`, a thin facade with `trace / debug / info / warn / error / fatal`. `PostHogSDK.flush()` drains all three queues.

`beforeSend` runs at the SDK layer before `logsQueue.add` so sensitive data is scrubbed before entering any internal pipeline; rate cap stays at the queue layer. Capture-time context (`distinctId`, `sessionId`, `screenName`, `appState`, `featureFlagKeys`) is snapshotted on the calling thread so identity / session changes between capture and flush can't mutate a queued record. Screen name is cached on `PostHogLogger.lastScreenName` via the `screenViewPublisher`. `PostHogSDK.screen()` (called manually or via the SwiftUI `.postHogScreenView` modifier) now broadcasts to that publisher, and the auto-capture swizzle is owned explicitly by `PostHogScreenViewIntegration` (no longer triggered as a side effect of subscribing), so SwiftUI apps see the modifier's name in `screen.name` and apps that don't enable `captureScreenViews` don't pay UIKit swizzling cost. Non-positive `rateCapMaxLogs`/`rateCapWindowSeconds` are clamped to 0 (disabled) at queue init, and the snapshotted fields (`serviceName`, `serviceVersion`, `environment`, `resourceAttributes`) are documented as set-before-`setup`.

`make test` — **512 tests in 105 suites pass.**

`PostHogTests/PostHogLogsCaptureTest.swift` covers main / background / concurrent capture, empty- and whitespace-only body drop, opt-out drop, captureLog-after-close smoke test, multi-byte unicode body round-trip, capture-time `distinctId` snapshotting (identify A → captureLog → identify B → flush → record carries A), `logger.info` ≡ `captureLog level: .info`, all six severity numbers on the wire, traceId/spanId/traceFlags on the OTLP wire, OTLP resource attributes, rate-cap drop at the SDK boundary, flush from background thread, and backgrounding triggers a logs flush. `PostHogTests/PostHogLogsQueueTest.swift` adds a rate-cap negative-input clamp test. `PostHogTests/PostHogScreenViewIntegrationTest.swift` adds three for the publisher refactor: integration owns the auto-capture lifecycle, manual `screen()` reaches subscribers, and `captureScreenViews=false` + manual `screen()` still feeds `logger.lastScreenName`.

`PostHogExample/Views/LogsView.swift` — demo each level, flush, traced-log, and a 1000-record flood (rate-cap demo). End-to-end verified in PostHog Cloud: a fresh log from the example app shows `screen.name = ContentView` (was `UIHostingController<ModifiedContent<AnyView, RootModifier>>` before the refactor).

## :pencil: Checklist

## If releasing new changes

- [x] Ran `pnpm changeset` to generate a changeset file (covers the combined feature).


PostHog/PostHogSDK.swift
```diff
@@ -890,6 +915,136 @@ let maxRetryDelay = 30.0
...
+        }
+        guard let logsQueue else { return }
+
+        // Capture-time context snapshot — cheap reads only, safe from any
+        // thread. Identity/session changes between capture and flush will not
+        // alter the record because the snapshot is frozen here.
+        let dId = getDistinctId()
+        let sId = sessionManager.getSessionId(readOnly: true)
+        let scr = currentScreenName
+        let appState = sessionManager.isAppInBackgroundSnapshot ? "background" : "foreground"
```
marandaneto: lets create consts for those strings

turnipdabeets: updated

PostHog/PostHogSDK.swift
```diff
@@ -890,6 +915,136 @@ let maxRetryDelay = 30.0
...
+            return
+        }
+        guard let logsQueue else { return }
+
+        // Capture-time context snapshot — cheap reads only, safe from any
+        // thread. Identity/session changes between capture and flush will not
+        // alter the record because the snapshot is frozen here.
+        let dId = getDistinctId()
+        let sId = sessionManager.getSessionId(readOnly: true)
+        let scr = currentScreenName
```
marandaneto: not a big fan of such shortened versions because its unclear when we read it

PostHog/Logs/PostHogLogRecord.swift
```diff
@@ -10,6 +10,12 @@ import Foundation
 /// Records are produced by `captureLog` (or `logger.<level>`), persisted to disk
 /// via `PostHogLogsQueue`, and serialized as OpenTelemetry log records on the wire.
 @objc(PostHogLogRecord) public final class PostHogLogRecord: NSObject {
+    /// Valid `appState` values.
+    enum AppState {
```
ioannisj: Could be a String enum?

PostHog/PostHogSDK.swift
```diff
@@ -159,6 +167,13 @@ let maxRetryDelay = 30.0
                 self?.notifyContextDidChange()
             }
 
+            // Cache the latest screen name so `captureLog` can tag log records
+            // without needing main-thread access to UIViewController helpers.
+            screenViewToken = DI.main.screenViewPublisher.onScreenView.subscribe { [weak self] name in
```
ioannisj: Feels like this should be responsibility of PostHogLogger instead?. PostHogSDK can read from logger instance logger.lastScreenName

.changeset/logs-feature.md
```diff
@@ -0,0 +1,13 @@
+---
+"posthog-ios": minor
+---
+
+Add a logs feature for shipping structured log records from iOS, macOS, tvOS, watchOS, and visionOS apps:
+
+```swift
+PostHog.shared.captureLog("hello", level: .info, attributes: ["k": "v"])
+PostHog.shared.logger.info("ready")
+PostHog.shared.flushLogs()
```
ioannisj: These should be PostHogSDK and flushLogs() is just flush() right?

PostHog/PostHogSDK.swift
```diff
@@ -890,6 +915,136 @@ let maxRetryDelay = 30.0
...
+    /// callers correlating logs with distributed traces; most apps should use
+    /// the simpler overload.
+    ///
+    /// - Parameters:
+    ///   - body: The log message. Required and non-empty.
+    ///   - level: Severity.
+    ///   - attributes: Per-record attributes. Same rules as the simpler overload.
+    ///   - traceId: 32-character lowercase hex W3C trace id.
+    ///   - spanId: 16-character lowercase hex W3C span id.
+    ///   - traceFlags: W3C trace flags bitfield (bit 0 is the `sampled` flag).
```
ioannisj, line 968: Should we validate these at capture time you think? Probably a question for @PostHog/logs team

PostHog/PostHogSDK.swift
```diff
@@ -890,6 +915,136 @@ let maxRetryDelay = 30.0
...
+        guard let logsQueue else { return }
+
+        // Capture-time context snapshot — cheap reads only, safe from any
+        // thread. Identity/session changes between capture and flush will not
+        // alter the record because the snapshot is frozen here.
+        let dId = getDistinctId()
+        let sId = sessionManager.getSessionId(readOnly: true)
+        let scr = currentScreenName
+        let appState = sessionManager.isAppInBackgroundSnapshot ? PostHogLogRecord.AppState.background : PostHogLogRecord.AppState.foreground
+        let flagKeys = remoteConfig?.getFeatureFlags()?.keys.map { String($0) } ?? []
```
ioannisj: Should these be only active flags? For normal events in dynamicContext() we include only active flags I think. Also on the first PR, the comment for this field calls out for active flags only

PostHog/Logs/PostHogLogRecord.swift
turnipdabeets: updated

PostHog/PostHogSDK.swift
turnipdabeets: updated

.changeset/logs-feature.md
turnipdabeets: yes they should! updated

PostHog/PostHogSDK.swift
turnipdabeets: yes, updated!

PostHog/PostHogSDK.swift
turnipdabeets: Checked the JS + RN sources — neither validates trace_id, span_id, or trace_flags. https://github.com/PostHog/posthog-js/blob/main/packages/core/src/logs/logs-utils.ts#L154-L162 - we probably wouldn't want to block but I could see us maybe log a warning on failed validation.

PostHog/PostHogSDK.swift
turnipdabeets: updated!

PostHog/PostHogSDK.swift
ioannisj: Okay, then if neither sdk is doing this I think we are okay. We can circle back to all of the to add warnings cause I imagine people will only notice logs missing after ingestion fails

.changeset/logs-feature.md
```diff
@@ -0,0 +1,13 @@
+---
+"posthog-ios": minor
+---
+
+Add a logs feature for shipping structured log records from iOS, macOS, tvOS, watchOS, and visionOS apps:
+
+```swift
+PostHogSDK.shared.captureLog("hello", level: .info, attributes: ["k": "v"])
+PostHogSDK.shared.logger.info("ready")
```
ioannisj:
```suggestion
PostHogSDK.shared.logger?.info("ready")
```

PostHog/ScreenViews/PostHogScreenViewIntegration.swift
```diff
@@ -46,31 +45,14 @@ final class PostHogScreenViewIntegration: PostHogIntegration {
...
 
-    /**
-     Start capturing screen view events
-     */
     func start() {
-        let screenViewPublisher = DI.main.screenViewPublisher
-        screenViewToken = screenViewPublisher.onScreenView.subscribe { [weak self] screen in
-            self?.captureScreenView(screen: screen)
+        DI.main.screenViewPublisher.startAutoCapture { [weak self] screenName in
+            self?.postHog?.screen(screenName)
```
ioannisj, line 54: Couldn't we skip this refactoring here by having ScreenViewIntegration call an internal `screen(screenName, notifyPublisher: false)` instead that would skip fanout of screen name?

PostHog/PostHogSDK.swift
```diff
@@ -1026,6 +1165,10 @@ let maxRetryDelay = 30.0
         }
 
         queueEvent(event, queue: queue)
+
+        // Fanout to subscribers (e.g. the logs feature's lastScreenName).
+        // Subscribers must not re-enter screen() from the callback.
+        DI.main.screenViewPublisher.onScreenView.invoke(screenTitle)
```
ioannisj: nit: I would rather not invoke directly but something in the line of  DI.main.screenViewPublisher.onNewScreenName(screenTitle). Then publisher does the .invoke call

Ideally consumers of PostHogMulticastCallback should only be able to .subscribe, at least that's the intent

PostHog/PostHogSessionManager.swift
```diff
@@ -123,6 +123,12 @@ import Foundation
         return currentSessionId
```
ioannisj, line 129: <p class='graphite__hidden'><i>[Re: line +38]</i></p>
I would seed with current application state here at the time of call since NotificationCenter doesn't replay past lifecycle events? In case of a late setup() call, isAppInBackground used in logs would stay false until another background/foreground event?
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/PostHog/posthog-ios/592?utm_source=unchanged-line-comment">Graphite</a>.</p>

ioannisj (approved):
Left a few more comments. I'll approve since this is very close and can be merged to feature branch. We can address those there?

PostHog/PostHogSDK.swift
turnipdabeets: Done

PostHog/PostHogSessionManager.swift
turnipdabeets: Good catch — done. Added var isInBackground: Bool { get } to AppLifecyclePublishing.

ioannisj: huh, just saw the "See this comment inline on Graphite". Weird, in gh this appears on the wrong line. Sorry :)

PostHog/ScreenViews/PostHogScreenViewIntegration.swift
turnipdabeets: I'm not sure we can do this actually since this is listening to an event.

the integration is still subscribed to the publisher, so manual screen("X") calls would emit twice: once from the user's call, once from the integration's callback firing on the publisher invoke. notifyPublisher: false stops the loop but not the duplicate. Would need a second flag or a re-entry guard to fix.

turnipdabeets: added a comment to this line so it's more clear

PR #593, opened 2026-05-06
refactor(api): share a single URLSession across all PostHogApi requests

turnipdabeets:

Stacked on [#592](https://github.com/PostHog/posthog-ios/pull/592). **Base: `feat/logs-public-api`**, not `main`.

Per-request `URLSession` is an anti-pattern. The SDK currently constructs a fresh `URLSession(configuration: ...).uploadTask(...)` for every batch / snapshot / logs / flags / remoteConfig call (5 sites in `PostHogApi.swift`). Each fresh session has its own connection pool, so back-to-back requests pay a TCP + TLS handshake every time and HTTP/2 multiplexing — which the PostHog backend supports (`us.i.posthog.com` and `eu.i.posthog.com` both negotiate HTTP/2 via Envoy) — never engages.

Apple's guidance (WWDC 2018 session 714 + DTS forum thread 84663) is explicit: reuse a single `URLSession` instance. Quinn the Eskimo: *"This is a common anti-pattern… creating a session per request prevents connection reuse, which can radically slow down back-to-back requests. This is especially bad for HTTP/2."*

This PR lifts `URLSession` to a stored property on `PostHogApi`, built once in `init`. All five API methods now share the same session and connection pool. Per-request gzip semantics move from session-level `httpAdditionalHeaders` onto `URLRequest` via `Content-Encoding`, so `/flags` and `/array/<token>/config` don't carry a misleading gzip header (their bodies aren't gzipped).

Concrete wins:

- **TLS handshake amortised** across requests instead of paid per-call (~150-300ms on cellular per handshake).
- **HTTP/2 multiplexing** kicks in when events / replay / logs / flags flush concurrently — one TCP connection carries all four streams.
- **Reduced per-request setup cost.** Apple says `URLSession` is "fairly expensive to create" with "non-trivial memory footprint."

`make test` — **503 tests pass**, no behaviour change at the wire level.

`xcodebuild -scheme PostHog -destination 'generic/platform=iOS Simulator' build` — green.

Manual: not yet validated against a real backend. Worth confirming with a captured network trace that subsequent requests reuse the connection (single SYN/ACK + TLS, then multiple POSTs over it) rather than opening fresh connections per request.

## :pencil: Checklist

## If releasing new changes

- [ ] Ran `pnpm changeset` to generate a changeset file. *(not yet — pure perf refactor with no observable wire-level change; will add if the team wants a changelog entry for it)*


marandaneto:
nice!

PR #601, opened 2026-05-14
feat(surveys): add survey translations support

lucasheriques:
## Summary
Brings the posthog-js / posthog-react-native survey translations feature to the native SDK. Surveys can carry per-language overrides for user-visible strings via a `translations` map keyed by language code. At display time the SDK resolves a language (init override → person property `language` → device locale), applies any matching translation onto the display model, and stamps the matched key as `$survey_language` on every survey event when a translation actually took effect.

- Wire model: `translations` on `PostHogSurvey` and on each question struct.
- Config: `PostHogSurveysConfig.overrideDisplayLanguage` (`@objc`) init option.
- Matching: case-insensitive exact match with base-language fallback (`pt-BR` → `pt`); returns the original-cased dict key for event reporting.
- Resolution: only stamps `matchedKey` when at least one user-visible field actually changed.
- Event threading: `survey shown`, `survey sent`, `survey dismissed` all carry `$survey_language` when matched; `$survey_questions` reflects translated text.
- Device locale read via `Locale.current.identifier` with `_`→`-` normalization for BCP-47 parity with other SDKs.
- `PostHogExample` enables `surveys = true` for reviewer testing.

Companion PR: PostHog/posthog-android#513

## Test plan
All tests pass in CI. New coverage:
- Wire-format decoding round-trip (Codable + fixture).
- Matching algorithm: exact, case-insensitive, base-language fallback, no-match, no-fallback-without-hyphen.
- Priority chain: override → person prop → device locale.
- Display model translation with field-level fallback to original.
- Integration: `$survey_language` present on all three events (`shown`/`sent`/`dismissed`) when matched, absent otherwise.

---

### DIFF
--- a/PostHog/Models/Surveys/PostHogSurvey+Display.swift
+++ b/PostHog/Models/Surveys/PostHogSurvey+Display.swift
@@ -2,66 +2,78 @@
     import Foundation
 
     extension PostHogSurvey {
-        func toDisplaySurvey() -> PostHogDisplaySurvey {
-            PostHogDisplaySurvey(
+        func toDisplaySurvey(
+            surveyTranslation: PostHogSurveyTranslation? = nil,
+            questionTranslations: [PostHogSurveyQuestionTranslation?]? = nil
+        ) -> PostHogDisplaySurvey {
+            let translatedQuestions = questions.enumerated().compactMap { index, question in
+                question.toDisplayQuestion(translation: questionTranslations?[safe: index] ?? nil)
+            }
+            return PostHogDisplaySurvey(
                 id: id,
-                name: name,
-                questions: questions.compactMap { $0.toDisplayQuestion() },
-                appearance: appearance?.toDisplayAppearance(),
+                name: surveyTranslation?.name ?? name,
+                questions: translatedQuestions,
+                appearance: appearance?.toDisplayAppearance(translation: surveyTranslation),
                 startDate: startDate,
                 endDate: endDate
             )
         }
     }
 
+    private extension Array {
+        subscript(safe index: Int) -> Element? {
+            indices.contains(index) ? self[index] : nil
+        }
+    }
+
     extension PostHogSurveyQuestion {
-        func toDisplayQuestion() -> PostHogDisplaySurveyQuestion? {
+        func toDisplayQuestion(translation: PostHogSurveyQuestionTranslation? = nil) -> PostHogDisplaySurveyQuestion? {
             switch self {
             case let .open(question):
                 return PostHogDisplayOpenQuestion(
                     id: question.id,
-                    question: question.question,
-                    questionDescription: question.description,
+                    question: translation?.question ?? question.question,
+                    questionDescription: translation?.description ?? question.description,
                     questionDescriptionContentType: question.descriptionContentType?.toDisplayContentType(),
                     isOptional: question.optional ?? false,
-                    buttonText: question.buttonText
+                    buttonText: translation?.buttonText ?? question.buttonText
                 )
 
             case let .link(question):
                 return PostHogDisplayLinkQuestion(
                     id: question.id,
-                    question: question.question,
-                    questionDescription: question.description,
+                    question: translation?.question ?? question.question,
+                    questionDescription: translation?.description ?? question.description,
                     questionDescriptionContentType: question.descriptionContentType?.toDisplayContentType(),
                     isOptional: question.optional ?? false,
-                    buttonText: question.buttonText,
-                    link: question.link ?? ""
+                    buttonText: translation?.buttonText ?? question.buttonText,
+                    link: translation?.link ?? question.link ?? ""
                 )
 
             case let .rating(question):
                 return PostHogDisplayRatingQuestion(
                     id: question.id,
-                    question: question.question,
-                    questionDescription: question.description,
+                    question: translation?.question ?? question.question,
+                    questionDescription: translation?.description ?? question.description,
                     questionDescriptionContentType: question.descriptionContentType?.toDisplayContentType(),
                     isOptional: question.optional ?? false,
-                    buttonText: question.buttonText,
+                    buttonText: translation?.buttonText ?? question.buttonText,
                     ratingType: question.display.toDisplayRatingType(),
                     scaleLowerBound: question.scale.range.lowerBound,
                     scaleUpperBound: question.scale.range.upperBound,
-                    lowerBoundLabel: question.lowerBoundLabel,
-                    upperBoundLabel: question.upperBoundLabel
+                    lowerBoundLabel: translation?.lowerBoundLabel ?? question.lowerBoundLabel,
+                    upperBoundLabel: translation?.upperBoundLabel ?? question.upperBoundLabel
                 )
 
             case let .singleChoice(question), let .multipleChoice(question):
                 return PostHogDisplayChoiceQuestion(
                     id: question.id,
-                    question: question.question,
-                    questionDescription: question.description,
+                    question: translation?.question ?? question.question,
+                    questionDescription: translation?.description ?? question.description,
                     questionDescriptionContentType: question.descriptionContentType?.toDisplayContentType(),
                     isOptional: question.optional ?? false,
-                    buttonText: question.buttonText,
-                    choices: question.choices,
+                    buttonText: translation?.buttonText ?? question.buttonText,
+                    choices: translation?.choices ?? question.choices,
                     hasOpenChoice: question.hasOpenChoice ?? false,
                     shuffleOptions: question.shuffleOptions ?? false,
                     isMultipleChoice: isMultipleChoice
@@ -99,7 +111,7 @@
     }
 
     extension PostHogSurveyAppearance {
-        func toDisplayAppearance() -> PostHogDisplaySurveyAppearance {
+        func toDisplayAppearance(translation: PostHogSurveyTranslation? = nil) -> PostHogDisplaySurveyAppearance {
             PostHogDisplaySurveyAppearance(
                 fontFamily: fontFamily,
                 backgroundColor: backgroundColor,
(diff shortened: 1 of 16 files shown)


PostHog/Surveys/PostHogSurveyIntegration.swift
```diff
@@ -662,12 +725,29 @@
...
                 activeSurveyQuestionIndex = 0
             }
         }
 
+        private func resolveDisplayLanguage() -> String? {
+            // `surveysConfig` is `@available(macOS, unavailable)`, but this integration is
+            // also compiled for non-iOS targets under TESTING. Read the backing property
+            // directly so the file stays compilable on every platform.
+            let override = config?._surveysConfig.overrideDisplayLanguage
+            let personProperties = storage?.getDictionary(forKey: .personPropertiesForFlags) as? [String: Any]
```
turnipdabeets: Could we read this from remoteConfig instead of storage?

I think `getPersonPropertiesForFlags()` does what we need. It's private today, so we'd need to drop that to internal so the surveys integration can call it.

PostHog/Models/Surveys/PostHogSurvey.swift
```diff
@@ -40,6 +40,43 @@ struct PostHogSurvey: Decodable, Identifiable {
     let endDate: Date?
     /// The schedule for the survey (optional). Determines how often the survey can be shown.
     let schedule: PostHogSurveySchedule?
+    let translations: [String: PostHogSurveyTranslation]?
+
+    init(
```
ioannisj: Is this needed, can we do with synthesized swift memberwise inits?

ioannisj: Same goes for other models in this dir

PostHog/Models/Surveys/PostHogSurvey+Display.swift
```diff
@@ -2,66 +2,78 @@
...
 
     extension PostHogSurvey {
-        func toDisplaySurvey() -> PostHogDisplaySurvey {
-            PostHogDisplaySurvey(
+        func toDisplaySurvey(
+            surveyTranslation: PostHogSurveyTranslation? = nil,
+            questionTranslations: [PostHogSurveyQuestionTranslation?]? = nil
+        ) -> PostHogDisplaySurvey {
+            let translatedQuestions = questions.enumerated().compactMap { index, question in
+                question.toDisplayQuestion(translation: questionTranslations?[safe: index] ?? nil)
```
ioannisj: nit: we're enumerating here, so [safe: index] is probably a bit redundant here. Could also lead to double optional since we're operating on `[PostHogSurveyQuestionTranslation?]`?

PostHog/Surveys/Utils/SurveyTranslationResolver.swift
```diff
@@ -0,0 +1,190 @@
...
+        }
+
+        if !surveyChanged, !anyQuestionChanged { return empty }
+
+        let matchedKey: String?
+        if surveyChanged {
+            matchedKey = surveyKey
+        } else {
+            matchedKey = questionKeys.compactMap { $0 }.first
+        }
```
ioannisj: Is this intentional? I don't completely get the logic of selecting surveyKey vs questionKey

PostHog/Surveys/Utils/SurveyTranslationResolver.swift
```diff
@@ -0,0 +1,190 @@
...
+        let surveyKey = findBestTranslationMatch(
+            translations: survey.translations,
+            targetLanguage: targetLanguage
+        )
+        let surveyTranslation = surveyKey.flatMap { survey.translations?[$0] }
+        let surveyChanged = surveyTranslation.map { surveyTranslationChangesAnything(survey: survey, translation: $0) } ?? false
+
+        var anyQuestionChanged = false
+        var questionTranslations: [PostHogSurveyQuestionTranslation?] = []
+        var questionKeys: [String?] = []
```
ioannisj: We are collecting questionKeys so that we can later filter for the first non-nil value right? We could just manage a `questionKey: String?` instead?

PostHog/Surveys/PostHogSurveyIntegration.swift
lucasheriques: yep done in 2fe68ce. made `getPersonPropertiesForFlags()` package-internal and the integration calls it directly now

PostHog/Models/Surveys/PostHogSurvey.swift
lucasheriques: yep dropped the explicit init in 2fe68ce, relying on the synthesized memberwise init now. one gotcha though: swift's `Decodable` synthesis ignores `let translations: T? = nil` (the property stays at the default during JSON decoding), so i had to drop the `= nil` default and update existing test callers to pass `translations: nil` explicitly

lucasheriques: applied to all four question structs too in 2fe68ce — same approach, synthesized memberwise inits with `translations: nil` passed explicitly from existing test callers

PostHog/Models/Surveys/PostHogSurvey+Display.swift
lucasheriques: yep nit fixed in 2fe68ce — dropped the extension and replaced with a direct bounds check inline. avoids the double-optional too

PostHog/Surveys/Utils/SurveyTranslationResolver.swift
lucasheriques: good question — clarified in 2fe68ce with a brief comment. in the common case both `surveyKey` and `questionKey` resolve to the same dict key (same target language). the only way they'd differ is if the survey-level and question-level dicts spelled the same language differently (e.g. "Fr" vs "fr"). when survey-level translations actually drove a change i preferred its key; otherwise fall back to whichever question's translation key matched first. lmk if you'd rather just unconditionally prefer surveyKey when both are non-nil

PostHog/Surveys/Utils/SurveyTranslationResolver.swift
lucasheriques: yep done in 2fe68ce — single `firstQuestionKey: String?` instead of the array + post-filter. applied the same simplification on the android side for parity

turnipdabeets (approved):
LGTM! Two small test gaps worth filling before/after merge: the "only stamp matchedKey when a field changed" guard and the translated $survey_questions payload aren't directly asserted.

lucasheriques:
thanks anna! both gaps filled in 86ea7c6:

- new unit test `returns nil matchedKey when translation matches but values are identical` with a dedicated no-op fixture
- new integration test `survey sent carries translated question text in $survey_questions` that asserts the translated string actually lands in the event payload

i also added the equivalent integration-level no-op test on the android PR for parity (the JVM unit test for it was already there)

PostHog/Models/Surveys/PostHogSurveyTranslation.swift
```diff
@@ -0,0 +1,34 @@
...
+//  PostHog
+//
+//  Created by PostHog Code on 2026-05-13.
+//
+
+import Foundation
+
+/// Localized overrides for a survey, keyed by language code (e.g. `"fr"`, `"pt-BR"`).
+///
+/// The survey-level `description` is intentionally NOT translatable — it is only used
```
turnipdabeets, line 12: is this comment still valid?

PostHogTests/PostHogSurveyTranslationsTest.swift
```diff
@@ -0,0 +1,409 @@
...
+                let survey = try decodeTranslationsFixture()
+                let resolved = resolveSurveyTranslations(survey: survey, targetLanguage: "fr")
+                #expect(resolved.matchedKey == "fr")
+                #expect(resolved.survey?.name == "Bonjour")
+                #expect(resolved.questions[0]?.question == "Comment etait-ce?")
+                #expect(resolved.questions[1]?.choices == ["Un", "Deux"])
+            }
+
+            @Test("falls back to base language")
+            func basesLanguageFallback() throws {
```
turnipdabeets: nit typo: 
```suggestion
            func baseLanguageFallback() throws {
```

PostHog/Surveys/Utils/SurveyTranslationResolver.swift
```diff
@@ -0,0 +1,181 @@
...
+
+        let normalizedTarget = target.lowercased()
+
+        if let exact = translations.keys.first(where: { $0.lowercased() == normalizedTarget }) {
+            return exact
+        }
+
+        if let hyphen = normalizedTarget.firstIndex(of: "-") {
+            let base = String(normalizedTarget[..<hyphen])
+            if let baseMatch = translations.keys.first(where: { $0.lowercased() == base }) {
```
turnipdabeets: nit: 2nd `translations.keys.first(where: { $0.lowercased()`

lucasheriques: good catch — pulled into a local `findKey` helper in 92785b3 so it's only written once. did the same on the android side for parity

PostHogTests/PostHogSurveyTranslationsTest.swift
lucasheriques: fixed in 92785b3, thanks!

PostHog/Models/Surveys/PostHogSurveyTranslation.swift
lucasheriques: yep — survey-level `description` is intentionally not translatable in posthog-js (the JS `SurveyTranslation` interface only covers `name` + the three `thankYouMessage*` fields). keeping it native-side too. happy to drop the comment if it reads as a footgun instead of documentation though

Issue #557, opened 2026-04-14
Next iOS major version plan


marandaneto:
### Description

* [Codable support](https://github.com/PostHog/posthog-ios/issues/52)
* Remove deprecated fields
* [Completion handler support](https://github.com/PostHog/posthog-ios/issues/43)
* Migrate to Swift 6 (and strong concurrency?)
* `setup` method should take the projectToken and not the `PostHogConfig`
* async/await instead of callbacks
* strongly typed `PostHogEvent.properties` for known values but still accepts any key/value?
* breakdown [session replay](https://github.com/PostHog/posthog-ios/issues/301) and error tracking in modules to reduce binary size due to libp and plcrashreporter and also naming clashes
* Remove fields eg `properties.build` in favor of only `properties.$app_build`, maybe there are more similar to this one
* Remove session replay wireframe mode?
* Simplify config file (eg sessionReplay should be part of PostHogSessionReplayConfig)
* Remove internal fields that are public
* Config should be nullable and prefer remote config if not set (eg sessionReplay = false (false), sessionReplay = true && remote config is true (true), sessionReplay = nil && remote config is true (true), this matches the JS SDK)
* Prob a bunch of ideas from [this](https://github.com/PostHog/posthog/issues/48234)

PR #604, opened 2026-05-20
Add downgrade compatibility CI test

dustinbyrne:

Adds a CI regression guard for https://github.com/PostHog/posthog-ios/issues/537.

The new job writes persisted SDK state with the current checkout, then starts an older SDK version against that same state. This catches downgrade-incompatible state changes like the UUID queue filename issue that caused 3.48.0 startup crashes after running 3.48.1+.

- `make testDowngradeCompatibility`
- `WRITE_REF=3.48.1 DOWNGRADE_REF=3.48.0 make testDowngradeCompatibility` fails as expected with the historical crash
- `bash -n scripts/test-downgrade-compatibility.sh`
- `git diff --check`
- Invalid ref hardening check: `DOWNGRADE_REF='--upload-pack=sh' make testDowngradeCompatibility` rejects the ref before git commands run
- Ran pre-PR review loop and addressed kept findings

## :pencil: Checklist

## If releasing new changes

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,3 +29,21 @@ jobs:
             ${{ runner.os }}-spm-
       - name: Test SDK
         run: make test
+
+  downgrade-compatibility:
+    runs-on: macos-15
+    strategy:
+      fail-fast: false
+      matrix:
+        downgrade_ref:
+          # Regression guard for https://github.com/PostHog/posthog-ios/issues/537:
+          # 3.48.0 force-parsed queue filenames as Doubles during startup.
+          - 3.48.0
+    name: Downgrade compatibility (${{ matrix.downgrade_ref }})
+    steps:
+      - uses: actions/checkout@v6
+      - uses: maxim-lobanov/setup-xcode@ed7a3b1fda3918c0306d1b724322adc0b8cc0a90 # v1.7.0
+        with:
+          xcode-version: latest-stable
+      - name: Test downgrade compatibility
+        run: make testDowngradeCompatibility downgrade_ref=${{ matrix.downgrade_ref }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -46,4 +46,6 @@ jobs:
         with:
           xcode-version: latest-stable
       - name: Test downgrade compatibility
-        run: make testDowngradeCompatibility downgrade_ref=${{ matrix.downgrade_ref }}
+        env:
+          DOWNGRADE_REF: ${{ matrix.downgrade_ref }}
+        run: make testDowngradeCompatibility
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -45,7 +45,7 @@ jobs:
           - 3.58.0
     name: Downgrade compatibility (${{ matrix.downgrade_ref }})
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - uses: maxim-lobanov/setup-xcode@ed7a3b1fda3918c0306d1b724322adc0b8cc0a90 # v1.7.0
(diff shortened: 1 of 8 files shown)


ioannisj:
Looks good, a couple of thoughts: 
- This only tests event queue. Ideally we should be testing all or more [storage keys](https://github.com/PostHog/posthog-ios/blob/cf6155963a6532c7b88e0cfa2b1d07595d6b6db6/PostHog/PostHogStorage.swift#L227)? (replay queue and newly added logs queue for example) 
- This basically checks if there's no hard crash, but should we be also making sure that old version can correctly parse the events? Old versions may be failing silently with events dropped for example
- This focuses to a single point-in-time version 3.48.0. Would it make sense to instead test last N versions? A change in persistence in version X, for example, could work well with X-1 but not X-3

dustinbyrne:
commit 31d6c1c9b0014d578397a4d57c498be9335f9269 fixes one more downgrade crash caught via:
https://github.com/PostHog/posthog-ios/actions/runs/26190386156/job/77056968579?pr=604#step:4:1290

worth another pair of eyes if this was valid or not

dustinbyrne:
>This only tests event queue. Ideally we should be testing all or more [storage keys](https://github.com/PostHog/posthog-ios/blob/cf6155963a6532c7b88e0cfa2b1d07595d6b6db6/PostHog/PostHogStorage.swift#L227)? (replay queue and newly added logs queue for example) 

I've updated this to also test the session replay + logging queue

> This basically checks if there's no hard crash, but should we be also making sure that old version can correctly parse the events? Old versions may be failing silently with events dropped for example

I'm open to thoughts on this. What I have in this PR is definitely less concerned about whether or not we still do the right thing, and more so just validating that we don't crash unexpectedly. Given we're testing code that's already shipped as part of a downgrade path, is it reasonable to assume that we can always gracefully migrate down?

> This focuses to a single point-in-time version 3.48.0. Would it make sense to instead test last N versions? A change in persistence in version X, for example, could work well with X-1 but not X-3

I think this is fine (even preferred), as long as we're testing specific versions where the persisted schema / storage protocol changed meaningfully. The last N versions might miss something.

.github/workflows/test.yml
```diff
@@ -29,3 +29,27 @@ jobs:
...
+          # Pin persisted-storage schema boundaries rather than the last N releases.
+          # 3.48.0 force-parsed queue filenames as Doubles during startup.
+          - 3.48.0
+          # 3.48.3 introduced the UUID event queue folder.
+          - 3.48.3
+          # 3.58.0 introduced the persisted logs queue.
+          - 3.58.0
+    name: Downgrade compatibility (${{ matrix.downgrade_ref }})
+    steps:
+      - uses: actions/checkout@v6
```
marandaneto: pin action

PostHog/PostHogStorage.swift
```diff
@@ -230,7 +230,8 @@ class PostHogStorage {
         case queue = "posthog.queueFolder.uuid" // queue from > 3.48.1
         case oldQueueFolder = "posthog.queueFolder" // queue from 3.0.0 - 3.48.1
         case oldQueuePlist = "posthog.queue.plist" // queue from pre-3.0.0
-        case replayQeueue = "posthog.replayFolder"
+        case replayQeueue = "posthog.replayFolder.uuid" // replay queue with UUID filenames
+        case oldReplayQueue = "posthog.replayFolder" // replay queue with legacy timestamp filenames
```
marandaneto, line 234: added to https://github.com/PostHog/posthog-ios/issues/557

marandaneto:
ideally we'd have an E2E test running on an emulator and doing some UI clicks with a mocked webserver so the calls are successful (to avoid prod flakiness(due to lag) like PH JS)
This E2E could restart the app, downgrade, upgrade, but this is a fair start and already catches some issues so lets go for it

marandaneto (approved):

PR #609, opened 2026-05-22
feat: auto-attach $screen_name to all events

### LINKED ISSUE #119: Set `$screen_name` to all events
### Description

- [ ] Android
- [ ] iOS
- [X] Flutter (when calling `screen` via PosthogObserver or manually)
- [X] RN (when calling `screen` via wix, react-navigation or manually)

turnipdabeets:

Closes PostHog/posthog-android#119 for iOS. RN and Flutter already attach `$screen_name` to every event once `screen()` has been called; iOS did not. Android equivalent: PostHog/posthog-android#530.

Six unit tests in `PostHogScreenNameTest`: before-screen baseline, after-screen attaches, caller override wins, `reset()` clears, `$exception` carries it (Option A cross-event scope), `$snapshot` does not (verifies `appendSharedProps = false` gate).

End-to-end on iPhone 17 simulator across all five customer-visible scenarios — confirmed via Xcode console and by querying the `events` table on `us.posthog.com`:

| Scenario | Event | `$screen_name` |
|---|---|---|
| Auto-capture (`captureScreenViews=true`) | `test_auto_capture` | `UIHostingController<ModifiedContent<AnyView, RootModifier>>` |
| Manual `screen("ManualScreen")` | `test_after_manual` | `ManualScreen` |
| Caller override | `test_override` | `OverrideValue` |
| After `reset()` | `test_after_reset` | _absent_ |
| Exception after `screen("ExceptionScreen")` | `$exception` | `ExceptionScreen` |

**Behavior change** for reviewers: existing customers who have not disabled `captureScreenViews` will start seeing `$screen_name` attached to `$exception`, `$identify`, etc. — not only `$screen` events. Opt-out is `captureScreenViews = false` plus avoiding manual `PostHogSDK.shared.screen()` calls.

## :pencil: Checklist

## If releasing new changes

### DIFF
--- a/.changeset/screen-name-on-all-events.md
+++ b/.changeset/screen-name-on-all-events.md
@@ -1,0 +1,5 @@
+---
+"posthog-ios": minor
+---
+
+Auto-attach `$screen_name` to every captured event after `PostHogSDK.shared.screen()` has been called, or whenever screen auto-capture is on (default). Cached value is cleared by `reset()` and `close()`. To opt out, set `PostHogConfig.captureScreenViews = false` and avoid calling `screen()` manually. Behavior change: existing events from `captureException`, `identify`, etc. will start carrying `$screen_name` for users who hadn't disabled screen capture. Closes posthog-android#119.
--- a/.changeset/screen-name-on-all-events.md
+++ b/.changeset/screen-name-on-all-events.md
@@ -2,18 +2,6 @@
 "posthog-ios": minor
 ---
 
-Auto-attach `$screen_name` to every captured event after `PostHogSDK.shared.screen()` has been called (manually or via screen-view auto-capture). Cached value is cleared by `reset()` and `close()`. Closes posthog-android#119.
+Auto-attach `$screen_name` to every captured event after `PostHogSDK.shared.screen()` has been called (manually or via screen-view auto-capture). Cached value is cleared by `reset()` and `close()`.
 
 **To opt out of `$screen_name` stamping entirely**, set `PostHogConfig.captureScreenViews = false` **and** stop calling `screen()` manually. Disabling `captureScreenViews` alone is not sufficient — a single manual `screen("Home")` call will re-enable stamping.
-
-## Behavior changes
-
-These all affect what your events carry on the wire. Review your dashboards/insights/HogQL queries:
-
-- **Cross-event stamping.** `$exception`, `$identify`, `$autocapture`, `$create_alias`, `$groupidentify`, custom events, etc. will start carrying `$screen_name` whenever a screen has been recorded in the session. Previously only `$screen` events carried it. `$snapshot` events are excluded.
-- **SwiftUI `$screen` event payloads change shape.** Names are now sanitized: `$screen_name = "UIHostingController<MyView>"` → `$screen_name = "MyView"`. **Hard cutover for HogQL filters** like `properties.$screen_name LIKE 'UIHostingController%'` or `properties.$screen_name LIKE 'ModifiedContent%'` — they will start returning zero matches. Customers will need to rewrite those filters using the bare type name.
-- **Type-erased SwiftUI roots stop emitting `$screen`.** Apps whose `body: some View` resolves to `AnyView` (e.g. mixed-type branches returning `AnyView(...)`) will see `$screen` event counts **drop to zero** for those screens — the swizzle still fires, but sanitize returns nil and the event is suppressed. To restore: expose a real type via `.postHogScreenView("MyScreen")` modifier, or call `screen("MyScreen")` manually.
-- **`$screen` event override semantics flipped.** `screen("Home", properties: ["$screen_name": "Override"])` now ships `$screen_name = "Override"` on the `$screen` event. Previously the `screenTitle` arg won (`{ prop, _ in prop }`). This aligns iOS with Android. Customers who passed `$screen_name` defensively in `properties` will see their override take effect.
-- **`screen("")` is silently dropped.** Previously an empty `screenTitle` emitted a `$screen` event with `$screen_name = ""`; now nothing is emitted and the cache is untouched. Customers using empty-string as a sentinel in dashboards will see those rows disappear.
-- **`screen("AnyView")` (literal manual call) is honored.** Only `AnyView` that surfaced from stripping `UIHostingController<...>` / `ModifiedContent<...>` wrappers is dropped (auto-capture noise from `body: some View` type erasure). A caller who literally typed `screen("AnyView")` gets that event through.
-- **Out-of-process crash reports carry `$screen_name` too.** `screen()` now refreshes the crash-replay context snapshot, so an `$exception` event reconstructed from a crash that happened on a given screen will ship with that screen name — matching the behavior of live `captureException(_:)` calls.
--- a/.changeset/screen-name-on-all-events.md
+++ b/.changeset/screen-name-on-all-events.md
@@ -2,11 +2,21 @@
 "posthog-ios": minor
 ---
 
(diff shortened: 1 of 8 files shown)


PostHog/PostHogSDK.swift
```diff
@@ -433,6 +439,13 @@ let maxRetryDelay = 30.0
             }
 
             props["$process_person_profile"] = hasPersonProcessing()
+
+            // Only stamp if the caller didn't supply one — `merging(properties)`
+            // below keeps the existing value on conflict, so seeding would
+            // shadow a caller-supplied override.
+            if let name = lastScreenName, properties?["$screen_name"] == nil {
```
ioannisj: We should be testing for .isEmpty as well, otherwise we may pass an empty string in the props?

PostHog/PostHogSDK.swift
```diff
@@ -1958,6 +1986,7 @@ let maxRetryDelay = 30.0
             didEnterBackgroundToken = nil
             logger?.detach()
             logger = nil
+            lastScreenLock.withLock { _lastScreenName = nil }
```
ioannisj: I think we better move this outside setupLock? Just for symmetry today, but if in the future these locks are acquired in the reverse order somewhere else this may deadlock

PostHogTests/PostHogScreenNameTest.swift
```diff
@@ -0,0 +1,120 @@
...
+
+            sut.reset()
+            sut.close()
+        }
+
+        it("caller-supplied screen_name overrides cached value") {
+            let sut = self.getSut(captured: captured)
+
+            sut.screen("Home")
+            sut.capture("event", properties: ["$screen_name": "Override"])
```
ioannisj, line 72: Should we also add a test for screen() override? 
```
screen("Home", properties: ["$screen_name": "Override"]) 
```
This is expected to carry `"$screen_name": "Override"` right?

PostHog/PostHogSDK.swift
```diff
@@ -38,6 +38,12 @@ let maxRetryDelay = 30.0
     private let cachedPersonPropertiesLock = NSLock()
     private var cachedPersonPropertiesHash: String?
 
+    private let lastScreenLock = NSLock()
+    private var _lastScreenName: String?
+    var lastScreenName: String? {
```
ioannisj, line 43: Can this be private? Do we need it to be internal?

ioannisj, line 43: PostHogLogger maintains its own `lastScreenName`. Could we reuse? Maybe a PostHogScreenNameCache holder so both logger and SDK instance can read from that, or update Logger to read from here?

ioannisj: Depends on https://github.com/PostHog/posthog-ios/pull/609#discussion_r3290700491 I guess

PostHog/PostHogSDK.swift
turnipdabeets: Done in 3bd547f4c — added `!name.isEmpty` on the cached value and `isNilOrEmpty` on the caller-supplied value (treats whitespace-empty as absent so we still fall back to the cache).

PostHog/PostHogSDK.swift
turnipdabeets: Done in 3bd547f4c — moved the `lastScreenLock` clear outside the `setupLock` block, with a one-line comment explaining the lock-ordering motivation.

PostHog/PostHogSDK.swift
turnipdabeets: Done in 3bd547f4c — tightened to `private`. Nothing outside this file reads it.

PostHogTests/PostHogScreenNameTest.swift
turnipdabeets: Added in 4fab84cdd. Also aligned iOS to match Android — caller-supplied `$screen_name` in `properties` now wins on the `$screen` event (previously the `screenTitle` arg won via the iOS-specific `{ prop, _ in prop }` closure). The new test asserts `$screen_name == "Override"` after `screen("Home", properties: ["$screen_name": "Override"])`.

PostHog/PostHogSDK.swift
turnipdabeets: Done in 4fab84cdd. SDK now owns the single cache; `PostHogLogger.lastScreenName` delegates to `PostHogSDK.lastScreenName`. `screen()` sanitizes once and the (sanitized) value goes to both the `$screen` event payload and the cache. Logger no longer subscribes to the screen-view publisher and `detach()` was removed (dead code). Side effect of doing the consolidation this way: SwiftUI `$screen` events now carry the sanitized name too — flagged in the changeset.

ioannisj: > SwiftUI $screen events now carry the sanitized name too — flagged in the changeset.

This will probably affect analytics dashboards/queries for people but I think it's a sensible change

ioannisj (approved):

PostHog/PostHogSDK.swift
turnipdabeets: True, but the string before probably wasn't very helpful. The solution is to use `.postHogScreenView("MyView")`

PostHog/PostHogSDK.swift
```diff
@@ -1148,13 +1177,30 @@ let maxRetryDelay = 30.0
             return
         }
 
+        // Strip SwiftUI wrappers (UIHostingController<X> → X). Drops degenerate
+        // inputs (empty, stripped-AnyView) — for those we skip the $screen event
+        // and leave the cache untouched so the last useful name survives.
+        guard let cleaned = PostHogLogger.sanitize(rawScreenName: screenTitle) else {
```
marandaneto: strange to have the attribute declared in this class, and the sanitize in the PostHogLogger class

PostHog/Logs/PostHogLogger.swift
```diff
@@ -9,51 +9,38 @@ import Foundation
...
     private weak var sdk: PostHogSDK?
-    private let lastScreenLock = NSLock()
-    private var _lastScreenName: String?
-    private var screenViewToken: RegistrationToken?
 
-    /// Latest reported screen name, populated by the screen-view publisher.
-    /// `nil` until the first navigation after SDK setup.
+    /// Latest reported screen name. Delegates to `PostHogSDK.lastScreenName`,
+    /// which is sanitized in `screen()` before being cached.
     var lastScreenName: String? {
```
marandaneto: probably lastScreenName should live in a single place and shared with this class and the posthogsdk instead, right now it feels its dupe

PostHog/PostHogSDK.swift
turnipdabeets: updated

.changeset/screen-name-on-all-events.md
```diff
@@ -0,0 +1,19 @@
...
+- **`screen("")` is silently dropped.** Previously an empty `screenTitle` emitted a `$screen` event with `$screen_name = ""`; now nothing is emitted and the cache is untouched. Customers using empty-string as a sentinel in dashboards will see those rows disappear.
+- **`screen("AnyView")` (literal manual call) is honored.** Only `AnyView` that surfaced from stripping `UIHostingController<...>` / `ModifiedContent<...>` wrappers is dropped (auto-capture noise from `body: some View` type erasure). A caller who literally typed `screen("AnyView")` gets that event through.
+- **Out-of-process crash reports carry `$screen_name` too.** `screen()` now refreshes the crash-replay context snapshot, so an `$exception` event reconstructed from a crash that happened on a given screen will ship with that screen name — matching the behavior of live `captureException(_:)` calls.
```
marandaneto: https://github.com/PostHog/posthog-android/pull/530/changes#r3304363607

marandaneto (approved):

PR #622, opened 2026-05-27
fix(compliance): repair SDK compliance adapter for current test harness

turnipdabeets:

The shared `sdk-test-harness:latest` image added capability gating on the capture suites and a new `feature_flags` suite. The iOS adapter (`sdk_compliance_adapter/`) hadn't caught up, so every PR's compliance report showed **0/16** — only the new flag suite ran (all 404, no `/get_feature_flag` endpoint) and the capture suites were silently skipped for lack of declared capabilities.

This is adapter/CI-tooling only — **no SDK changes**, so no changeset/release.

Ran the harness locally against the adapter (colima + `ghcr.io/posthog/sdk-test-harness:latest`). Result: **0/16 → 33/45** (capture 24/29, feature flags 9/16).

Three adapter fixes:
- **Declare `capabilities: ["capture_v0", "encoding_gzip"]`** in `/health` — re-enables the capability-gated capture suites the harness was skipping.
- **`close()` (not `reset()`) + on-disk storage wipe** on `/reset` and `/init` — `PostHogSDK.reset()` reloads flags as the anon user and races the harness's mock reset (the "Expected 0, got 1" lifecycle failures); `close()` avoids that and lets each test's config (token/host) actually apply, and clearing storage stops queued events/groups leaking across tests.
- **Implement `POST /get_feature_flag`** — newly required by the harness contract.

Remaining failures are genuine SDK behavior or server-vs-mobile semantics (retry status-code policy/backoff timing; `flag_keys_to_evaluate`, `geoip_disable`, `group_properties`, `person_properties.distinct_id` auto-inject), left as known gaps rather than worked around in the adapter.

## :pencil: Checklist


sdk_compliance_adapter/Sources/main.swift
```diff
@@ -15,6 +15,19 @@ class AdapterState {
...
+// leak across tests. close() tears down the SDK but leaves its file-backed queue and
+// storage on disk; the harness resets before every test and expects a clean slate.
+// Mirrors the SDK's own path logic (PostHogStorage): Application Support / <bundleId>,
+// where bundleId falls back to "com.posthog.test" under the -DTESTING build.
+func clearPostHogStorage() {
+    let bundleId = Bundle.main.bundleIdentifier ?? "com.posthog.test"
+    guard let appSupport = FileManager.default.urls(for: .applicationSupportDirectory, in: .userDomainMask).first else {
+        return
+    }
+    try? FileManager.default.removeItem(at: appSupport.appendingPathComponent(bundleId))
```
ioannisj: What I found is that this could still bleed between test cases. 

Instead, what worked before is generating a unique apiKey for each test. Persistence is per apiKey for this SDK, so this way each test gets its own isolated and clean directory to work with. No tear-down races 

Could be a global var, and on init we 

```swift
let apiKey = "\(initReq.apiKey)_\(iteration)"
iteration += 1
```

sdk_compliance_adapter/Sources/main.swift
```diff
@@ -190,8 +294,13 @@ app.get("state") { req async throws -> Response in
 app.post("reset") { req async throws -> Response in
     print("[ADAPTER] POST /reset")
 
-    // Reset the SDK
-    state.posthogSDK?.reset()
+    // Fully tear down the SDK (close, not reset). PostHogSDK.reset() reloads
+    // feature flags as the anonymous user, which races the harness's mock-server
```
ioannisj: What about close() then reset()?

turnipdabeets: I think close() then reset() is functionally identical to just close(). reset no-ops in that case.

sdk_compliance_adapter/Sources/main.swift
turnipdabeets: The problem is that the SDK uses that same apiKey value for the folder it stores files in and the token it stamps on every event and a couple of compliance tests check exactly that token, so this would give us no tear-down races but 2 tests would fail.

turnipdabeets: Updated the adapter to isolate the storage in [121e324](https://github.com/PostHog/posthog-ios/pull/622/commits/121e32456436ad368fa21aeb44baede4e8c7eebe)

sdk_compliance_adapter/Sources/main.swift
```diff
@@ -142,6 +173,86 @@ app.post("capture") { req async throws -> Response in
...
+            case personProperties = "person_properties"
+            case groups
+            case groupProperties = "group_properties"
+            case disableGeoip = "disable_geoip"
+            case forceRemote = "force_remote"
+        }
+    }
+
+    let flagReq = try req.content.decode(FlagRequest.self)
+    print("[ADAPTER] POST /get_feature_flag - key: \(flagReq.key), distinct_id: \(flagReq.distinctId)")
```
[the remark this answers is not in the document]
turnipdabeets: iOS evaluates flags for the current cached user — `getFeatureFlag(key)` takes no per-call `distinct_id` — and the only way to set one, `identify()`, fires its own `/flags` reload, which would break the `assert_flags_request_count: 1` assertion. The tests that assert a caller-supplied `distinct_id` are server-only, so the field is decoded to match the harness contract but intentionally not applied here.

sdk_compliance_adapter/Sources/main.swift
```diff
@@ -142,6 +173,86 @@ app.post("capture") { req async throws -> Response in
...
+
+    // Read the resolved value, capturing the documented $feature_flag_called side effect.
+    let value = sdk.getFeatureFlag(flagReq.key, sendFeatureFlagEvent: true)
+    print("[ADAPTER] Flag \(flagReq.key) resolved to: \(String(describing: value))")
+
+    // Flush that $feature_flag_called event now and wait for it to land. With flushAt=1
+    // it would otherwise auto-flush asynchronously and could arrive in the *next* test's
+    // mock window, inflating $feature_flag_called counts (the side_effect test failure).
+    sdk.flush()
+    try await Task.sleep(nanoseconds: 1_000_000_000)
```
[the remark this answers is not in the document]
turnipdabeets: Fixed in `dcfbebbed` — extracted a shared `flushSettleNanoseconds` (2s) used by both `/flush` and `get_feature_flag`, so the flag handler no longer waits only 1s and the two stay in sync.

sdk_compliance_adapter/Sources/main.swift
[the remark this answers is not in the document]
dustinbyrne: this is another strange side effect of running as a server SDK

sdk_compliance_adapter/Sources/main.swift
```diff
@@ -190,8 +306,13 @@ app.get("state") { req async throws -> Response in
 app.post("reset") { req async throws -> Response in
     print("[ADAPTER] POST /reset")
 
-    // Reset the SDK
-    state.posthogSDK?.reset()
+    // Fully tear down the SDK (close, not reset). PostHogSDK.reset() reloads
+    // feature flags as the anonymous user, which races the harness's mock-server
+    // reset and surfaces as a spurious /flags request (the "Expected 0, got 1"
```
dustinbyrne: Is the expectation wrong? I don't know that I'd consider this to be spurious.

If we `reset()`, we're clearing any active identity and property state from the client. We _need_ to reload flags afterwards, otherwise the feature flag cache would remain stale.

Not totally sure what these comments refer to exactly (it looks like prompt leak), so perhaps I'm missing something, but it does look like we're working around intended behavior

dustinbyrne (approved):
Worth looking into the flags call expectation in regard to `reset()` imo

sdk_compliance_adapter/Sources/main.swift
```diff
@@ -2,6 +2,21 @@ import Foundation
...
+// directory under it) here; HOME covers any path that derives from it. Set before any SDK
+// call so the first storage lookup already uses the sandbox.
+let adapterStorageHome = (NSTemporaryDirectory() as NSString).appendingPathComponent("posthog-ios-compliance-home")
+setenv("CFFIXED_USER_HOME", adapterStorageHome, 1)
+setenv("HOME", adapterStorageHome, 1)
+
+// How long to wait after sdk.flush() for the SDK's async upload to land before returning.
+// Shared by /flush and get_feature_flag so an event flushed in one test can't spill into the
+// next test's mock window (the side-effect timing failure). Based on browser-SDK experience.
+let flushSettleNanoseconds: UInt64 = 2_000_000_000 // 2 seconds
```
dustinbyrne, line 18: Possible to do this deterministically? 2s seems reasonable, but this type of thing can be flaky or slow

turnipdabeets: Done in 85f479847 — replaced the 2s sleep with deterministic tracking.

turnipdabeets: Tried the deterministic approach in 85f479847, reverted in cbe434123. Dropped us from 33 -> 27 passing

PR #629, opened 2026-06-02
fix: always invoke reloadFeatureFlags callback

turnipdabeets:

https://posthoghelp.zendesk.com/agent/tickets/58925 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/61593)

`reloadFeatureFlags(_:)` promises to invoke its callback when the reload finishes, but two early-returns skipped it:

- `if !isEnabled()` — SDK disabled or opted-out
- `remoteConfig?` being nil

When the callback is never invoked, any caller awaiting it stalls. This surfaced via the Flutter SDK, whose `reloadFeatureFlags()` returns a `Future` that completes from this callback — `await Posthog().reloadFeatureFlags()` would hang forever in those states. (Once execution reaches `remoteConfig.reloadFeatureFlags`, the callback already fires on every network outcome, so only the front-door guards were affected.)

The change only adds the callback invocation to the two guard returns (and makes the nil-`remoteConfig` case explicit); the normal path is unchanged. Happy to add a unit test asserting the callback fires when `!isEnabled()` if preferred.

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/PostHog/PostHogSDK.swift
+++ b/PostHog/PostHogSDK.swift
@@ -1849,10 +1849,17 @@ let maxRetryDelay = 30.0
     @objc(reloadFeatureFlagsWithCallback:)
     public func reloadFeatureFlags(_ callback: @escaping () -> Void) {
         if !isEnabled() {
+            // Still invoke the callback so awaiting callers aren't left hanging.
+            callback()
             return
         }
 
-        remoteConfig?.reloadFeatureFlags { _ in
+        guard let remoteConfig else {
+            callback()
+            return
+        }
+
+        remoteConfig.reloadFeatureFlags { _ in
             callback()
         }
     }
--- a/PostHog/PostHogSDK.swift
+++ b/PostHog/PostHogSDK.swift
@@ -1843,9 +1843,10 @@ let maxRetryDelay = 30.0
         }
     }
 
-    /// Reloads feature flags and invokes a callback when the request finishes.
+    /// Reloads feature flags and invokes a callback when finished.
     ///
-    /// - Parameter callback: Called after the reload request completes.
+    /// - Parameter callback: Invoked when the reload finishes, or immediately if the reload
+    ///   is skipped (SDK disabled/opted-out, or no remote config available).
     @objc(reloadFeatureFlagsWithCallback:)
     public func reloadFeatureFlags(_ callback: @escaping () -> Void) {
         if !isEnabled() {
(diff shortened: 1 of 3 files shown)


PostHog/PostHogSDK.swift
```diff
@@ -1849,10 +1849,17 @@ let maxRetryDelay = 30.0
     @objc(reloadFeatureFlagsWithCallback:)
     public func reloadFeatureFlags(_ callback: @escaping () -> Void) {
         if !isEnabled() {
+            // Still invoke the callback so awaiting callers aren't left hanging.
+            callback()
```
marandaneto, line 1853: > Parameter callback: Called after the reload request completes.

if we do this, the method docs have to be fixed

turnipdabeets: Updated docs here. I realize someone might have been using this as a "success", but it seems like `reloadFeatureFlags` doesn't guarantee an update anyway. 

Other alternatives I see, but don't think any are great: 
1. create a new API with a Result in callback, that expands the public API. 
2. add a timeout to flutter to guarantee it completes, that veers from the native API contract and adds a pointless wait.
3. add these same !isEnabled check to flutter, but that leaks implementation.

Curious what you think @ioannisj 

Flutter related link: https://github.com/PostHog/posthog-flutter/pull/409

marandaneto: theres an issue somewhere about adding a new callback with more attributes, i think @ioannisj created after a discussion we had
a success flag, the flags and the error if any
this is a breaking change and requires a major so thats the reason its not done yet

> I realize someone might have been using this as a "success"

yes, thats what i am worried here
i dont think its a big issue tho, we already trigger the callback if the API errored, right? (please confirm) so its a similar problem that already happens

dustinbyrne (approved):
looks good. greptile addressed?

turnipdabeets:
@greptile-apps Addressed in 348da417c51b49c342d457c280dbb95d2708f7fa — added a test asserting the callback fires when the SDK is disabled, covering the `!isEnabled()` guard.

The other guard (`remoteConfig == nil`) isn't separately testable: `remoteConfig` is only nil when the SDK isn't set up, and in that state `!isEnabled()` returns first — so it's a defensive guard that can't be reached independently in a test.

Commit: https://github.com/PostHog/posthog-ios/commit/348da417c51b49c342d457c280dbb95d2708f7fa

PostHog/PostHogSDK.swift
turnipdabeets: confirmed we already trigger the callback if the API errored

PR #630, opened 2026-06-02
fix: keep session replay active across identify/reset

turnipdabeets:

https://posthoghelp.zendesk.com/agent/tickets/58925 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/61593)

After an in-session `identify()` / `reset()`, session replay stopped recording and only resumed after an app restart.

The session recording config comes only from `/config`, fetched once at startup. `reset()` cleared the cached config, so the `/flags` reload that follows had nothing to re-evaluate from and replay stayed off until the next startup.

Error tracking had the identical bug (disabled after `identify()`/`reset()` until restart), so this PR fixes it the same way — bringing iOS to parity with posthog-android #547, which keeps session replay, error tracking and capture-performance config across `reset()`.

Verified on the iOS simulator: `isSessionReplayActive` goes `false → true` after `reset()` → `identify()` → reload, without an app restart. The config is project-level, so keeping it in memory across an identity change doesn't change behavior for disabled projects (cleared on an explicit `false`) or linked-flag projects (re-evaluated against the new user's flags). Storage-survives-reset is covered by tests for both `.sessionReplay` and `.errorTracking`.

## :pencil: Checklist

## If releasing new changes

### DIFF
--- a/PostHog/PostHogRemoteConfig.swift
+++ b/PostHog/PostHogRemoteConfig.swift
@@ -176,9 +176,6 @@ class PostHogRemoteConfig {
                 // process error tracking config
                 self.processErrorTrackingConfig(config)
 
-                // process capture performance (network timing) config
-                self.processCapturePerformanceConfig(config)
-
                 // notify
                 DispatchQueue.main.async {
                     self.onRemoteConfigLoaded.invoke(config)
@@ -222,11 +219,11 @@ class PostHogRemoteConfig {
     }
 
     private func preloadSessionReplay() {
-        var sessionReplay: [String: Any]?
-        var featureFlags: [String: Any]?
-        featureFlagsLock.withLock {
-            sessionReplay = self.storage.getDictionary(forKey: .sessionReplay) as? [String: Any]
-            featureFlags = self.getCachedFeatureFlags()
+        let sessionReplay = remoteConfigLock.withLock {
+            getCachedRemoteConfig()?["sessionRecording"] as? [String: Any]
+        }
+        let featureFlags = featureFlagsLock.withLock {
+            self.getCachedFeatureFlags()
         }
 
         if let sessionReplay = sessionReplay {
@@ -344,14 +341,11 @@ class PostHogRemoteConfig {
                         self.getCachedFeatureFlags() ?? [:]
                     }
 
-                    // /flags is quota-limited; re-arm project-level config (replay / error tracking /
-                    // capture performance) from the cached config so it survives a reset()-then-quota reload.
-                    let remoteConfig = self.remoteConfigLock.withLock { self.getCachedRemoteConfig() }
+                    // quota-limited /flags carries no config; re-arm from the cached remote config
                     #if os(iOS)
-                        self.processSessionRecordingConfig(remoteConfig, featureFlags: cachedFeatureFlags)
+                        self.processSessionRecordingConfig(nil, featureFlags: cachedFeatureFlags)
                     #endif
-                    self.processErrorTrackingConfig(remoteConfig)
-                    self.processCapturePerformanceConfig(remoteConfig)
+                    self.processErrorTrackingConfig(nil)
 
                     self.notifyFeatureFlagsAndRelease(cachedFeatureFlags)
                     return callback(cachedFeatureFlags)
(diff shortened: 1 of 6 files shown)


PostHog/PostHogRemoteConfig.swift
```diff
@@ -438,8 +442,17 @@ class PostHogRemoteConfig {
...
                 }
                 storage.setDictionary(forKey: .sessionReplay, contents: sessionRecording)
+            } else {
+                // /flags doesn't carry sessionRecording; re-evaluate from the cached config instead of clobbering it.
+                sessionReplayLock.withLock {
+                    if let lastConfig = lastSessionRecordingConfig {
+                        sessionReplayFlagActive = isRecordingActive(featureFlags, lastConfig)
+                    }
+                }
             }
```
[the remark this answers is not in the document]
turnipdabeets: Good catch — fixed in 2c40749b3. The re-evaluate branch now also restores `recordingSampleRate` and `recordingMinimumDuration` from the cached config, matching the Map path.

PostHog/PostHogRemoteConfig.swift
```diff
@@ -435,14 +435,27 @@ class PostHogRemoteConfig {
...
                 sessionReplayLock.withLock {
-                    recordingSampleRate = parseSampleRate(sessionRecording["sampleRate"])
-                    recordingMinimumDuration = parseMinimumDuration(sessionRecording["minimumDurationMilliseconds"])
-                    sessionReplayFlagActive = isRecordingActive(featureFlags, sessionRecording)
+                    applySessionRecordingConfig(sessionRecording, featureFlags: featureFlags)
                 }
                 storage.setDictionary(forKey: .sessionReplay, contents: sessionRecording)
+            } else {
+                // /flags doesn't carry sessionRecording; re-evaluate from the persisted config (survives reset) instead of clobbering.
+                if let cached = storage.getDictionary(forKey: .sessionReplay) as? [String: Any] {
```
ioannisj: we could read from the local mem remoteConfig cache directly I think here?

PostHog/PostHogRemoteConfig.swift
```diff
@@ -435,14 +435,27 @@ class PostHogRemoteConfig {
...
+                    sessionReplayLock.withLock {
+                        applySessionRecordingConfig(cached, featureFlags: featureFlags)
+                    }
+                }
             }
         }
 
+        /// Applies a `sessionRecording` config dict to the in-memory replay state (active flag,
+        /// sample rate, minimum duration). The caller must already hold `sessionReplayLock`.
+        private func applySessionRecordingConfig(_ recordingConfig: [String: Any], featureFlags: [String: Any]) {
```
ioannisj: nit: would it be good convention to start suffixing these method names with `applySessionRecordingConfigLocked()` just to make the precondition for holding a lock super obvious on the call site? I'm pretty sure we don't do this in other places throughout, so just a thought

PostHog/PostHogRemoteConfig.swift
ioannisj: I think we are missing the case where `sessionRecording` is Bool type (e.g false, the first if case in this chain) here, right?

ioannisj: Maybe we just need to 

```
let sessionRecording = data?["sessionRecording"] ?? remoteConfig["sessionRecording"]
```

firs thing in this method, then let the existing if-else do its thing

turnipdabeets: local mem does not survive reset()

PostHog/PostHogRemoteConfig.swift
turnipdabeets: sure updated

PostHog/PostHogRemoteConfig.swift
turnipdabeets: updated

ioannisj: given then premise that config is project-scoped, should we keep it on reset() then?

ioannisj (approved):
Just this comment here [pending](https://github.com/PostHog/posthog-ios/pull/630#discussion_r3351353903) but not a blocker for this I think

PostHog/PostHogRemoteConfig.swift
turnipdabeets: updated to use in memory!

PR #635, opened 2026-06-05
ci: run the iOS test target on a simulator and fix the rot it surfaced

turnipdabeets:

**The iOS test target was never running in CI.** `make test` runs `swift test` on macOS, which compiles out every `#if os(iOS)` block, so the session replay, autocapture, tracing-header, and crash-reporting suites never executed. They silently rotted: broken/outdated tests "passed" because they weren't running.

This adds a `test-ios-simulator` CI job that runs the suite on a real simulator, then fixes everything that turned red once the tests actually ran. **No SDK source changes** — pure test/CI infrastructure.

What the run surfaced and we fixed:

- **Crash-processor tests** were a no-op (compiled out because the SDK imports `PHPLCrashReporter` `@_implementationOnly`). Wired the module into the test target so all 13 now run.
- **Replay/sampling tests** still read the old `.sessionReplay` storage slice that #630 removed. Migrated them to the whole-config `.remoteConfig` model.
- **Cross-suite leak:** a React Native test set the global `postHogSdkName` and never restored it, dropping every later test into RN mode (which no-ops sessions) and breaking the replay-trigger + tracing-header suites. Now restored on teardown.
- **Missing survey fixture** in the Xcode project caused a fatal crash under `xcodebuild`. Added it.
- **Flaky queue test** (`flush respecting flushAt`) had a real async race against the background flush. Deflaked it.
- Plus a few async-timing fixes and a missing `testAPIKey` test constant.
- Made `testOniOSSimulator` pick an available simulator dynamically (was pinned to `iPhone 15`).

- Full suite green locally: **564 tests** under `swift test` (macOS), **557 Swift Testing + 161 XCTest** on the iOS simulator.
- Crash-processor suite (13 tests) now runs on both SPM and simulator (was compiled out).
- Ran the previously-flaky queue test **8× consecutively** on the simulator — all green.

## :pencil: Checklist

## If releasing new changes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,6 +80,7 @@ swiftFormat:
 # job — a genuinely broken test fails all 3 and stays red.
 testOniOSSimulator:
 	@device="$$(xcrun simctl list devices available | grep -E '^[[:space:]]*iPhone' | head -1 | sed -E 's/^[[:space:]]*//; s/ \(.*//')"; \
+	[ -n "$$device" ] || { echo "No available iPhone simulator found; install one via Xcode or 'xcrun simctl create'."; exit 1; }; \
 	echo "Testing on simulator: $$device"; \
 	set -o pipefail && xcrun xcodebuild test -scheme PostHog -destination "platform=iOS Simulator,name=$$device" -retry-tests-on-failure -test-iterations 3 | xcpretty
 
--- a/Makefile
+++ b/Makefile
@@ -77,12 +77,14 @@ swiftFormat:
 # -retry-tests-on-failure -test-iterations 3: a few tests assert real-time behaviour (autocapture
 # debounce/flush windows) that can't be made deterministic; on slow, load-variable CI runners those
 # windows occasionally slip. Rerun a *failed* test up to 3 times so a transient miss doesn't fail the
-# job — a genuinely broken test fails all 3 and stays red.
+# job — a genuinely broken test fails all 3 and stays red. Retries can *mask* flakiness, so we tee the
+# raw log to xcodebuild-ios.log; CI reads it back to surface tests that only passed after a retry (the
+# macOS `test` job runs without retries, so a genuine flake still hard-fails there).
 testOniOSSimulator:
 	@device="$$(xcrun simctl list devices available | grep -E '^[[:space:]]*iPhone' | head -1 | sed -E 's/^[[:space:]]*//; s/ \(.*//')"; \
 	[ -n "$$device" ] || { echo "No available iPhone simulator found; install one via Xcode or 'xcrun simctl create'."; exit 1; }; \
 	echo "Testing on simulator: $$device"; \
-	set -o pipefail && xcrun xcodebuild test -scheme PostHog -destination "platform=iOS Simulator,name=$$device" -retry-tests-on-failure -test-iterations 3 | xcpretty
+	set -o pipefail && xcrun xcodebuild test -scheme PostHog -destination "platform=iOS Simulator,name=$$device" -retry-tests-on-failure -test-iterations 3 | tee xcodebuild-ios.log | xcpretty
 
 testOnMacSimulator:
 	set -o pipefail && xcrun xcodebuild test -scheme PostHog -destination 'platform=macOS' | xcpretty
--- a/Makefile
+++ b/Makefile
@@ -73,10 +73,15 @@ swiftLint:
 swiftFormat:
 	swiftformat . --swiftversion 5.3
 
-# use -test-iterations 10 if you want to run the tests multiple times
 # use -only-testing:PostHogTests/PostHogQueueTest to run only a specific test
+# -retry-tests-on-failure -test-iterations 3: a few tests assert real-time behaviour (autocapture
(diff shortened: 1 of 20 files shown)


PostHogTests/PostHogSessionReplayEventTriggersTest.swift
```diff
@@ -25,11 +25,28 @@
             config.disableReachabilityForTesting = true
             config.disableQueueTimerForTesting = true
             config.preloadFeatureFlags = false
+            // Drive recording config from the seeded .remoteConfig below, not the async /config fetch,
+            // so the tests stay deterministic and independent of global stub state from other suites.
+            config.disableRemoteConfigForTesting = true
```
dustinbyrne, line 30: A bunch of tests below call `waitForRemoteConfig`, which looks to break after a 2s timeout. It looks like those calls are essentially just a `sleep` now

PostHogTests/PostHogSessionManagerTest.swift
```diff
@@ -398,6 +398,14 @@ enum PostHogSessionManagerTest {
             posthog = PostHogSDK.with(config)
         }
 
+        deinit {
+            // Restore globals this suite mutates so it doesn't leak RN mode / a mocked clock into the
+            // other serialized suites (a struct can't deinit, hence the class).
+            postHogSdkName = postHogiOSSdkName
+            now = { Date() }
+            DI.main.appLifecyclePublisher = ApplicationLifecyclePublisher.shared
+        }
```
dustinbyrne, line 409: should we close `posthog` as well?

PostHogTests/TestUtils/TestPostHog.swift
```diff
@@ -33,6 +53,14 @@ func waitFlagsRequest(_ server: MockPostHogServer) {
     }
 }
 
+// waitFlagsRequest only proves the /flags request arrived; the SDK processes it async. lastRequestId
+// is set atomically with the flags + v4 metadata under one lock, so wait on it to know the fresh
+// response was fully stored (a flag getter can return a stale cached value before that).
+func waitForFeatureFlagsLoaded(_ server: MockPostHogServer, _ sut: PostHogSDK) {
+    waitFlagsRequest(server)
+    expect(sut.remoteConfig?.lastRequestId).toEventuallyNot(beNil(), timeout: .seconds(10))
```
dustinbyrne: Seems possible that `lastRequestId` could start with a non-nil value

dustinbyrne (approved):
Greptile comments look legit

PostHogTests/PostHogRemoteConfigTest.swift
```diff
@@ -533,14 +534,16 @@ enum PostHogRemoteConfigTest {
...
                 server.returnReplay = true
 
+                await withCheckedContinuation { continuation in
+                    sut.reloadRemoteConfig { _ in continuation.resume() }
+                }
                 await withCheckedContinuation { continuation in
                     sut.loadFeatureFlags(distinctId: "distinctId", anonymousId: "anonymousId", groups: ["group": "value"], callback: { _ in
                         continuation.resume()
                     })
                 }
```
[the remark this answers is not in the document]
turnipdabeets: Addressed in 26600143d: extracted `reloadRemoteConfig` / `loadFeatureFlags` / `reloadConfigThenFlags` helpers and replaced all inline `withCheckedContinuation` reload sites across the replay, linked-flags, error-tracking, and autocapture suites.

Makefile
```diff
@@ -73,10 +73,15 @@ swiftLint:
...
 # use -only-testing:PostHogTests/PostHogQueueTest to run only a specific test
+# -retry-tests-on-failure -test-iterations 3: a few tests assert real-time behaviour (autocapture
+# debounce/flush windows) that can't be made deterministic; on slow, load-variable CI runners those
+# windows occasionally slip. Rerun a *failed* test up to 3 times so a transient miss doesn't fail the
+# job — a genuinely broken test fails all 3 and stays red.
 testOniOSSimulator:
-	set -o pipefail && xcrun xcodebuild test -scheme PostHog -destination 'platform=iOS Simulator,name=iPhone 15,OS=latest' | xcpretty
+	@device="$$(xcrun simctl list devices available | grep -E '^[[:space:]]*iPhone' | head -1 | sed -E 's/^[[:space:]]*//; s/ \(.*//')"; \
+	echo "Testing on simulator: $$device"; \
+	set -o pipefail && xcrun xcodebuild test -scheme PostHog -destination "platform=iOS Simulator,name=$$device" -retry-tests-on-failure -test-iterations 3 | xcpretty
```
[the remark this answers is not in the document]
turnipdabeets: Addressed in 26600143d: added an early `[ -n "$device" ]` guard in the `testOniOSSimulator` target that fails with a clear message when no iOS simulator is available.

PostHogTests/PostHogSessionManagerTest.swift
turnipdabeets: yep, added

PostHogTests/PostHogSessionReplayEventTriggersTest.swift
turnipdabeets: updated

PR #637, opened 2026-06-09
feat(identity): bootstrap anonymousId via config and runtime setter

### LINKED ISSUE #471: Ability to set anonymous ID manually
### Problem Statement

This is related to https://github.com/PostHog/posthog-js/issues/862

It would be great if the anonymous ID could be manually set both on initialize and after reset.

### Solution Brainstorm

_No response_

tsushanth:
## Why

Closes #471 (and mirrors what \`posthog-js#862\` is asking for on the JS side).

Two real users on #471 hit the same wall: \`Application Installed\` and \`Application Updated\` are captured synchronously during \`setup()\`, before there's any chance to call \`identify()\` or to mutate \`PostHogStorageManager\` (which is public for backwards compatibility but not really intended for app code). Those lifecycle events end up tied to the SDK's auto-generated UUID instead of the caller's account ID.

## What

Two additive surfaces. Both honour empty-string as a no-op so a sentinel can't accidentally clobber the stored value.

**1. \`PostHogConfig.anonymousId: String?\`** — set before \`setup()\`. On the first call to \`getAnonymousId()\`, if nothing is persisted, the SDK uses this value instead of generating a UUID v7. Pre-existing \`PostHogConfig.getAnonymousId\` (UUID → UUID hook) is unaffected.

\`\`\`swift
let config = PostHogConfig(apiKey: \"phc_...\")
config.anonymousId = \"A-\\(myUserId)\"
// \"Application Installed\" now reports anonymousId = \"A-...\"
\`\`\`

**2. \`PostHogSDK.setAnonymousId(_:)\`** — runtime override, intended to seed the next anonymous session after \`reset()\`:

\`\`\`swift
PostHogSDK.shared.reset()
PostHogSDK.shared.setAnonymousId(\"A-guest-\\(UUID().uuidString)\")
\`\`\`

This is the iOS equivalent of \`posthog.reset({ bootstrap: { distinctID } })\` from the JS proposal in posthog-js#862.

## Tests

Three new specs in \`PostHogStorageManagerTest\`:
- bootstrap value picked up on fresh install
- empty bootstrap falls back to UUID
- bootstrap re-applies after a \`reset()\` that clears persisted storage

Note: \`swift test\` is currently broken on \`main\` (pre-existing test target rot, see commit 853e71679 — \"ci: run the iOS test target on a simulator and fix the rot it surfaced\"). The SDK target builds clean; the new tests follow the existing Quick/Nimble pattern and should pass on the iOS-simulator CI path.

## Compatibility

- No public API removals, no signature changes.
- Existing call sites compile unchanged (both new surfaces have defaults / are additive).
- \`@objc\` annotations preserved on both new members.

### DIFF
--- a/PostHog/PostHogBootstrap.swift
+++ b/PostHog/PostHogBootstrap.swift
@@ -1,0 +1,45 @@
+//
+//  PostHogBootstrap.swift
+//  PostHog
+//
+
+import Foundation
+
+/// Pre-seeded values applied on the very first SDK launch when no per-device state has
+/// been persisted yet.
+///
+/// Set ``PostHogConfig/bootstrap`` before calling `setup(_:)` to ensure events captured
+/// synchronously during initialization (`Application Installed` / `Application Updated`,
+/// pre-identify lifecycle events) carry a caller-controlled `$distinct_id` rather than
+/// the SDK-generated UUID. Mirrors the [`bootstrap` option in `posthog-js`](https://posthog.com/docs/feature-flags/bootstrapping).
+///
+/// Bootstrap only seeds the very first session. Once an anonymous ID is persisted on
+/// disk, or `identify(...)` has been called, the bootstrap values are ignored — they
+/// never override an already-identified user or re-link traffic across a previous
+/// anon→identified merge.
+@objc(PostHogBootstrap) public class PostHogBootstrap: NSObject {
+    /// The distinct ID to seed on first launch.
+    ///
+    /// When ``isIdentifiedID`` is `false` (the default), this becomes the anonymous ID —
+    /// the `$distinct_id` on pre-identify events. When `true`, it is treated as an
+    /// already-identified user's distinct ID and the SDK skips the `$identify` merge
+    /// for it.
+    @objc public var distinctID: String?
+
+    /// Whether ``distinctID`` represents an already-identified user.
+    ///
+    /// Defaults to `false`. Set to `true` when the host application has resolved the
+    /// user's identity outside the SDK (for example from a backend session token) and
+    /// wants the iOS SDK to treat them as identified from the first event onward.
+    @objc public var isIdentifiedID: Bool = false
+
+    @objc public override init() {
+        super.init()
+    }
+
+    @objc public init(distinctID: String?, isIdentifiedID: Bool = false) {
+        self.distinctID = distinctID
+        self.isIdentifiedID = isIdentifiedID
+        super.init()
+    }
+}
--- a/PostHog/PostHogConfig.swift
+++ b/PostHog/PostHogConfig.swift
@@ -187,6 +187,16 @@ public typealias BeforeSendBlock = (PostHogEvent) -> PostHogEvent?
     /// - Returns: The UUID to persist as the anonymous ID.
     @objc public var getAnonymousId: ((UUID) -> UUID) = { uuid in uuid }
 
+    /// Pre-seeded anonymous ID used on the first call to `getAnonymousId()` when no
+    /// persisted value exists.
+    ///
+    /// Set this before `setup()` to ensure events captured synchronously during initialization
+    /// (such as `Application Installed` / `Application Updated`) use your own ID rather than
+    /// the SDK's auto-generated UUID. Ignored when an anonymous ID is already persisted.
+    ///
+    /// Defaults to `nil` (use the SDK-generated UUID).
+    @objc public var anonymousId: String?
+
     /// Flag to reuse the anonymous Id between `reset()` and next `identify()` calls
     ///
     /// If enabled, the anonymous Id will be reused for all anonymous users on this device,
--- a/PostHog/PostHogConfig.swift
+++ b/PostHog/PostHogConfig.swift
@@ -187,15 +187,21 @@ public typealias BeforeSendBlock = (PostHogEvent) -> PostHogEvent?
     /// - Returns: The UUID to persist as the anonymous ID.
     @objc public var getAnonymousId: ((UUID) -> UUID) = { uuid in uuid }
 
-    /// Pre-seeded anonymous ID used on the first call to `getAnonymousId()` when no
-    /// persisted value exists.
+    /// Pre-seeded identity applied on the very first SDK launch when no per-device state
+    /// has been persisted yet.
     ///
-    /// Set this before `setup()` to ensure events captured synchronously during initialization
-    /// (such as `Application Installed` / `Application Updated`) use your own ID rather than
-    /// the SDK's auto-generated UUID. Ignored when an anonymous ID is already persisted.
+    /// Set this before calling `setup(_:)` to ensure events captured synchronously during
+    /// initialization (`Application Installed` / `Application Updated`, pre-identify
+    /// lifecycle events) carry a caller-controlled `$distinct_id` rather than the
+    /// SDK-generated UUID. Mirrors the [`bootstrap` option in `posthog-js`](https://posthog.com/docs/feature-flags/bootstrapping).
+    ///
+    /// Bootstrap only seeds the very first session. Once an anonymous ID is persisted
+    /// on disk, or `identify(...)` has been called, the bootstrap values are ignored —
+    /// they never override an already-identified user or re-link traffic across a
+    /// previous anon→identified merge.
     ///
     /// Defaults to `nil` (use the SDK-generated UUID).
-    @objc public var anonymousId: String?
+    @objc public var bootstrap: PostHogBootstrap?
 
     /// Flag to reuse the anonymous Id between `reset()` and next `identify()` calls
     ///
--- a/PostHog/PostHogStorageManager.swift
+++ b/PostHog/PostHogStorageManager.swift
@@ -21,7 +21,7 @@ public class PostHogStorageManager {
     private let identifiedLock = NSLock()
     private let personProcessingLock = NSLock()
     private let idGen: (UUID) -> UUID
-    private let bootstrapAnonymousId: String?
+    private let bootstrap: PostHogBootstrap?
 
     private var distinctId: String?
     private var cachedDistinctId = false
@@ -33,7 +33,33 @@ public class PostHogStorageManager {
     init(_ config: PostHogConfig) {
         storage = PostHogStorage(config)
         idGen = config.getAnonymousId
-        bootstrapAnonymousId = config.anonymousId
+        bootstrap = config.bootstrap
+        applyBootstrapIfNeeded()
+    }
+
+    /// Persists the bootstrap distinct ID exactly once, on the very first launch with no
+    /// per-device state. Skipped when the device already has an anonymous ID on disk,
+    /// when the user is already identified, or when the caller did not supply one.
+    ///
+    /// When `bootstrap.isIdentifiedID` is `true`, the value is treated as the
+    /// already-identified distinct ID — both `.anonymousId` and `.distinctId` are seeded
+    /// to the same value and `isIdentified` is set, so subsequent events are emitted on
+    /// the identified person without an `$identify` merge.
+    private func applyBootstrapIfNeeded() {
+        guard let bootstrap, let bootstrapId = bootstrap.distinctID, !bootstrapId.isEmpty else {
+            return
+        }
+        // Persisted state wins — never override an existing anonymous ID, and never
+        // re-link traffic across a previous anon→identified merge.
+        if storage.getString(forKey: .anonymousId) != nil { return }
+        if storage.getBool(forKey: .isIdentified) == true { return }
+
+        setAnonymousId(bootstrapId)
+
+        if bootstrap.isIdentifiedID {
+            setDistinctId(bootstrapId)
+            setIdentified(true)
+        }
     }
 
     /// Returns the persisted anonymous ID, creating one if needed.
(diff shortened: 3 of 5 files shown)


dustinbyrne (commented):
Hey @tsushanth, thank you for the pull request!

You mentioned that application lifecycle events are being recorded with an anonymous identifier. Once `identify()` is called, those anonymous events should become linked with the identified person. It sounds like this isn't what you want, so out of curiosity, could you describe what the specific gap is? It'd be helpful for me to understand if there's a particular usage pattern that we're not accounting for here.

tsushanth:
Hey @dustinbyrne — thanks for the look.

You're right that once `identify()` runs, anonymous events become linked to the identified person via `$anon_distinct_id`. The gap is the events captured **before** `identify()` — they're still attributed to the SDK-generated UUID, and there are a few flows where the host app already knows a better anonymous ID at install time:

**1. Cross-SDK / cross-platform funnels.** Apps that ship a backend or a web SDK alongside iOS often want a single anonymous identity across all three before the user signs in. With `bootstrap`, web hands the iOS app its `anonymous_id` cookie value (via a deep link or login handoff) and iOS picks up the same key from the very first event, rather than generating a throwaway UUID that has to be merged later. This matches what `posthog-js` and `posthog-android` already do.

**2. Stable feature-flag bucketing pre-identify.** Feature flags evaluated before sign-in bucket by the anonymous ID. When that ID is deterministic from the app's own auth/install attribution layer, A/B exposures are stable across reinstalls (where the iOS SDK would otherwise mint a new UUID and shuffle the bucket).

**3. `Application Installed` / `Application Updated` events.** Those fire synchronously during `setup()` — there's no clean window to call `setAnonymousId` before they go out. Bootstrap is the only way to land them on a caller-controlled ID without monkey-patching the lifecycle hooks.

The PR is the matching iOS surface for the `bootstrap` option that already exists on the JS/Android SDKs, so this is mostly an SDK-family-parity move.

---

**Re: the Greptile note about `reset()` semantics** — that's a fair callout and worth tightening in the doc. As written, `config.anonymousId` is re-applied on any session where storage is empty, which includes the post-`reset()` (e.g. logout) case, not just first install. For someone who bootstraps with something like an interpolated `"A-<userId>"` value and then calls `reset()` on logout, the next anonymous session re-seeds the same ID and silently re-links activity to the prior identity.

Two cleanish options if you'd like me to follow up:

- **(a)** Doc-only: explicitly call out that `config.anonymousId` is reused on every fresh (post-`reset()`) session, and steer callers who want install-only behavior toward setting it once via `setAnonymousId(...)` from app code after detecting first launch.
- **(b)** Track a one-shot "bootstrap_applied" flag in persistent storage so `config.anonymousId` only seeds genuinely first-install sessions. (`posthog-android` opted for (a) — happy to match.)

Happy to push either as a follow-up commit on this branch — let me know your preference.

dustinbyrne:
Hi @tsushanth, thanks for that information!

What I'm hoping to understand is your specific use case in which `identify()` and merge not are ideal. Regarding feature flag evaluations, there are some different strategies for maintaining stable flag evaluations across `distinct_id` changes (e.g., device bucketing, persisted feature flags, using `setPropertiesForFlags`). How would you be seeding a known anonymous identifier on initialization?

> The PR is the matching iOS surface for the bootstrap option that already exists on the JS/Android SDKs, so this is mostly an SDK-family-parity move.

iOS and Android both have `getAnonymousId` hooks which can be set on the configuration used when initializing PostHog. Bootstrapping the JS client is somewhat different in that you can provide a `distinct_id` and indicate whether it's identified or anonymous. This is somewhat of a more common case in web given that it's trivial to seed a known identity via server-side rendering, query parameters, etc.

Sorry for all the questions, I'm mainly trying to understand if this is part of a broader workflow we don't have great support for, or if it's more of a targeted fix specifically in response to https://github.com/PostHog/posthog-ios/issues/471

tsushanth:
Fair point — `getAnonymousId: ((UUID) -> UUID)` does cover the *customise the UUID* case, and you're right that feature-flag stability has cleaner answers via persisted flags / `setPropertiesForFlags` / device bucketing. Let me sharpen the gap I'm actually trying to close:

The hook's signature constrains the anonymous ID to a `UUID`. That's fine when the consumer's "known ID" is itself a UUID (or can be deterministically hashed into one), but the case I keep running into is when the external system hands me a **non-UUID string** that needs to land on the wire exactly as-is for cross-system joins:

- Web hands iOS its `ph_anonymous_id` cookie value via a Universal Link or post-login redirect (e.g. `"A-r3K7-cookie-…"`).
- A backend issues a stable opaque per-device key for unauth'd telemetry (a JWT `sub` for an "anonymous session" token, or a server-generated `device_session_id`).
- A first-party attribution SDK (AppsFlyer / Adjust install referrer) returns a known string that the server-side analytics pipeline keys off.

In each of those, I want PostHog's first event from iOS to carry that **exact string** as `$distinct_id`, so downstream joins (web↔iOS in the same dashboard, server logs↔PostHog events) work without a `.identify()` merge step or a post-hoc reconciliation table. `getAnonymousId` can't accept those — I'd have to derive a UUID from the string, at which point the on-wire ID no longer matches what the other system sees, and the join is broken.

Concrete shape:

```swift
// Web has already set the user as anonymous_id = "A-r3K7-cookie-2A8F"
// and handed it to iOS via a Universal Link.
let cookieFromWeb = deepLink.queryItem("anon")  // "A-r3K7-cookie-2A8F"

let config = PostHogConfig(apiKey: …)
config.anonymousId = cookieFromWeb   // ← what this PR adds
// First event from iOS carries $distinct_id = "A-r3K7-cookie-2A8F", same as web
```

Without that, the same flow needs either (a) a synchronous `identify()` immediately after `setup()` (which produces a `$identify` merge event and bumps the user record, even though there's no real identification happening), or (b) a custom workaround using `setAnonymousId` after the SDK has already emitted lifecycle events on the SDK-generated UUID.

**Scope honesty:** yes, this is targeted at #471 — but the issue itself was filed against the same gap (the reporter wanted to seed an ID matching their web/server tracking on init). I think it generalises cleanly because it mirrors what `posthog-js` already exposes via `bootstrap.distinctID` and what `posthog-android` exposes via `PostHogConfig#getAnonymousId` (which on Android *does* accept a `String?`, not a `UUID` transform — that platform asymmetry is what made me reach for this).

If you'd rather not expand the iOS surface and the recommended path is "use `setAnonymousId` after init and accept the brief window where lifecycle events go out on a throwaway UUID," I'm happy to close — would just want to capture that as guidance in the doc so the next person who hits #471 sees it. Or if you think the right shape is different from what I drafted (e.g. matching JS's `bootstrap` struct with explicit `isIdentified`), happy to refactor.

marandaneto:
i think the public api should match https://posthog.com/docs/feature-flags/bootstrapping
so its an object that takes more things to bootstrap and not only the anon id
also i'd not expose any other public api besides the bootstrap config unless its needed (does not look like it)

marandaneto:
another concern (is it?) is if the user is already identified with a stable distinct id and a merged anon id, and then someone starts using the bootstraped distinct if after the user is identified, not sure if this would mess with the user merging
maybe we should use the bootstraped distinct id only if not identified yet?

tsushanth:
Both calls land cleanly — pushed [`ea0e05614`](https://github.com/PostHog/posthog-ios/pull/637/commits/ea0e05614) on this branch with the refactor.

**Public API → `Bootstrap` struct matching JS.** Dropped `config.anonymousId: String?` and the runtime `setAnonymousId(_:)` method entirely. Replaced with one optional config field:

```swift
@objc public var bootstrap: PostHogBootstrap?
```

`PostHogBootstrap` mirrors the documented [`bootstrap` shape in posthog-js](https://posthog.com/docs/feature-flags/bootstrapping):

```swift
@objc(PostHogBootstrap) public class PostHogBootstrap: NSObject {
    @objc public var distinctID: String?
    @objc public var isIdentifiedID: Bool = false
}

// Usage
let config = PostHogConfig(apiKey: "...")
config.bootstrap = PostHogBootstrap(distinctID: cookieFromWeb)              // anonymous-only seed
config.bootstrap = PostHogBootstrap(distinctID: knownUserId, isIdentifiedID: true)  // already-identified
```

I kept the surface deliberately narrow per your note — just `distinctID` and `isIdentifiedID` for now. Happy to extend with `featureFlags` / `featureFlagPayloads` if you want JS feature-parity, but figured it's better as a follow-up if and when feature-flag bootstrap shows up as a real consumer request.

**Don't clobber an already-identified user.** Added a second gate in `applyBootstrapIfNeeded()`:

```swift
if storage.getString(forKey: .anonymousId) != nil { return }
if storage.getBool(forKey: .isIdentified) == true { return }
```

The first gate (existing anon ID wins) was already in the original PR; the second is new. So bootstrap only seeds the very first session and never re-links traffic across a prior anon→identified merge. When `isIdentifiedID: true`, `.anonymousId`, `.distinctId`, and `.isIdentified` are all seeded together so the SDK emits subsequent events on the identified person without an `$identify` merge — matches JS's `isIdentifiedID: true` semantics.

Two new tests cover the gates: one for *"a second SDK init must not override the persisted anonymous ID"*, one for *"a second SDK init with `isIdentifiedID: true` must not re-link an already-identified user."* Existing tests updated to the new surface; all 564 pass under `make test`.

Let me know if you'd like a different field naming convention (e.g. `distinctId` lowercase-d to match Swift conventions over the JS `distinctID`), or if you want me to drop `isIdentifiedID` and keep the surface even narrower for the first ship.

PostHog/PostHogBootstrap.swift
```diff
@@ -0,0 +1,45 @@
...
+    /// already-identified user's distinct ID and the SDK skips the `$identify` merge
+    /// for it.
+    @objc public var distinctID: String?
+
+    /// Whether ``distinctID`` represents an already-identified user.
+    ///
+    /// Defaults to `false`. Set to `true` when the host application has resolved the
+    /// user's identity outside the SDK (for example from a backend session token) and
+    /// wants the iOS SDK to treat them as identified from the first event onward.
+    @objc public var isIdentifiedID: Bool = false
```
ioannisj, line 34: Let's rename these to follow naming conventions for this repo

```swift
@objc public var distinctId: String
@objc public var isIdentified: Bool
```

PostHog/PostHogBootstrap.swift
```diff
@@ -0,0 +1,45 @@
...
+
+    @objc public override init() {
+        super.init()
+    }
+
+    @objc public init(distinctID: String?, isIdentifiedID: Bool = false) {
+        self.distinctID = distinctID
+        self.isIdentifiedID = isIdentifiedID
+        super.init()
+    }
```
ioannisj, line 44: I don't think there is a need for an empty constructor here and a default parameter (default does not bridge to objc anw)
I would make both params required since the root config is optional anyway

ioannisj: If you want we can drop a convenience init for anonymous ids just to make the intent super clear, but imho it's probably not needed just yet. Up to you

```swift
@objc public convenience init(anonymousId: String) {
    self.init(distinctId: anonymousId, isIdentified: false)
}
```

PostHog/PostHogConfig.swift
```diff
@@ -187,6 +187,22 @@ public typealias BeforeSendBlock = (PostHogEvent) -> PostHogEvent?
     /// - Returns: The UUID to persist as the anonymous ID.
     @objc public var getAnonymousId: ((UUID) -> UUID) = { uuid in uuid }
 
+    /// Pre-seeded identity applied on the very first SDK launch when no per-device state
+    /// has been persisted yet.
+    ///
+    /// Set this before calling `setup(_:)` to ensure events captured synchronously during
+    /// initialization (`Application Installed` / `Application Updated`, pre-identify
+    /// lifecycle events) carry a caller-controlled `$distinct_id` rather than the
+    /// SDK-generated UUID. Mirrors the [`bootstrap` option in `posthog-js`](https://posthog.com/docs/feature-flags/bootstrapping).
```
ioannisj, line 196: Link is a bit confusing cause it's linking to feature flag bootstrapping. I think we can remove

PostHog/PostHogConfig.swift
```diff
@@ -187,6 +187,22 @@ public typealias BeforeSendBlock = (PostHogEvent) -> PostHogEvent?
...
+    ///
+    /// Bootstrap only seeds the very first session. Once an anonymous ID is persisted
+    /// on disk, or `identify(...)` has been called, the bootstrap values are ignored —
+    /// they never override an already-identified user or re-link traffic across a
+    /// previous anon→identified merge.
+    ///
```
ioannisj, line 202: Let's also mention that an empty id will be a no-op

PostHog/PostHogStorageManager.swift
```diff
@@ -32,6 +33,33 @@ public class PostHogStorageManager {
...
+        // Persisted state wins — never override an existing anonymous ID, and never
+        // re-link traffic across a previous anon→identified merge.
+        if storage.getString(forKey: .anonymousId) != nil { return }
+        if storage.getBool(forKey: .isIdentified) == true { return }
+
+        setAnonymousId(bootstrapId)
+
+        if bootstrap.isIdentifiedID {
+            setDistinctId(bootstrapId)
+            setIdentified(true)
```
ioannisj, line 61: we probably need `if config.personProfiles != .never { setPersonProcessing(true) }` here as well cc @marandaneto

ioannisj (commented):
Left a couple of comments. You'd want to merge main, add a changeset entry and regenerate public api with `make apiUpdate` to make CI happy

ioannisj:
Hey @tsushanth, any update on this one? Happy to help if you’re blocked somewhere

ioannisj:
@tsushanth I'll pick this up and expand on bootstrapping flags as well. Thanx for the contribution!

tsushanth:
Closing — inactive. Happy to revisit if there's interest in this feature direction.

PR #656, opened 2026-06-12
feat: add exception steps to error tracking

ioannisj:

Adds **exception steps** to error tracking.

Exception steps are breadcrumb-style context records that accumulate over time and attach to the next captured `$exception` as `$exception_steps`, giving the error tracking UI a timeline of what happened right before the exception.

```swift
PostHogSDK.shared.addExceptionStep("User tapped Checkout", properties: ["screen": "cart"])
```

### What's included

- **`addExceptionStep(_:properties:)`** public API (with an `@objc` no-properties overload).
- **`PostHogExceptionStepsBuffer`** — thread-safe FIFO buffer bounded by a UTF-8 byte budget; oldest steps evicted first (steps closest to the exception are the most valuable), a single oversized step is rejected.
- **`PostHogExceptionStepsConfig`** on `errorTrackingConfig` — `enabled` (default `true`) and `maxBytes` (default `32768`).
- **Capture wiring** — buffered steps attach to a `$exception` only when the caller didn't already provide `$exception_steps`; the buffer clears after a successful capture and is preserved when the event is dropped by `beforeSend`.
- **Fatal-crash durability** — steps are persisted with the crash context so they survive a hard crash and attach to the crash `$exception` reported on the next launch.

- Added `PostHogExceptionStepsBufferTest` (buffer FIFO / byte-budget eviction / reserved-key validation) and `PostHogExceptionStepsTest` (attach / no-overwrite / clear-on-capture / preserve-on-drop / disabled no-op).
- Added an exception-steps demo to the sample app.

CleanShot 2026-06-17 at 12 07 30@2x

## :pencil: Checklist

## If releasing new changes


PostHog/PostHogSDK.swift
```diff
@@ -1250,9 +1268,16 @@ let maxRetryDelay = 30.0
...
         guard let posthogEvent else {
+            // Event was dropped (e.g. by beforeSend). Preserve the buffer so steps roll forward to
+            // the next exception rather than being lost.
             return
         }
 
+        // The exception survived beforeSend → clear the now-flushed steps so they aren't re-attached.
+        if isExceptionEvent {
+            clearExceptionStepsAfterCapture()
+        }
```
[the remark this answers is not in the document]
ioannisj: yes, known working on this

PostHog/PostHogSDK.swift
```diff
@@ -2459,6 +2484,107 @@ let maxRetryDelay = 30.0
...
+    }
+
+    /// Clear the exception-steps buffer after an exception was successfully captured (survived
+    /// `beforeSend`), and refresh the persisted crash context so it no longer holds the flushed steps.
+    /// Runs on the serial queue, ordered after any in-flight `addExceptionStep`.
+    private func clearExceptionStepsAfterCapture() {
+        exceptionStepsQueue.async { [weak self] in
+            guard let self, let buffer = self.exceptionStepsBuffer, !buffer.isEmpty else { return }
+            buffer.clear()
+            self.notifyContextDidChange()
```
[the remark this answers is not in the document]
ioannisj: Also known and working on this

PostHog/PostHogSDK.swift
```diff
@@ -2442,6 +2469,108 @@ let maxRetryDelay = 30.0
...
+            return
+        }
+
+        // Timestamp at call time (before dispatch) so the breadcrumb timeline is accurate.
+        let timestamp = toISO8601String(now())
+
+        // Build and buffer the step on the serial queue so the caller is never blocked. Re-read the
+        // buffer inside the closure so a step enqueued just before close() becomes a no-op instead of
+        // writing into the cleared directory.
+        exceptionStepsQueue.async { [weak self] in
```
dustinbyrne: Considering this happens asynchronously, if we call `addExceptionStep` right before a fatal exception, is it reasonable to assume that we'd lose that last exception step?

Should synchronous persistence be an option?

PostHog/PostHogSDK.swift
```diff
@@ -2442,6 +2469,108 @@ let maxRetryDelay = 30.0
...
+    }
+
+    /// The session-scoped buffered steps to attach to an exception, or `nil` when disabled or empty.
+    private var attachableExceptionSteps: [[String: Any]]? {
+        guard config.errorTrackingConfig.exceptionSteps.enabled,
+              let buffer = exceptionStepsBuffer
+        else { return nil }
+        let steps = buffer.getAttachable()
+        return steps.isEmpty ? nil : steps
+    }
```
dustinbyrne, line 2568: Do these need to be synchronized? It's also used in `addExceptionStep` and modified in `clear` without a lock

dustinbyrne (commented):
Overall the implementation looks good, I left a couple of comments. I'm not super familiar with the requirements of this API, so probably worth @PostHog/team-error-tracking approving

PostHog/PostHogSDK.swift
```diff
@@ -48,6 +48,11 @@ let maxRetryDelay = 30.0
     }
 
     private var queue: PostHogQueue<PostHogEvent>?
+    private var exceptionStepsBuffer: PostHogExceptionStepsBuffer?
```
turnipdabeets: It looks like `exceptionStepsBuffer` is accessed from two different synchronization contexts with no common lock. The live race is in optIn()

PostHog/ErrorTracking/PostHogExceptionStepsBuffer.swift
```diff
@@ -0,0 +1,154 @@
...
+            if let filename = removed.filename { evicted.append(filename) }
+        }
+        return evicted
+    }
+
+    private func persist(_ data: Data) -> String? {
+        guard let directory else { return nil }
+        let filename = UUID.v7().uuidString
+        do {
+            try data.write(to: directory.appendingPathComponent(filename))
```
turnipdabeets: It look like files are not excluded from iCloud backup, is that intentional?

ioannisj:
Thanx folks. I'll circle back and address the comments after opening draft PRs for the other SDKs (may need to adjust things for Flutter, RN)

PostHog/ErrorTracking/PostHogExceptionStepsBuffer.swift
ioannisj: Oh that's actually a really good point. I'll push a fix for this buffer. However, none of our disk-backed queues are excluded from backup currently (depending on the location we store them). I think it should cause it could lead to duplicate events being ingested right?

ioannisj: Actually I'll exclude the whole app folder once on it's own PR, I think it's cleaner that way cause it will have sdk-wide effect and we can discuss about it more there

ioannisj: https://github.com/PostHog/posthog-ios/issues/664

PostHog/PostHogSDK.swift
ioannisj: True, I'll leave this unresolved though cause it depends on https://github.com/PostHog/posthog-ios/pull/656#discussion_r3415879951

PostHog/PostHogSDK.swift
ioannisj: Yeah I think that's the case. So step A, step B, step C -> exception (manual or crash) would risk dropping step C since it may not have finished writing to disk yet. However, with the current one-step-per-file persistence, sync write would probably be too slow and blocking, so I would not add it as an option for it. 

I'll see if we can use PLCrashReporter's `customData` instead of our custom persistence since I think PLCrashReporter holds that in memory and flushes to disk only at crash time, which is ideal. Setting that could be sync

PostHog/PostHogSDK.swift
ioannisj: resolved with persistence refactor

PostHog/PostHogSDK.swift
ioannisj: Updated so that now these are sync and appended in PLCrashReporter's `customData` instead which is in-memory and only flushed to disk as part of the crash report

PostHog/PostHogSDK.swift
```diff
@@ -2442,6 +2476,94 @@ let maxRetryDelay = 30.0
...
+    /// ```swift
+    /// PostHogSDK.shared.addExceptionStep("User tapped Checkout", properties: ["screen": "cart"])
+    /// ```
+    ///
+    /// - Parameters:
+    ///   - message: A short, non-empty description of what happened. Empty messages are ignored.
+    ///   - properties: Optional additional context to attach to the step.
+    ///
+    @objc(addExceptionStep:properties:)
+    public func addExceptionStep(_ message: String, properties: [String: Any]? = nil) {
```
hpouillot, line 2504: Not an iOS expert but apparently this method is blocking ? Just to be sure it is intentional since the PR description states otherwise

hpouillot (approved):

PostHog/PostHogSDK.swift
ioannisj: Yeah stale PR description, updated. This is intentional as discussed here https://github.com/PostHog/posthog-ios/pull/656#discussion_r3415879951

PostHog/PostHogSDK.swift
```diff
@@ -2131,6 +2157,9 @@ let maxRetryDelay = 30.0
 
         setupLock.withLock {
             installIntegrations()
+            // Start the buffer for this run, then notify it of the current context for crash reporting.
+            createExceptionStepsBufferIfNeeded()
+            notifyContextDidChange()
```
turnipdabeets, line 2164: **Bug:** `notifyContextDidChange()` fires `onEventContextChanged` so the new crash writer gets identity context, but `onExceptionStepsChanged` is never fired — new subscribers don't get replayed state. Until the next `addExceptionStep()` call, `crashReporter.customData` has context but empty steps. A crash in that window loses all buffered steps from the on-disk report.

Fix:
```swift
notifyContextDidChange()
if let steps = exceptionStepsBuffer?.getAttachable(), !steps.isEmpty {
    onExceptionStepsChanged.invoke(steps)
}
```

ioannisj: Right, setup path should be okay but this affects the opt-out/opt-in path. 
Fixed in [d77f35d](https://github.com/PostHog/posthog-ios/pull/656/commits/d77f35d54806c55ec73daf4f50170971b0e8d33a) and added a couple of more tests

PR #658, opened 2026-06-12
chore(ci): public API snapshot check

marandaneto:

Public API changes should be easy to spot during review. This adds a generated Swift public API snapshot and a pull request CI check that compares the current iOS symbol graph against the checked-in snapshot.

The checker builds the PostHog scheme for iOS, extracts a symbol graph from the built module, and compares it against the snapshot so iOS-specific public API is covered instead of only the default macOS SwiftPM API surface.

This also keeps the iOS simulator retry behavior working with newer Xcode: if xcodebuild exits non-zero after reporting zero test failures, or only because a flaky test failed before passing on retry, the job is treated as successful while genuine unresolved test failures still fail.

- [x] `make apiUpdate`
- [x] `make apiCheck`
- [x] `scripts/check-ios-test-result.sh` with synthetic retried, zero-failure, and unresolved failure logs
- [x] `make -n testOniOSSimulator`

## :pencil: Checklist

## If releasing new changes


.github/workflows/build.yml
```diff
@@ -43,6 +43,18 @@ jobs:
       - name: Build SDK (${{ matrix.platform }})
         run: ${{ matrix.command }}
 
+  public-api:
+    if: github.event_name == 'pull_request'
```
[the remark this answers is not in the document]
marandaneto: Thanks for the suggestion. This check is intentionally scoped to pull requests: it is meant to be a PR gate for proposed SDK changes, and this repository does not need the extra main-push job for this workflow. Leaving the PR-only guard as-is.

scripts/check-public-api.sh
```diff
@@ -0,0 +1,106 @@
...
+    rm -f "$TMP_FILE"
+}
+trap cleanup EXIT
+
+cd "$ROOT_DIR"
+if [[ -d "$ROOT_DIR/.build" ]]; then
+    find "$ROOT_DIR/.build" -path '*/symbolgraph' -type d -prune -exec rm -rf {} +
+fi
+
+if ! swift package dump-symbol-graph --minimum-access-level public --skip-synthesized-members >/dev/null; then
```
ioannisj: This will generate only macOS api, so we'll miss the public apis from other platforms

marandaneto: maybe only iOS is enough since its just a safe guard, let me check if its possible to do iOS only

marandaneto: Good catch — fixed in 41e531f69 by generating the snapshot from an iOS xcodebuild symbol graph instead of SwiftPM's default macOS symbol graph, and updated the checked-in snapshot.

ioannisj:
@marandaneto CI not happy?

marandaneto:
> @marandaneto CI not happy?

it was an unrelated flaky test, in cases like this, approve to unblock pls, i can always fix with a follow up

ioannisj (approved):

ioannisj:
> > @marandaneto CI not happy?
> 
> it was an unrelated flaky test, in cases like this, approve to unblock pls, i can always fix with a follow up

My bad thought I saw automerge on

marandaneto:
https://github.com/PostHog/posthog-ios/pull/660#issuecomment-4708805831

PR #661, opened 2026-06-15
fix: respect remote session replay sampling

marandaneto:
Remote session replay sample rates can arrive after the replay integration has already started. In that case, the SDK defaulted to recording the session and did not re-check sampling once remote config loaded, so dashboard-configured sampling could be ignored for the current session.

This updates replay to re-evaluate sampling after remote config is applied when replay is already running.

- Added an iOS regression test covering remote sample rate `0` stopping an already-started replay integration after config loads.
- Ran `make testOniOSSimulator`.

## :pencil: Checklist

## If releasing new changes

### DIFF
--- a/PostHog/Replay/PostHogReplayIntegration.swift
+++ b/PostHog/Replay/PostHogReplayIntegration.swift
@@ -433,9 +433,13 @@
         }
 
         func applyRemoteConfig(remoteConfig: [String: Any]?) {
+            let wasEnabled = isEnabled
             updatePlugins(from: remoteConfig)
             updateEventTriggers(from: remoteConfig)
             updateCachedMinimumDuration()
+            if wasEnabled {
+                reevaluateSampling()
+            }
         }
 
         private func updateCachedMinimumDuration() {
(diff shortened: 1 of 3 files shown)


PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -433,9 +433,13 @@
         }
 
         func applyRemoteConfig(remoteConfig: [String: Any]?) {
+            let wasEnabled = isEnabled
             updatePlugins(from: remoteConfig)
             updateEventTriggers(from: remoteConfig)
             updateCachedMinimumDuration()
+            if wasEnabled {
```
ioannisj: I think we can reevaluateSampling() unconditionally here to catch the other end of the stick as well (did not start but is now sampled and should start)

ioannisj (approved):

PostHog/Replay/PostHogReplayIntegration.swift
marandaneto: Good catch! Fixed in 8431a2bec — remote config changes now reevaluate session replay sampling unconditionally, so replay can start when it becomes sampled/enabled as well as stop when it no longer should run.

PR #666, opened 2026-06-17
feat(errortracking): add `ignoredExceptionTypes` config to skip RN-duplicate native crashes (closes #653)

### LINKED ISSUE #653: Native crash autocapture double-reports fatal React Native JS errors (RCTFatalException)
### Problem

When the React Native SDK enables native crash autocapture (`@posthog/react-native-plugin`, `errorTracking.autocapture.nativeCrashes`, see PostHog/posthog-js#3677), a **fatal JS error can be captured twice**:

1. The JS layer captures it via `uncaughtExceptions` autocapture (with the JS stack trace), and
2. React Native rethrows fatal JS errors on the native side through `RCTFatal(...)`, which raises an uncaught `NSException` named `RCTFatalException`. posthog-ios' crash reporter then captures that as a separate native crash (with a native stack trace).

The result is two `$exception` events for one logical error.

### Suggested fix

posthog-ios should expose a way to ignore native exceptions that originate from React Native's fatal-JS path, and/or skip reporting uncaught `NSException`s whose name is `RCTFatalException`, so the JS-captured event is the single source of truth.

For reference, sentry-react-native does the equivalent on the native side:
> // React native internally throws a JavascriptException. we want to ignore it on the native side to avoid sending it twice.

(`addIgnoredExceptionForType(...)` — they ignore the RN-internal exception type so it isn't sent twice.)

### Notes

- This pairs with the Android equivalent (sibling issue in posthog-android: PostHog/posthog-android#567).
- Worth verifying empirically against current RN (0.79+) release builds, since the exact rethrow path can change between RN versions.
- Discovered while reviewing PostHog/posthog-js#3677.

tsushanth:
## Why

Closes #653. Pairs with the Android side at PostHog/posthog-android#567 (fixed by PostHog/posthog-android#569).

React Native rethrows fatal JS errors via \`RCTFatal(...)\` as an uncaught \`NSException\` named \`RCTFatalException\`. When the \`@posthog/react-native-plugin\` enables native crash autocapture (\`errorTracking.autocapture.nativeCrashes\`), the iOS crash reporter captures that as a separate \`\$exception\` event with a native stack trace — duplicating the event the JS layer already captured with its own (richer) JS stack trace. Result: two \`\$exception\` events for one logical error, the duplicate carrying less useful context.

sentry-react-native solves the equivalent on its iOS side with \`addIgnoredExceptionForType(...)\`. posthog-android shipped a parallel \`errorTrackingConfig.ignoredExceptionTypes\` filter via PostHog/posthog-android#569. This PR brings the same lever to posthog-ios so callers can keep native crash autocapture on but dedup the RCTFatalException churn.

### New config field

\`\`\`swift
// PostHogErrorTrackingConfig.swift
@objc public var ignoredExceptionTypes: [String] = []
\`\`\`

Caller-facing usage from the React Native plugin (or any consumer hitting this duplicate):

\`\`\`swift
config.errorTrackingConfig.ignoredExceptionTypes = [\"RCTFatalException\"]
\`\`\`

### Filter site

In \`PostHogErrorTrackingAutoCaptureIntegration.processPendingCrashReport()\`, after \`PostHogCrashReportProcessor.processReport(...)\` produces the merged properties and before \`captureInternal(\"\$exception\", ...)\` runs:

\`\`\`swift
let ignored = postHog.config.errorTrackingConfig.ignoredExceptionTypes
if !ignored.isEmpty, PostHogErrorTrackingAutoCaptureIntegration.exceptionListMatchesIgnoredTypes(finalProperties, ignoredTypes: ignored) {
    hedgeLog(\"Crash report skipped: exception type is in errorTrackingConfig.ignoredExceptionTypes\")
    return
}
\`\`\`

The helper walks the full \`\$exception_list\` rather than only the outermost entry — a wrapped \`RCTFatalException\` underneath an outer \`NSException\` wrapper is still suppressed. Match is exact and case-sensitive (NSException class names are stable identifiers, not free text).

## Tests

New \`PostHogTests/PostHogErrorTrackingIgnoredTypesTest.swift\` using the Swift Testing framework. Seven cases:

1. **Empty list is a no-op** — \`ignoredExceptionTypes: []\` passes all exceptions through (the default behavior, fully backwards-compatible).
2. **Outer-type match suppresses** — the typical \`[\"RCTFatalException\"]\` case.
3. **Inner-chain match suppresses** — a wrapped RCTFatalException underneath an outer NSException still triggers the filter.
4. **Mismatched type passes through** — \`NSGenericException\` with a \`RCTFatalException\` ignore list isn't filtered.
5. **Missing \`\$exception_list\` key is safe** — properties without exceptions return \`false\` cleanly.
6. **Empty \`\$exception_list\` is safe** — same.
7. **Match is case-sensitive** — \`\"rctfatalexception\"\` does NOT match \`\"RCTFatalException\"\` (the field is a class name, not user input).

Plus a sanity check that the config field defaults to \`[]\`.

## Backwards compatibility

Strict additive change:

- \`ignoredExceptionTypes\` is a new property with a default of \`[]\`, so existing apps with no config changes see identical behavior.
- The filter only runs when the list is non-empty (\`if !ignored.isEmpty\`), so the happy path adds zero cost for callers who don't opt in.
- No existing public API changes; no behavior change on Mach exceptions, POSIX signals, or NSExceptions whose type isn't in the ignore list.

## Local verification

- \`swift build\` clean.
- I couldn't run \`swift test\` to completion locally because the main branch's \`PostHogSessionManagerTest\` has pre-existing unrelated compile errors (\`value of type 'PostHogSDK' has no member 'getSessionManager'\`). CI should handle the test target correctly. The new tests are pure data-flow against a static helper, so I'm confident in the logic.

### DIFF
--- a/PostHog/ErrorTracking/PostHogErrorTrackingAutoCaptureIntegration.swift
+++ b/PostHog/ErrorTracking/PostHogErrorTrackingAutoCaptureIntegration.swift
@@ -204,10 +204,8 @@ import Foundation
             }
             let ignored = Set(ignoredTypes)
             return exceptionList.contains { entry in
-                if let type = entry["type"] as? String, ignored.contains(type) {
-                    return true
-                }
-                return false
+                guard let exType = entry["type"] as? String else { return false }
+                return ignored.contains(exType)
             }
         }
     }
--- a/PostHog/ErrorTracking/PostHogErrorTrackingAutoCaptureIntegration.swift
+++ b/PostHog/ErrorTracking/PostHogErrorTrackingAutoCaptureIntegration.swift
@@ -211,7 +211,7 @@ import Foundation
     }
 
 #else
-    // watchOS stub - crash reporting is not available
+    // watchOS/visionOS stub - crash reporting is not available on these platforms
     class PostHogErrorTrackingAutoCaptureIntegration: PostHogIntegration {
         var requiresSwizzling: Bool { false }
 
@@ -222,5 +222,10 @@ import Foundation
         func uninstall(_: PostHogSDK) { /* no-op */ }
         func start() { /* no-op */ }
         func stop() { /* no-op */ }
+
+        /// Crash reporting is unavailable on this platform; always returns `false`.
+        static func exceptionListMatchesIgnoredTypes(_: [String: Any], ignoredTypes _: [String]) -> Bool {
+            false
+        }
     }
 #endif
--- a/PostHog/ErrorTracking/PostHogErrorTrackingAutoCaptureIntegration.swift
+++ b/PostHog/ErrorTracking/PostHogErrorTrackingAutoCaptureIntegration.swift
@@ -164,6 +164,17 @@ import Foundation
                     finalProperties[PostHogExceptionStepFields.stepsKey] = crashSteps
                 }
 
+                // Honor `errorTrackingConfig.ignoredExceptionTypes`. The
+                // primary use case is React Native's `RCTFatalException`,
+                // which is rethrown for every fatal JS error and would
+                // otherwise duplicate the JS-side `$exception` event with
+                // a redundant native stack trace (see #653).
+                let ignored = postHog.config.errorTrackingConfig.ignoredExceptionTypes
+                if !ignored.isEmpty, PostHogErrorTrackingAutoCaptureIntegration.exceptionListMatchesIgnoredTypes(finalProperties, ignoredTypes: ignored) {
+                    hedgeLog("Crash report skipped: exception type is in errorTrackingConfig.ignoredExceptionTypes")
+                    return
+                }
+
                 // Collect crash timestamp
                 let crashTimestamp = PostHogCrashReportProcessor.getCrashTimestamp(crashReport)
 
@@ -179,6 +190,24 @@ import Foundation
                 hedgeLog("Crash report processed and captured")
             } catch {
                 hedgeLog("Failed to process crash report: \(error)")
+            }
+        }
+
+        /// Returns `true` if any entry in `properties["$exception_list"]` has a
+        /// `type` matching one of `ignoredTypes`. Walks the exception list rather
+        /// than only the outermost entry so a wrapped exception whose underlying
+        /// cause is, e.g., `RCTFatalException` is still suppressed. Match is
+        /// case-sensitive and exact (the field is a class name, not free text).
+        static func exceptionListMatchesIgnoredTypes(_ properties: [String: Any], ignoredTypes: [String]) -> Bool {
+            guard let exceptionList = properties["$exception_list"] as? [[String: Any]] else {
+                return false
+            }
+            let ignored = Set(ignoredTypes)
+            return exceptionList.contains { entry in
+                if let type = entry["type"] as? String, ignored.contains(type) {
+                    return true
+                }
+                return false
             }
         }
     }
--- a/PostHog/ErrorTracking/PostHogErrorTrackingConfig.swift
+++ b/PostHog/ErrorTracking/PostHogErrorTrackingConfig.swift
@@ -107,30 +107,36 @@ import Foundation
 
     /// List of exception class names (the `NSException.name.rawValue` /
     /// `$exception_list[*].type` field) that should be skipped during
-    /// autocapture instead of being sent as `$exception` events.
+    /// autocapture and manual `captureException(_:)` instead of being sent
+    /// as `$exception` events.
     ///
     /// Exists primarily to dedup fatal exceptions when the React Native
     /// plugin's native-crash autocapture is enabled. React Native rethrows
     /// fatal JS errors via `RCTFatal(...)` as an `NSException` named
     /// `RCTFatalException`, which the iOS crash reporter then captures
     /// as a separate native crash — duplicating the event the JS layer
-    /// already captured with its own stack trace. Add `RCTFatalException`
-    /// here to make the JS-side `$exception` event the single source of
-    /// truth. Mirrors `addIgnoredExceptionForType(...)` on
+    /// already captured with its own stack trace. The default `["RCTFatalException"]`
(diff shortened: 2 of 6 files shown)


ioannisj:
Thank you @tsushanth, will have a look shortly. In the meantime can you please add changeset entry (see https://github.com/PostHog/posthog-ios/blob/92138fd8b48aa34a08ce54bdd366e4495d29d405/RELEASING.md)  and run `make apiUpdate` since we are expanding the api surface?

tsushanth:
Thanks @ioannisj — added [`.changeset/feat-ignored-exception-types.md`](https://github.com/PostHog/posthog-ios/pull/666/files#diff-feat-ignored-exception-types) as `minor` since this surfaces a new public knob on `PostHogErrorTrackingConfig`.

While here I also took greptile's second suggestion ([9f961d5](https://github.com/PostHog/posthog-ios/pull/666/commits/9f961d54c)) — collapsed the `exceptionListMatchesIgnoredTypes` closure to a `guard` + direct return and renamed the local binding off `type` to avoid shadowing `type(of:)`. Left the test-parameterisation suggestion alone since the seven cases read more clearly as separate `@Test` methods, but happy to fold them with `@Test(arguments:)` if you'd prefer.

Ready for another look whenever you have a moment.

tsushanth:
Thanks for the patience @ioannisj. Rebased onto current `main`, resolved the conflict against the new `exceptionSteps` block in `PostHogErrorTrackingConfig.swift` (the two sit side-by-side now), and added the matching `api/posthog-ios.public-api.txt` entry. Ready for another look whenever you have a moment.

ioannisj:
@tsushanth taking a look now, apologies for the delay

ioannisj:
@tsushanth looks like it needs a fresh rebase since I can see reverting some changes on main (workflow, PR template etc)

PostHogTests/PostHogErrorTrackingIgnoredTypesTest.swift
```diff
@@ -0,0 +1,132 @@
...
+                        ignoredTypes: ["rctfatalexception"]
+                    ) == false
+            )
+        }
+
+        @Test("config field default is empty so callers see no behavior change")
+        func defaultIsEmpty() {
+            let config = PostHogErrorTrackingConfig()
+            #expect(config.ignoredExceptionTypes.isEmpty)
+        }
```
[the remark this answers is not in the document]
ioannisj: Yes let's use parameterized tests where possible

PostHog/ErrorTracking/PostHogErrorTrackingConfig.swift
```diff
@@ -103,6 +103,35 @@ import Foundation
...
+    /// ```swift
+    /// config.errorTrackingConfig.ignoredExceptionTypes = [
+    ///     "RCTFatalException",
+    /// ]
+    /// ```
+    ///
+    /// See https://github.com/PostHog/posthog-ios/issues/653.
+    ///
+    /// Default: empty (no filtering).
+    @objc public var ignoredExceptionTypes: [String] = []
```
ioannisj: Would it make sense providing a sensible default ["RCTFatalException"] here?

PostHog/ErrorTracking/PostHogErrorTrackingAutoCaptureIntegration.swift
```diff
@@ -164,6 +164,17 @@ import Foundation
                     finalProperties[PostHogExceptionStepFields.stepsKey] = crashSteps
                 }
 
+                // Honor `errorTrackingConfig.ignoredExceptionTypes`. The
```
ioannisj, line 167: Similar to Android, we should also check the ignore list through the manual capture api

tsushanth:
Thanks @ioannisj — pushed [d64a9123c](https://github.com/PostHog/posthog-ios/pull/666/commits/d64a9123c) dropping the workaround commit that reverted the workflow / PR-template updates. The diff is now scoped to the five real files (errortracking config + integration, the new test, the changeset, and the `api/posthog-ios.public-api.txt` snapshot entry). Ready for another pass whenever you have a moment.

tsushanth:
Thanks @ioannisj — sorry for the delay, addressed all three in [8b7d0a4](https://github.com/PostHog/posthog-ios/pull/666/commits/8b7d0a4):

**1. Parameterized tests.** Collapsed the seven hand-rolled cases into a single `@Test(arguments:)` over a `MatchCase` table — adding a new shape is now one row instead of another `@Test` block. Swift Testing reports all seven sub-cases individually so failures still point at the right shape.

**2. Default to `["RCTFatalException"]`.** Good call — the only audience this matters to is React Native apps, and native-only iOS apps never raise that exception type so there's no behavioral risk. Docstring shows how to extend or clear the default. Updated the `defaultIsEmpty` test accordingly (now `defaultIsRCTFatalException`).

**3. Manual capture path.** Threaded the same `ignoredExceptionTypes` check through `captureExceptionEvent` so both `captureException(_:Error)` and `captureException(_:NSException)` honor it — matching what posthog-android's counterpart does. Crash-report autocapture path keeps its own check (the two paths surface different exception sources).

`make test` clean — 594 tests pass.

ioannisj:
@tsushanth look like builds are failing on non-ios platforms. Mind making sure the CI is green before approving?

tsushanth:
Fixed the watchOS/visionOS build failures: `exceptionListMatchesIgnoredTypes` was defined only inside the `#if os(iOS) || os(macOS) || os(tvOS)` block, so the watchOS/visionOS stub class was missing it. `PostHogSDK.captureExceptionEvent` calls it unconditionally, causing the compile error on those platforms. Added a no-op stub (returns `false`) to the `#else` branch — crash reporting is unavailable on watchOS/visionOS anyway so the filter is a no-op.

ioannisj (approved):

PR #674, opened 2026-06-25
fix(flags): retry flag requests on transient network errors

marandaneto:

`/flags/?v=2` evaluation should tolerate transient network failures, but should not retry HTTP/API responses such as 408, 429, or 5xx.

This PR adds bounded retry/backoff only for transient network failures (timeouts, connection resets/lost connections, and EOF-style failures) and keeps HTTP/API status errors terminal.

See the repo-specific command below.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Implemented by pi after a maintainer-directed cross-SDK pass. The chosen policy is transient network-only retry/backoff for `/flags/?v=2`; HTTP/API statuses remain non-retryable by design. Connection-refused failures also fail fast where the platform exposes that distinction.

Tested with:

```bash
make test filter=PostHogApiTests
```

Note: build output included existing Swift warnings in unrelated queue tests; `PostHogApiTests` passed.

### DIFF
--- a/PostHog/PostHogApi.swift
+++ b/PostHog/PostHogApi.swift
@@ -46,11 +46,11 @@ class PostHogApi {
     /// Shared so connection pool, TLS state, and HTTP/2 streams survive
     /// between calls instead of being torn down per request.
     private let session: URLSession
-    private let flagsRetryDelay: TimeInterval
 
-    init(_ config: PostHogConfig, flagsRetryDelay: TimeInterval = retryDelay) {
+    static let flagsRetryDelay: TimeInterval = 0.3
+
+    init(_ config: PostHogConfig) {
         self.config = config
-        self.flagsRetryDelay = flagsRetryDelay
 
         // Copy first so SDK mutations don't leak back to the caller's object.
         let sessionConfig = (config.urlSessionConfiguration?.copy() as? URLSessionConfiguration)
@@ -288,7 +288,7 @@ class PostHogApi {
             if let error {
                 if Self.isRetryableFlagsError(error), retryCount < self.config.featureFlagRequestMaxRetries {
                     let nextRetryCount = retryCount + 1
-                    let delay = min(TimeInterval(nextRetryCount) * self.flagsRetryDelay, maxRetryDelay)
+                    let delay = Self.featureFlagsRetryDelay(forFailedAttempt: nextRetryCount)
                     hedgeLog(
                         "Error calling the flags API: \(error). Retrying in \(delay) seconds (attempt \(nextRetryCount)/\(self.config.featureFlagRequestMaxRetries))."
                     )
@@ -328,6 +328,10 @@ class PostHogApi {
                 completion(nil, error)
             }
         }.resume()
+    }
+
+    static func featureFlagsRetryDelay(forFailedAttempt failedAttempt: Int) -> TimeInterval {
+        min(flagsRetryDelay * pow(2.0, TimeInterval(failedAttempt - 1)), maxRetryDelay)
     }
 
     private static func isRetryableFlagsError(_ error: Error) -> Bool {
--- a/PostHog/PostHogApi.swift
+++ b/PostHog/PostHogApi.swift
@@ -286,8 +286,7 @@ class PostHogApi {
     ) {
         session.uploadTask(with: request, from: payload) { data, response, error in
             if let error {
-                let nsError = error as NSError
-                if nsError.domain == NSURLErrorDomain, retryCount < self.config.maxRetries {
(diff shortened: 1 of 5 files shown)


PostHogTests/PostHogApiTest.swift
```diff
@@ -196,6 +220,56 @@ enum PostHogApiTests {
...
+
+        @Test("does not retry HTTP 429 responses")
+        func doesNotRetryHTTP429() async throws {
+            try await testFlagsDoesNotRetryHTTPStatus(429)
+        }
+
+        @Test("does not retry HTTP 500 responses")
+        func doesNotRetryHTTP500() async throws {
+            try await testFlagsDoesNotRetryHTTPStatus(500)
+        }
```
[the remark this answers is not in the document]
marandaneto: Addressed in 62acc20dd: added a retry-exhaustion test that sets featureFlagRequestMaxRetries = 2, forces transient URLSession errors on every attempt, and asserts the final error plus initial + 2 retry requests.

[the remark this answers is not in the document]
marandaneto: Addressed in 62acc20dd: replaced the three duplicate HTTP status wrappers with a single Swift Testing parameterized test over [408, 429, 500].

PostHog/PostHogApi.swift
```diff
@@ -273,9 +275,31 @@ class PostHogApi {
...
+                    let nextRetryCount = retryCount + 1
+                    let delay = min(TimeInterval(nextRetryCount) * self.flagsRetryDelay, maxRetryDelay)
+                    hedgeLog(
+                        "Error calling the flags API: \(error). Retrying in \(delay) seconds (attempt \(nextRetryCount)/\(self.config.maxRetries))."
+                    )
+                    DispatchQueue.global(qos: .utility).asyncAfter(deadline: .now() + delay) {
+                        self.uploadFlagsRequest(request, payload: payload, retryCount: nextRetryCount, completion: completion)
+                    }
+                    return
+                }
```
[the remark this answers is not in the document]
marandaneto: Addressed in 62acc20dd/current diff: cancelled requests are terminal because only timed-out and network-connection-lost NSURLError codes are retryable, and the non-transient URLSession error test now explicitly covers NSURLErrorCancelled.

PostHog/PostHogApi.swift
```diff
@@ -46,9 +46,11 @@ class PostHogApi {
     /// Shared so connection pool, TLS state, and HTTP/2 streams survive
     /// between calls instead of being torn down per request.
     private let session: URLSession
+    private let flagsRetryDelay: TimeInterval
 
-    init(_ config: PostHogConfig) {
+    init(_ config: PostHogConfig, flagsRetryDelay: TimeInterval = retryDelay) {
```
marandaneto: keep `flagsRetryDelay` internal instead of a ctor param for `PostHog/PostHogApi.swift`

PostHog/PostHogApi.swift
```diff
@@ -273,9 +275,30 @@ class PostHogApi {
...
+            if let error {
+                if Self.isRetryableFlagsError(error), retryCount < self.config.featureFlagRequestMaxRetries {
+                    let nextRetryCount = retryCount + 1
+                    let delay = min(TimeInterval(nextRetryCount) * self.flagsRetryDelay, maxRetryDelay)
+                    hedgeLog(
+                        "Error calling the flags API: \(error). Retrying in \(delay) seconds (attempt \(nextRetryCount)/\(self.config.featureFlagRequestMaxRetries))."
+                    )
+                    DispatchQueue.global(qos: .utility).asyncAfter(deadline: .now() + delay) {
+                        self.uploadFlagsRequest(request, payload: payload, retryCount: nextRetryCount, completion: completion)
+                    }
```
marandaneto, line 297: should we just sleep here instead? or is it better with `DispatchQueue.global`?

PostHog/PostHogApi.swift
marandaneto: Fixed in 594af8c24: removed the `flagsRetryDelay` constructor parameter and kept the feature-flag retry delay as an internal `PostHogApi` value/helper, with coverage for 300ms/600ms/1.2s.

PostHog/PostHogApi.swift
marandaneto: Keeping `DispatchQueue.global(qos: .utility).asyncAfter` here rather than `sleep`: it schedules the retry without blocking the URLSession completion callback thread. A blocking sleep would tie up the callback worker for the whole backoff window.

turnipdabeets (approved):

turnipdabeets (approved):

PR #683, opened 2026-06-29
feat: add requestHeaders config option

turnipdabeets:

Part of the cross-SDK work for PostHog/posthog-js#2132 (no way to include `Authorization` headers in reverse-proxy setups). React Native session replay and native crash uploads are sent directly by this native SDK, so they bypass any JS-layer header support. This adds first-class custom-header support to the iOS SDK so those (and all other) requests can carry an `Authorization` header for an authenticated reverse proxy.

- Added a `Custom request headers` suite in `PostHogApiTest` asserting the header reaches `/batch` and `/flags`, and is absent when unset (uses the existing `MockPostHogServer` / OHHTTPStubs harness, same pattern as the Content-Encoding tests).
- Verified the SDK compiles (`xcrun swift build --arch arm64`). The simulator test run couldn't execute from my local git worktree due to the project's local SPM package path resolving outside the worktree; CI runs the suite on-simulator.
- `swiftformat --lint` and `swiftlint` clean on the changed files.

## :pencil: Checklist

## If releasing new changes

## 🤖 Agent context

Directed by @ioannisj / @marandaneto while wiring up `requestHeaders` for React Native — the native SDK needs to honor the header for replay/crash uploads. Chose to apply the headers per-request in `PostHogApi` (via a small `applyCustomHeaders` helper on each `URLRequest`) rather than via `URLSessionConfiguration.httpAdditionalHeaders`, because session-level headers are not observable through OHHTTPStubs and the per-request approach is both testable and keeps SDK-set headers (gzip `Content-Encoding`) applied last. Tool used: Claude Code.

---

Related: PostHog/posthog-js#2132

### DIFF
--- a/.changeset/request-headers.md
+++ b/.changeset/request-headers.md
@@ -1,5 +1,5 @@
 ---
-'posthog-ios': patch
+'posthog-ios': minor
 ---
 
 Add `requestHeaders` config option to send custom headers (e.g. `Authorization`) with every request to the PostHog API. Useful for reverse-proxy setups that require authentication.
--- a/.changeset/request-headers.md
+++ b/.changeset/request-headers.md
@@ -1,0 +1,5 @@
+---
+'posthog-ios': patch
+---
+
+Add `requestHeaders` config option to send custom headers (e.g. `Authorization`) with every request to the PostHog API. Useful for reverse-proxy setups that require authentication.
--- a/PostHog/PostHogApi.swift
+++ b/PostHog/PostHogApi.swift
@@ -42,8 +42,7 @@ class PostHogApi {
 
     private let config: PostHogConfig
 
-    /// Snapshot taken at init so the per-request and redirect paths stay consistent even if the
-    /// caller mutates `config.requestHeaders` afterwards.
+    /// Snapshot at init; treated as immutable after setup.
     private let customRequestHeaders: [String: String]
 
     // default is 60s but we do 10s
@@ -94,8 +93,7 @@ class PostHogApi {
         return request
     }
 
-    /// Applies the custom headers to requests for the configured host. SDK-set headers take
-    /// precedence, and requests to rewritten hosts (e.g. the static-config CDN) are skipped.
+    /// Adds custom headers to requests for the configured host, skipping SDK-managed keys.
     private func applyCustomHeaders(_ request: inout URLRequest) {
         guard !customRequestHeaders.isEmpty, request.url?.host == config.host.host else { return }
         for (key, value) in customRequestHeaders
@@ -106,9 +104,7 @@ class PostHogApi {
         }
     }
 
-    /// Headers the SDK manages itself; custom values for these are ignored. Compared
-    /// case-insensitively since HTTP header names are case-insensitive. `Content-Type`,
-    /// `User-Agent`, and `Accept-Encoding` are set on the session; `Content-Encoding` per request.
+    /// SDK-managed headers that custom values can't override (compared lowercase).
     private static let reservedHeaderKeys: Set<String> = [
         "content-type", "user-agent", "accept-encoding", "content-encoding",
     ]
--- a/PostHog/PostHogApi.swift
+++ b/PostHog/PostHogApi.swift
@@ -98,10 +98,20 @@ class PostHogApi {
     /// precedence, and requests to rewritten hosts (e.g. the static-config CDN) are skipped.
     private func applyCustomHeaders(_ request: inout URLRequest) {
         guard !customRequestHeaders.isEmpty, request.url?.host == config.host.host else { return }
-        for (key, value) in customRequestHeaders where request.value(forHTTPHeaderField: key) == nil {
+        for (key, value) in customRequestHeaders
+            where request.value(forHTTPHeaderField: key) == nil
+            && !Self.reservedHeaderKeys.contains(key.lowercased())
+        {
             request.setValue(value, forHTTPHeaderField: key)
         }
     }
+
+    /// Headers the SDK manages itself; custom values for these are ignored. Compared
+    /// case-insensitively since HTTP header names are case-insensitive. `Content-Type`,
+    /// `User-Agent`, and `Accept-Encoding` are set on the session; `Content-Encoding` per request.
+    private static let reservedHeaderKeys: Set<String> = [
+        "content-type", "user-agent", "accept-encoding", "content-encoding",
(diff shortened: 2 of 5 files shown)


PostHogTests/PostHogApiTest.swift
```diff
@@ -233,6 +233,49 @@ enum PostHogApiTests {
...
+                sut.batch(events: [], completion: completion)
+            }
+            let request = try #require(server.batchRequests.first)
+            #expect(request.value(forHTTPHeaderField: "Authorization") == nil)
+        }
+    }
+
     @Suite("Test flags endpoint with different host paths")
     class TestFlagsEndpoint: BaseTestSuite {
         @Test("feature flag retry delay starts at 300ms and doubles")
```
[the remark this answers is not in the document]
turnipdabeets: Kept these as separate explicitly-named tests: `attachesToBatch` and `attachesToFlags` exercise two genuinely different request builders (the POST `getURLRequest` path vs. the GET builder), and the third covers the unset case. For just these I found the explicit form more readable than an `arguments:`-driven enum switch, but happy to parameterize if you'd prefer to standardize it.

turnipdabeets:
Unblocks the RN plugin bridge: PostHog/posthog-js#4018 sets `config.requestHeaders` and can only merge once this lands + releases.

PostHog/PostHogApi.swift
```diff
@@ -76,12 +76,22 @@ class PostHogApi {
...
         }
         return request
     }
 
+    /// Applies the caller-supplied `config.requestHeaders` (e.g. an `Authorization` header for a
+    /// reverse proxy) to every request the SDK sends.
+    private func applyCustomHeaders(_ request: inout URLRequest) {
+        guard let requestHeaders = config.requestHeaders else { return }
+        for (key, value) in requestHeaders {
+            request.setValue(value, forHTTPHeaderField: key)
```
marandaneto, line 104: this will overwrite current values
see https://github.com/PostHog/posthog-android/pull/592#discussion_r3496943754

PostHog/PostHogConfig.swift
```diff
@@ -354,6 +354,10 @@ public typealias BeforeSendBlock = (PostHogEvent) -> PostHogEvent?
     /// Useful for testing, proxying, or custom network configurations
     @objc public var urlSessionConfiguration: URLSessionConfiguration?
 
+    /// Custom headers to send with every request to the PostHog API.
+    /// Useful for reverse-proxy setups that require authentication, e.g. an `Authorization` header.
+    @objc public var requestHeaders: [String: String]?
```
marandaneto, line 360: we have setters for setBeforeSend, we do the same here?

PostHog/PostHogApi.swift
```diff
@@ -76,12 +76,22 @@ class PostHogApi {
...
         if gzipped {
             request.setValue("gzip", forHTTPHeaderField: "Content-Encoding")
         }
         return request
     }
 
+    /// Applies the caller-supplied `config.requestHeaders` (e.g. an `Authorization` header for a
+    /// reverse proxy) to every request the SDK sends.
+    private func applyCustomHeaders(_ request: inout URLRequest) {
+        guard let requestHeaders = config.requestHeaders else { return }
```
ioannisj: we should probably work off a local copy here. Config should be immutable even though modeled as a class in this repo for ObjC interop

PostHog/PostHogConfig.swift
```diff
@@ -354,6 +354,10 @@ public typealias BeforeSendBlock = (PostHogEvent) -> PostHogEvent?
     /// Useful for testing, proxying, or custom network configurations
     @objc public var urlSessionConfiguration: URLSessionConfiguration?
 
+    /// Custom headers to send with every request to the PostHog API.
+    /// Useful for reverse-proxy setups that require authentication, e.g. an `Authorization` header.
```
ioannisj, line 358: On Authorization headers, any redirects will not strip the header, so on cross-origin calls this could leak? We currently have a config option for a custom `URLSessionConfiguration` but not a custom `URLSession` (where people could define their delegate so they could potentially do the stripping?) 

Not sure if we should ship without or leave a note? wdyt?

.changeset/request-headers.md
```diff
@@ -0,0 +1,5 @@
+---
+'posthog-ios': patch
```
ioannisj: would probably release this as minor since it adds a feature?

PostHog/PostHogConfig.swift
ioannisj: I think setBeforeSend is the only explicit set* and was added because of swift closures + objc interop. I think this config bridges okay so I'm okay to leave as-is?

PostHog/PostHogApi.swift
ioannisj: +1

turnipdabeets: Fixed in 69a84d73f

PostHog/PostHogConfig.swift
turnipdabeets: Fixed — a URLSessionTaskDelegate strips custom headers on redirects leaving the configured host.

PostHog/PostHogApi.swift
turnipdabeets: should be immutable now, also carried this idea to android

PostHog/PostHogApi.swift
```diff
@@ -76,12 +86,29 @@ class PostHogApi {
...
+            && !Self.reservedHeaderKeys.contains(key.lowercased())
+        {
+            request.setValue(value, forHTTPHeaderField: key)
+        }
+    }
+
+    /// SDK-managed headers that custom values can't override (compared lowercase).
+    private static let reservedHeaderKeys: Set<String> = [
+        "content-type", "user-agent", "accept-encoding", "content-encoding",
+    ]
```
marandaneto, line 111: do we do this for android? if not we should

marandaneto (approved):
left a few comments but unblocking, CI is unhappy

PostHog/PostHogApi.swift
turnipdabeets: it does, OkHttp's BridgeInterceptor has already put Content-Type/User-Agent/Accept-Encoding on the request by the time it runs. So the existing request.header(key) ?: … guard ("skip if already present") sees them and skips


added a test to android for it.

PR #686, opened 2026-06-30
feat(surveys): re-translate displayed survey on language change

bs1180:

Once a translated survey was displayed, later changes to the user's language (e.g. via `setPersonProperties` / `identify` with a `language` property) had no effect — the survey kept the language it was first rendered with. The translation was resolved only at display time.

This makes a survey already on screen follow the user's language: when the person properties used for flags change and a matching translation exists, the survey re-translates in place, keeping the current question and any progress.

Closes #66998

Added unit tests (run on the iOS simulator, where survey code compiles):

- Changing the `language` person property while a survey is active pushes a re-translated survey to the delegate and updates the tracked survey language.
- Re-resolving to the same language is a no-op (no churn of `$survey_language` or re-renders).
- No active survey ⇒ no update pushed.
- `SurveyDisplayController.updateSurvey` preserves the current question index and completion state, and ignores updates for a different / absent survey.

I was unable to run `make test` / the iOS build in this environment (no Xcode available), so CI should be relied on for the full build + simulator test run.

## :pencil: Checklist

## If releasing new changes

## 🤖 Agent context

Implemented from a Slack thread request, logged as #66998. Built with the PostHog Slack app (Claude Code). Key decisions:

- **Trigger:** chose an internal `NotificationCenter` signal posted from `PostHogRemoteConfig.setPersonPropertiesForFlags` (the single chokepoint all person-property updates flow through) rather than sniffing the captured event stream — the `$set`/`$identify` ordering relative to the flag-property update is inconsistent, so the notification is the order-correct, robust hook.
- **UI update:** added an optional `updateSurvey(_:)` to the delegate and made `SurveySheet` observe the display controller (instead of the snapshot passed by `.sheet(item:)`), so an in-place content swap with the same survey id renders live while preserving SwiftUI `@State` (current question, typed answers).
- **Guardrails:** only commit + render when the matched language actually changes; the whole path is a no-op when no survey is displayed.

One known limitation worth a reviewer's eye: for choice questions, selection is keyed by the option's string value, so a mid-survey language switch can drop an in-progress selection on those question types (labels change). Text/rating answers are preserved.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C07QD3LT8U9/p1781615014522869?thread_ts=1781615014.522869&cid=C07QD3LT8U9)*


turnipdabeets:
@lucasheriques I know you worked a bit on this, mind giving this a review?

turnipdabeets:
Thanks @bs1180 instead of a CHANGELOG we actually need a changes, please see `RELEASING.md`

ioannisj:
@bs1180 friendly ping on this one

bs1180:
Hey @turnipdabeets and @ioannisj, sorry for the delay. This PR is the product of letting PostHog AI loose on a customer complaint :D I'm not an IOS dev so not sure if this implementation makes sense and I've done literally zero QA. Is that OK with you? Won't be offended this just gets rejected.

ioannisj:
@bs1180 thank you, I can do the QA part and see also how this will affect Flutter surveys as well. we'll probably need to land a similar feature for posthog-android as well I imagine? To get some more context though, do you have the original link to the customer complaint? I personally don't think that on-the-fly language changes are the norm so I'm not completely sure we should be supporting this

ioannisj:
Would like to know @turnipdabeets thoughts as well on this, maybe it's just me :)

bs1180:
Yup totally understood - https://github.com/PostHog/posthog/issues/66998 is the original feature request, and this is the [customer Slack thread](https://posthog.slack.com/archives/C09J17L1K09/p1781607912021619)

turnipdabeets:
> Would like to know @turnipdabeets thoughts as well on this, maybe it's just me :)

I think we need on-device QA, if you want to take that @ioannisj. Maybe we need to check out the sheet dismissal animation too. Not sure if we need @PostHog/team-surveys on par before making this change? And it makes sense to do this for the other mobile if it's needed, android flutter and RN.

PostHog/PostHogRemoteConfig.swift
```diff
@@ -648,6 +648,10 @@ class PostHogRemoteConfig {
             // Persist to disk
             storage.setDictionary(forKey: .personPropertiesForFlags, contents: personPropertiesForFlags)
         }
+        // Notify observers (e.g. surveys) so a survey already on screen can re-resolve its
+        // language if the user's `language` property changed. Posted outside the lock so
+        // observers don't run while we hold it.
+        NotificationCenter.default.post(name: PostHogSDK.personPropertiesForFlagsDidChange, object: nil)
```
ioannisj: nit: I would use the existing existing [PostHogMulticastCallback](https://github.com/PostHog/posthog-ios/blob/76a3e1db35f3153f3388bdf077a717c7350e5f9e/PostHog/Utils/PostHogMulticastCallback.swift#L27) pattern here, but this also works

PostHog/Surveys/PostHogSurveyIntegration.swift
```diff
@@ -318,6 +335,45 @@
...
+
+            let language = resolveDisplayLanguage()
+            let translations = resolveSurveyTranslations(survey: activeSurvey, targetLanguage: language)
+
+            // Commit the new translation only if it targets a different language than what's
+            // currently shown (and the active survey hasn't changed underneath us).
+            let didChange: Bool = activeSurveyLock.withLock {
+                guard self.activeSurvey?.id == activeSurvey.id else { return false }
+                guard self.activeSurveyLanguage != translations.matchedKey else { return false }
+                self.activeSurveyLanguage = translations.matchedKey
```
ioannisj, line 359: updateSurvey is a optional, so any custom delegate that doesn't implement it (e.g Flutter) no-ops the visual update while internal state advances, then sendSurveySentEvent/sendSurveyDismissedEvent stamp $survey_language and translated $survey_questions for a language the user never saw. 
This bites custom delegates and Flutter currently. Consider gating the state change on the delegate actually responding to the selector?

PostHog/Surveys/Utils/PostHogSurveyMatching.swift
```diff
@@ -0,0 +1,135 @@
+//
+//  PostHogSurveyMatching.swift
```
ioannisj, line 2: Should we move this inside Surveys/Utils?

ioannisj (approved):
LG. Tested this on iOS and Flutter and left some minor comments. We'll probably want to follow this up on Android and Flutter side

ioannisj:
Hey @bs1180, want me to take this over if you've moved on to something else?

bs1180:
Hey @ioannisj, sorry didn't see your latest comments. Yes please - as warned I'm not a mobile dev so my next steps would be to just keep prompting in Posthog Code :D 

Thank you!

PostHog/Surveys/SurveySheet.swift
```diff
@@ -11,55 +11,60 @@
...
+                        displayManager.onNextQuestion(index: displayManager.currentQuestionIndex, response: .rating(resp))
                     }
                 case let currentQuestion as PostHogDisplayChoiceQuestion:
                     if currentQuestion.isMultipleChoice {
                         MultipleChoiceQuestionView(question: currentQuestion) { resp in
-                            onNextQuestionClicked(currentQuestionIndex, .multipleChoice(resp))
+                            displayManager.onNextQuestion(index: displayManager.currentQuestionIndex, response: .multipleChoice(resp))
                         }
                     } else {
                         SingleChoiceQuestionView(question: currentQuestion) { resp in
```
turnipdabeets, line 66: The PR note says a language switch can drop an in-progress choice selection, but it's a bit sharper than that: `selectedChoices` in the choice views is `@State Set<String>` keyed by label text and survives the in-place swap — so after a switch the highlight vanishes but `canSubmit` stays true, and submitting records the stale old-language label. For the open choice, `response == openChoice` compares the old label against the new-language open choice, fails, and submits the label instead of the typed text. Could we key selection by choice index, or clear/remap it in `.onChange(of: question.choices)`?

PostHog/Surveys/SurveySheet.swift
```diff
@@ -11,55 +11,60 @@
...
     @available(iOS 15, *)
     struct SurveySheet: View {
-        let survey: PostHogDisplaySurvey
-        let isSurveyCompleted: Bool
-        let currentQuestionIndex: Int
-        let onClose: () -> Void
-        let onNextQuestionClicked: (_ index: Int, _ response: PostHogSurveyResponse) -> Void
+        // Observed so the sheet re-renders when the displayed survey is updated in place
+        // (e.g. re-translated after a language change), not just when questions advance.
+        @ObservedObject var displayManager: SurveyDisplayController
```
turnipdabeets, line 16: [suggestion] Now that the sheet observes the controller live, `dismissSurvey()`'s synchronous reset of `currentQuestionIndex`/`isSurveyCompleted` re-renders question 1 of `fallbackSurvey` during the slide-out (pre-PR, the `let` snapshots kept the outgoing frame frozen). Since `showSurvey` re-initializes both anyway, could we drop those two resets from `dismissSurvey()`?

PostHogTests/PostHogSurveyTranslationsTest.swift
```diff
@@ -360,6 +360,354 @@
...
+                ) {}
+
+                func cleanupSurveys() {}
+            }
+
+            @Suite("Test live translation updates", .serialized)
+            class TestLiveTranslationUpdate {
+                let server: MockPostHogServer
+
+                init() {
```
turnipdabeets, line 412: [suggestion] `postHog.close()`/`reset()` at the tail of each test body get skipped when a `#require` throws (there's one in `getSurveyIntegration` on every path), leaking a live SDK and its subscription into the next `.serialized` test. The suites in PostHogSurveysTest.swift build the SDK in `init()` and tear down in `deinit` — could we do the same here?

PostHog/Surveys/PostHogSurveyIntegration.swift
```diff
@@ -318,6 +328,50 @@
...
+
+                return activeSurvey.toDisplaySurvey(
+                    surveyTranslation: translations.survey,
+                    questionTranslations: translations.questions
+                )
+            }
+
+            guard let displaySurvey else { return }
+
+            DispatchQueue.main.async {
```
turnipdabeets: [suggestion] The lock comment above guarantees commit order, but this enqueue happens after the lock is released — two racing refreshes can commit fr→de yet enqueue de→fr, leaving the screen on fr while `activeSurveyLanguage` (and `$survey_language`) say de. Moving the `DispatchQueue.main.async` inside the `withLock` block makes enqueue order match commit order (async enqueue doesn't block).

PostHog/Surveys/PostHogSurveyIntegration.swift
```diff
@@ -318,6 +328,50 @@
...
+            // calls with different languages) could commit out of order and leave the survey
+            // on a stale language.
+            let displaySurvey: PostHogDisplaySurvey? = activeSurveyLock.withLock {
+                guard let activeSurvey = self.activeSurvey else { return nil }
+
+                let language = resolveDisplayLanguage()
+                let translations = resolveSurveyTranslations(survey: activeSurvey, targetLanguage: language)
+
+                // Commit only if the update targets a different language than what's currently shown
+                guard self.activeSurveyLanguage != translations.matchedKey else { return nil }
```
turnipdabeets: [question] If a refresh lands in the window after `setActiveSurvey` but before the initial `renderSurvey` reaches the controller, the language commits here but the dispatched update is dropped (`displayedSurvey` is nil, no pending survey) — and the next refresh no-ops since the language already matches. Tiny window, sticky outcome. Worth reconciling in `handleSurveyShown` when the shown snapshot's language differs from `activeSurveyLanguage`?

PostHog/Surveys/PostHogSurveysConfig.swift
```diff
@@ -66,6 +66,18 @@ public typealias OnPostHogSurveyClosed = (_ survey: PostHogDisplaySurvey) -> Voi
...
+    /// Called when the content of the currently displayed survey changes without restarting it,
+    /// for example when the user's language is updated and a new translation becomes available.
+    ///
+    /// Implementations should update the on-screen survey in place — preserving the current
+    /// question, progress, and any in-progress answers — rather than presenting it again.
+    /// Optional: delegates that don't support live updates can omit it and the survey simply
+    /// keeps the language it was first rendered with.
+    ///
+    /// - Parameter survey: The survey with refreshed (e.g. re-translated) content. Its `id`
+    ///   matches the survey passed to `renderSurvey`.
```
turnipdabeets, line 83: [suggestion] Two contract points a custom-delegate author needs here: this is always delivered on the main thread, and it can arrive before the survey is visibly shown (e.g. during `surveyPopupDelaySeconds` — that's why the default delegate refreshes `pendingSurvey`). A delegate with its own display delay will silently drop the update otherwise. Could we add both to the doc comment?

PostHog/PostHogRemoteConfig.swift
```diff
@@ -752,6 +757,10 @@ class PostHogRemoteConfig {
             // Persist to disk
             storage.setDictionary(forKey: .personPropertiesForFlags, contents: personPropertiesForFlags)
         }
+        // Notify subscribers (e.g. surveys) so a survey already on screen can re-resolve its
+        // language if the user's `language` property changed. Invoked outside the lock so
+        // subscribers don't run while we hold it.
+        onPersonPropertiesForFlagsChanged.invoke(())
```
turnipdabeets: [nit] This fires even when the merge changed nothing, so every `capture()` carrying `userProperties` re-runs translation resolution under `activeSurveyLock` while a survey is up. Skipping the invoke when no merged key changed value would keep it off hot paths.

PostHog/Surveys/PostHogSurveyIntegration.swift
```diff
@@ -318,6 +328,50 @@
...
+                self.activeSurveyLanguage = translations.matchedKey
+                self.activeSurveyQuestionTranslations = translations.questions
```
turnipdabeets, line 360: [question] Once this commits mid-survey, `survey sent` stamps all `$survey_questions` (and `$survey_language`) with the final translations — answers given before the switch are reported with question text the user never saw, and `survey shown` vs `sent` disagree on the language. Acceptable for analytics, or should we freeze the language after the first response?

ioannisj: Yeah, tbh I think this is a product call? It may make sense to track each answer with the $survey_language that matches the UI at that point. There's no other SDK doing mid-display translation changes, so I can't really reference a previous implementation. @PostHog/team-surveys what's your call here? Should $survey_language correctly reflect the language shown to the user for each question or should we fix it so that all survey events match (even though the user may have seen something else)

PostHog/Surveys/PostHogSurveyIntegration.swift
ioannisj: Good catch, confirmed it's real. The window is nanosecond-scale though: the refresh has to land between `setActiveSurvey` releasing the lock and `renderSurvey` getting enqueued to main, and the impact is display-only since the tracked (analytics) language is already the new one.

Fixed in 6d3803e: I track the language the survey was actually rendered with, kept separate from the committed one and untouched by the refresh so we can spot the drop, and reconcile in `handleSurveyShown`. If the rendered language fell behind it pushes a single `updateSurvey` so the display catches up. Added tests for the reconcile and the no-op case.

turnipdabeets (approved):

PostHog/Surveys/PostHogSurveyIntegration.swift
ioannisj: Asked here https://posthog.slack.com/archives/C07QD3LT8U9/p1784785031568259

PR #689, opened 2026-07-01
fix(error-tracking): late-install integration when remote config loads after first launch

tsushanth:
Fixes #551

## Problem

On first launch with no disk-cached remote config, `isAutocaptureExceptionsEnabled()` returns `false` and `installIntegrations()` skips `PostHogErrorTrackingAutoCaptureIntegration`. When the live remote config fetch later completes with error tracking enabled, nothing re-attempts the install — any crash on that first launch is permanently lost.

## Fix

After `installIntegrations()`, subscribe to `onRemoteConfigLoaded` and call `installErrorTrackingIntegration()` if error tracking wasn't installed. The subscription is one-shot — the token is cleared once it fires. This mirrors the late-install path already used by session replay (`installReplayIntegration()` is called from `startSessionRecording()` when `replayIntegration == nil`).

```swift
#if os(iOS) || os(macOS) || os(tvOS)
    if !isErrorTrackingInstalled() {
        errorTrackingRemoteConfigToken = remoteConfig?.onRemoteConfigLoaded.subscribe { [weak self] _ in
            guard let self else { return }
            setupLock.withLock {
                installErrorTrackingIntegration()
            }
            errorTrackingRemoteConfigToken = nil
        }
    }
#endif
```

`installErrorTrackingIntegration()` follows the same shape as `installReplayIntegration()`:
- guards `isErrorTrackingInstalled()` to avoid double-install
- appends to `installedIntegrations` so the integration participates in context-change notifications

Also added `clearInstalls()` to `PostHogErrorTrackingAutoCaptureIntegration` (matching other integrations) and `getErrorTrackingIntegration()` accessor on `PostHogSDK` for testability.

### DIFF
--- a/PostHog/ErrorTracking/PostHogErrorTrackingAutoCaptureIntegration.swift
+++ b/PostHog/ErrorTracking/PostHogErrorTrackingAutoCaptureIntegration.swift
@@ -26,9 +26,15 @@ import Foundation
         private var crashCustomData: PostHogCrashCustomDataWriter?
         private var contextChangedToken: RegistrationToken?
         private var exceptionStepsChangedToken: RegistrationToken?
+        private var remoteConfigLoadedToken: RegistrationToken?
 
         func install(_ postHog: PostHogSDK) -> PostHogIntegrationInstallResult {
-            if postHog.remoteConfig?.isAutocaptureExceptionsEnabled() == false {
+            // Only block on remote config once it has actually loaded. Before the first
+            // fetch completes (e.g. first launch with no disk cache) we default to installing
+            // so that a crash on that very first launch is not silently missed.
+            if postHog.remoteConfig?.hasFetchedRemoteConfig == true,
+               postHog.remoteConfig?.isAutocaptureExceptionsEnabled() == false
+            {
                 return .skipped(.disabledByRemoteConfig)
             }
 
@@ -50,6 +56,18 @@ import Foundation
                     exceptionStepsChangedToken = postHog.onExceptionStepsChanged.subscribe { [weak crashCustomData] steps in
                         crashCustomData?.setSteps(steps)
                     }
+
+                    // If remote config was not yet loaded at install time, subscribe so we can
+                    // uninstall if the freshly-loaded config disables autocapture.
+                    if postHog.remoteConfig?.hasFetchedRemoteConfig == false {
+                        remoteConfigLoadedToken = postHog.remoteConfig?.onRemoteConfigLoaded.subscribe { [weak self, weak postHog] _ in
+                            guard let self, let postHog else { return }
+                            self.remoteConfigLoadedToken = nil
+                            if postHog.remoteConfig?.isAutocaptureExceptionsEnabled() == false {
+                                self.uninstall(postHog)
+                            }
+                        }
+                    }
                 }
             }
         }
@@ -57,6 +75,7 @@ import Foundation
         func uninstall(_ postHog: PostHogSDK) {
             uninstallIfNeeded(from: postHog, installedPostHog: self.postHog, state: Self.integrationInstallState) {
                 stop()
+                remoteConfigLoadedToken = nil
                 contextChangedToken = nil
                 exceptionStepsChangedToken = nil
                 crashCustomData = nil
--- a/PostHog/ErrorTracking/PostHogErrorTrackingAutoCaptureIntegration.swift
+++ b/PostHog/ErrorTracking/PostHogErrorTrackingAutoCaptureIntegration.swift
@@ -64,7 +64,7 @@ import Foundation
                             guard let self, let postHog else { return }
                             self.remoteConfigLoadedToken = nil
                             if postHog.remoteConfig?.isAutocaptureExceptionsEnabled() == false {
-                                self.uninstall(postHog)
+                                self.stop()
                             }
                         }
                     }
--- a/PostHog/ErrorTracking/PostHogErrorTrackingAutoCaptureIntegration.swift
+++ b/PostHog/ErrorTracking/PostHogErrorTrackingAutoCaptureIntegration.swift
@@ -14,6 +14,10 @@ import Foundation
     class PostHogErrorTrackingAutoCaptureIntegration: PostHogIntegration {
         private static let integrationInstallState = PostHogIntegrationInstallState()
 
+        static func clearInstalls() {
+            integrationInstallState.clear()
+        }
+
         var requiresSwizzling: Bool { false }
 
         private weak var postHog: PostHogSDK?
--- a/PostHog/ErrorTracking/PostHogErrorTrackingAutoCaptureIntegration.swift
+++ b/PostHog/ErrorTracking/PostHogErrorTrackingAutoCaptureIntegration.swift
@@ -29,10 +29,12 @@ import Foundation
         private var remoteConfigLoadedToken: RegistrationToken?
 
         func install(_ postHog: PostHogSDK) -> PostHogIntegrationInstallResult {
-            // Only block on remote config once it has actually loaded. Before the first
-            // fetch completes (e.g. first launch with no disk cache) we default to installing
-            // so that a crash on that very first launch is not silently missed.
-            if postHog.remoteConfig?.hasFetchedRemoteConfig == true,
+            // Block installation when remote config (live or disk-cached) explicitly disables
+            // autocapture. When there is no remote config at all (first launch, no disk cache)
+            // we default to installing so a crash on that very first launch is not missed.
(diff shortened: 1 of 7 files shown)


turnipdabeets:
Two things worth sorting before we go deeper. 

First, this only installs the crash handler after remote config loads, so a crash on that very first launch (before config arrives) still gets missed . In the issue we'd leaned toward just defaulting autocapture to on until config says otherwise. 

Second, right now the late-install doesn't check errorTrackingConfig.autoCapture, so an app that never turned crash capture on would still get a crash reporter installed if the project has it enabled server-side — which could step on their existing one (Crashlytics/Sentry).

We also need to make CI pass and add proper tests

tsushanth:
Thanks @turnipdabeets — both points are clear. Before I rework:

**On default-on until config says otherwise**: to install immediately then uninstall if remote config opts out — would the right pattern be to install the handler at setup time (ignoring remote config), subscribe to `onRemoteConfigLoaded`, and in the callback call `uninstall` if the loaded config has error tracking disabled? Or is there a cleaner 'override' mechanism the session-replay integration uses for the same lifecycle?

**On checking `errorTrackingConfig.autoCapture`**: I can add a guard in `installErrorTrackingIntegration()` that returns early when `postHog.config.errorTrackingConfig.autoCapture == false` — does that cover the case, or does the integration's own `install(_:)` method already gate on that flag?

Once those two are clear I can update the code, add the missing tests, and fix CI.

turnipdabeets:
Cc @ioannisj curious what approach you think is best here

ioannisj:
- Local config always wins over remote config. Integration is already gated and will be installed only when`errorTrackingConfig.autoCapture` [here](https://github.com/PostHog/posthog-ios/blob/3f108c786c8ce1de8aa405287eb517e38f81d599/PostHog/PostHogConfig.swift#L439-L441) so I think no additional logic is required to handle this case. 
- The problematic gate in `PostHogErrorTrackingAutoCaptureIntegration` is this check [here](https://github.com/PostHog/posthog-ios/blob/3f108c786c8ce1de8aa405287eb517e38f81d599/PostHog/ErrorTracking/PostHogErrorTrackingAutoCaptureIntegration.swift#L27-L29). We should instead continue to install the integration when remote config has not been loaded yet, and not depend on cached values. Lots of ways to do this, we could default to `true` for `PostHogRemoteConfig.autoCaptureExceptions` for example, or expose `PostHogRemoteConfig.remoteConfigDidFetch` and check on the integration side, or even remove the check entirely on the integration side if we want autocapture to always be on (when enabled in local config) until fresh remote config value arrives
- Then, similar to how other integrations do (e.g [session replay](https://github.com/PostHog/posthog-ios/blob/3f108c786c8ce1de8aa405287eb517e38f81d599/PostHog/Replay/PostHogReplayIntegration.swift#L180-L183)) the integration itself can subscribe to remote config notifications and react accordingly. e.g uninstall/stop

I believe these should be enough to handle the case of first-installs and stale cached configs?

ioannisj:
@tsushanth Left some rough guidelines [here](https://github.com/PostHog/posthog-ios/pull/689#issuecomment-4862642939) If you’re still interested in finishing this PR, I’m happy to review another iteration. Otherwise, I’m also happy to pick it up if you’ve moved on.

tsushanth:
Updated with the approach you outlined @ioannisj:

1. **Fixed compile error** — added explicit `self.` in the `setupLock.withLock` closure (Swift strict concurrency requires it)
2. **Default install before remote config loads** — the gate in `install()` now checks `hasFetchedRemoteConfig` first; if config hasn't arrived yet we proceed with installation so first-launch crashes aren't missed
3. **React to remote config on late-installs** — if we installed before config loaded, we subscribe to `onRemoteConfigLoaded` and call `uninstall()` if the fresh config disables autocapture — same pattern as session replay's `applyRemoteConfig`

Still need tests — happy to add them if you can point me to the test file / pattern you'd like covered.

ioannisj:
Thanks @tsushanth hasFetchedRemoteConfig needs a visibility change (CI builds are failing). I'd say tests can go in `PostHogTests/PostHogIntegrationInstallationTest.swift` or a new test file if you prefer, no strong preference there as long as the tests are there

tsushanth:
Added tests + fixed the remaining compile error:

**Tests** (in `PostHogIntegrationInstallationTest.swift`):
1. Installs before remote config arrives (first-launch default-on)
2. Does not install when cached remote config is already loaded and says disabled
3. Uninstalls when `/config` arrives with `autocaptureExceptions: false`
4. Stays installed when `/config` arrives with `autocaptureExceptions: true`

**Compile fix**: `hasFetchedRemoteConfig` was inside `#if os(iOS)` only — moved it outside so macOS/tvOS builds can also use it (the integration targets all three platforms).

PostHog/PostHogSDK.swift
```diff
@@ -222,6 +223,21 @@ let maxRetryDelay = 30.0
                 // don't install integrations if in opt-out state
                 installIntegrations()
 
+                // Error tracking may have been skipped because remote config was not yet loaded
+                // (first launch, no disk cache). Subscribe to re-attempt installation once the
+                // live remote config arrives — mirroring the late-install path session replay uses.
```
ioannisj: This is not true anymore, so these changes here should be reverted

PostHog/ErrorTracking/PostHogErrorTrackingAutoCaptureIntegration.swift
```diff
@@ -46,13 +56,26 @@ import Foundation
...
                     }
+
+                    // If remote config was not yet loaded at install time, subscribe so we can
+                    // uninstall if the freshly-loaded config disables autocapture.
+                    if postHog.remoteConfig?.hasFetchedRemoteConfig == false {
+                        remoteConfigLoadedToken = postHog.remoteConfig?.onRemoteConfigLoaded.subscribe { [weak self, weak postHog] _ in
+                            guard let self, let postHog else { return }
+                            self.remoteConfigLoadedToken = nil
+                            if postHog.remoteConfig?.isAutocaptureExceptionsEnabled() == false {
+                                self.uninstall(postHog)
```
ioannisj: This leaves `PostHogSDK.installedIntegrations` in a bad state, however. so we either just `stop` the integration (remains installed but crashReporter is released) or we add a way to uninstall the integration (behind proper locks) from PostHogSDK instance (e.g posthog.uninstall(self))

I would probably just move everything in `uninstall(_ postHog: PostHogSDK)` into what is now a no-op `stop()` and call stop() here. Integration remains installed but it's no-op because of remote config. Up to you, however

PostHog/ErrorTracking/PostHogErrorTrackingAutoCaptureIntegration.swift
```diff
@@ -46,13 +56,26 @@ import Foundation
...
+
+                    // If remote config was not yet loaded at install time, subscribe so we can
+                    // uninstall if the freshly-loaded config disables autocapture.
+                    if postHog.remoteConfig?.hasFetchedRemoteConfig == false {
+                        remoteConfigLoadedToken = postHog.remoteConfig?.onRemoteConfigLoaded.subscribe { [weak self, weak postHog] _ in
+                            guard let self, let postHog else { return }
+                            self.remoteConfigLoadedToken = nil
+                            if postHog.remoteConfig?.isAutocaptureExceptionsEnabled() == false {
+                                self.stop()
+                            }
```
dustinbyrne, line 70: this looks like a no-op given the definition of `stop` for this class

PostHog/ErrorTracking/PostHogErrorTrackingAutoCaptureIntegration.swift
ioannisj: @tsushanth https://github.com/PostHog/posthog-ios/pull/689/changes#r3560987406

turnipdabeets:
@tsushanth friendly ping on this. Could we also add a changeset?

tsushanth:
Fixed the test race: added `configResponseDelay` to `MockPostHogServer` (mirrors the existing `flagsResponseDelay` pattern) and set it to 0.5 s in the racy test so the pre-config assertion at line 184 always runs before /config arrives. Changeset already committed in the previous push.

akarabach (commented):
lgtm

ioannisj:
Thanks for kicking this off, @tsushanth. The first-launch race is a real bug and your approach is what shipped. I've taken it over in #731 to carry it across the line: re-signed the commits (branch protection needs signed history), folded in the review feedback (using `removeIntegration`/`uninstall` instead of the no-op `stop`, and reverting the late-subscribe approach), and added the changeset. I'm closing this in favour of the takeover; your commits and authorship are preserved there. Thanks again for the contribution!

PR #704, opened 2026-07-08
fix: Session Replay crash on AVAggregateAssetDownl oadTask when resume() probes currentRequest

### LINKED ISSUE #703: Session Replay crashes on AVAggregateAssetDownloadTask when resume() probes currentRequest
### Version

3.58.1

### Steps to Reproduce

- iOS app
- AVAssetDownloadURLSession
- HLS/offline download via AVAggregateAssetDownloadTask
- PostHog Session Replay enabled
- Session Replay network telemetry enabled
- crash occurs on task.resume()

### Expected Result

Request rewriting in the resume() swizzle should be best-effort:
- if currentRequest cannot be read safely, skip rewriting
- if mutating the request fails, skip it
- always continue to the original resume() implementation
- unsupported task subclasses should never crash the host app

### Actual Result

```
Fatal Exception: NSGenericException
0  CoreFoundation                             0x11523c __exceptionPreprocess
1  libobjc.A.dylib                            0x31224 objc_exception_throw
2  CFNetwork                                  0x1a2454 -[__NSCFBackgroundAVAggregateAssetDownloadTaskNoChildTask originalRequest]
3  PostHog_37D97A3654AA4B4B_PackageProduct    0x995f0 URLSessionSwizzler.TaskResume.modifyTaskRequests(_:) + 309 (URLSessionSwizzler.swift:309)
4  PostHog_37D97A3654AA4B4B_PackageProduct    0x9950c closure #1 in closure #1 in URLSessionSwizzler.TaskResume.swizzle() + 275 (URLSessionSwizzler.swift:275)
5  PostHog_37D97A3654AA4B4B_PackageProduct    0x10958 <deduplicated_symbol>
6  Downloads                                  0xf7b0 partial apply for closure #3 in closure #1 in closure #2 in AssetDownloadService.triggerNewDownload(for:assetURL:updateDownloadQueue:backupAssetURL:) + 410 (AssetDownloadService.swift:410)
```

wowrumal:
Fixes #703.

Session Replay network instrumentation currently assumes every URLSessionTask subclass safely exposes and mutates currentRequest during the swizzled resume() path. That breaks for AVFoundation-backed download tasks such as AVAggregateAssetDownloadTask, which can raise NSGenericException on request access instead of returning nil.

This change makes request probing and mutation best-effort by routing both operations through a small Objective-C @try/@catch helper. Unsupported task subclasses now skip request rewriting instead of crashing the host app.

Ran the focused iOS regression suite:
`xcrun xcodebuild test -scheme PostHog -destination 'platform=iOS Simulator,OS=18.6,name=iPhone 16' -only-testing:PostHogTests/PostHogSessionReplayPluginTests`

Ran a SwiftPM build to verify the new Objective-C helper target and package wiring:
`swift build`

## :pencil: Checklist

## If releasing new changes

## 🤖 Agent context
Autonomy: Human-driven (agent-assisted)

This change was prepared with Codex in Conductor.

A narrow class-name skip for AVFoundation tasks was considered as a stopgap, but it only fixes the reported crash and leaves the underlying assumption intact. The final change uses an internal Objective-C exception boundary because the real failure mode is NSException from request access/mutation, and Swift cannot catch that safely.

Regression coverage was added for three cases: AVFoundation-style getter exceptions, generic getter exceptions, and setter exceptions, while preserving the normal request-rewrite path for standard request-backed tasks.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -26,13 +26,22 @@ let package = Package(
         .target(
             name: "PostHog",
             dependencies: [
+                "PostHogObjCExceptionSupport",
                 "phlibwebp",
                 .target(name: "PHPLCrashReporter", condition: .when(platforms: [.iOS, .macOS, .tvOS])),
             ],
             path: "PostHog",
+            exclude: [
+                "ObjCExceptionSupport",
+            ],
             resources: [
                 .copy("Resources/PrivacyInfo.xcprivacy"),
             ]
+        ),
+        .target(
+            name: "PostHogObjCExceptionSupport",
+            path: "PostHog/ObjCExceptionSupport",
+            publicHeadersPath: "."
         ),
         .target(
             name: "phlibwebp",
--- a/PostHog/Replay/Plugins/Network/URLSessionSwizzler.swift
+++ b/PostHog/Replay/Plugins/Network/URLSessionSwizzler.swift
@@ -9,6 +9,7 @@
 #if os(iOS)
 
     import Foundation
+    @_implementationOnly import PostHogObjCExceptionSupport
 
     final class URLSessionInstrumentation {
         typealias RequestModifier = (URLRequest) -> URLRequest
@@ -306,13 +307,23 @@
                 }
             }
 
-            private func modifyTaskRequests(_ task: URLSessionTask) {
+            // Internal, used for testing
+            func modifyTaskRequests(_ task: URLSessionTask) {
                 // Only rewrite `currentRequest` for the async/await fallback path.
                 // This is the request closest to what will go over the wire and avoids
-                // mutating `originalRequest` unnecessarily. Use KVC instead of invoking
-                // a private setter selector directly.
-                if let currentRequest = task.currentRequest {
-                    task.setValue(modifyRequest(currentRequest) as NSURLRequest, forKey: Self.currentRequestKey)
+                // mutating `originalRequest` unnecessarily. Some task subclasses throw
+                // on request access or mutation, so this path has to stay best-effort.
+                guard let currentRequest = PHURLSessionTaskSafeAccess.currentRequest(from: task) as URLRequest? else {
+                    hedgeLog("[Session Replay] Skipping request header injection for task \(NSStringFromClass(type(of: task))) because currentRequest is unavailable")
+                    return
+                }
+
+                if !PHURLSessionTaskSafeAccess.setCurrentRequest(
+                    modifyRequest(currentRequest),
+                    on: task,
+                    key: Self.currentRequestKey
+                ) {
+                    hedgeLog("[Session Replay] Failed to inject tracing headers into currentRequest for task \(NSStringFromClass(type(of: task)))")
                 }
             }
         }
--- a/PostHogTests/PostHogSessionReplayPluginTest.swift
+++ b/PostHogTests/PostHogSessionReplayPluginTest.swift
@@ -1,6 +1,9 @@
 #if os(iOS)
     import Foundation
     @testable import PostHog
+    #if !SWIFT_PACKAGE
+        import CwlCatchException
+    #endif
     import Testing
 
     @Suite("Session Replay Plugin Remote Config Tests")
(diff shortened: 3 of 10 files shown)


Package.swift
```diff
@@ -34,6 +35,11 @@ let package = Package(
                 .copy("Resources/PrivacyInfo.xcprivacy"),
             ]
         ),
+        .target(
+            name: "PostHogObjCExceptionSupport",
+            path: "vendor/ObjCExceptionSupport",
```
turnipdabeets: [question] `vendor/` has been third-party-only so far (`vendor/UPSTREAM.md` and license tooling assume that), and this helper is first-party. Was `PostHog/ObjCExceptionSupport` + an `exclude:` on the main target not workable, or is `vendor/` deliberate?

PostHog/Replay/Plugins/Network/URLSessionSwizzler.swift
```diff
@@ -306,14 +307,20 @@
...
-                // a private setter selector directly.
-                if let currentRequest = task.currentRequest {
-                    task.setValue(modifyRequest(currentRequest) as NSURLRequest, forKey: Self.currentRequestKey)
+                // mutating `originalRequest` unnecessarily. Some task subclasses throw
+                // on request access or mutation, so this path has to stay best-effort.
+                guard let currentRequest = PHObjCTryCatch.currentRequest(from: task) as URLRequest? else {
+                    return
                 }
+
+                _ = PHObjCTryCatch.setCurrentRequest(
```
turnipdabeets: [suggestion] Both bail-out paths are silent (the early return above, and the discarded result here) — when this fires, tracing headers just quietly never attach for that task. Could we `hedgeLog` on failure so the next "session correlation missing on some requests" ticket is diagnosable?

PostHog/Replay/Plugins/Network/URLSessionSwizzler.swift
```diff
@@ -306,14 +307,20 @@
                 }
             }
 
-            private func modifyTaskRequests(_ task: URLSessionTask) {
+            func modifyTaskRequests(_ task: URLSessionTask) {
```
turnipdabeets, line 311: [suggestion] Repo convention marks members widened for tests (cf. `PostHogQueue.swift:102`):
```suggestion
            // Internal, used for testing
            func modifyTaskRequests(_ task: URLSessionTask) {
```

PostHogTests/PostHogSessionReplayPluginTest.swift
```diff
@@ -1,6 +1,9 @@
 #if os(iOS)
     import Foundation
     @testable import PostHog
+    #if !SWIFT_PACKAGE
+        import CwlCatchException
```
turnipdabeets, line 5: [nit] `CwlCatchException` isn't declared on the PostHogTests target — it resolves transitively through Nimble today, so a Nimble bump could silently break these tests. Worth declaring it explicitly in the xcodeproj.

PostHogTests/PostHogSessionReplayPluginTest.swift
```diff
@@ -218,5 +221,143 @@
...
+                }
+
+                let url = try #require(URL(string: "https://example.com"))
+                let task = ThrowingCurrentRequestSetterTask(request: URLRequest(url: url))
+
+                let exception = NSException.catchException {
+                    swizzler.modifyTaskRequests(task)
+                }
+
+                #expect(exception == nil)
```
turnipdabeets, line 312: [suggestion] This proves we don't crash, but not that we degraded to a no-op — if the helper misbehaved while still swallowing the exception, it'd pass. Cheap to also pin the request:
```suggestion
                #expect(exception == nil)
                #expect(task.currentRequest?.value(forHTTPHeaderField: "X-PostHog-Test") == nil)
```

PostHogTests/PostHogSessionReplayPluginTest.swift
```diff
@@ -218,5 +221,143 @@
...
+                let task = ThrowingCurrentRequestTask()
+
+                let exception = NSException.catchException {
+                    swizzler.modifyTaskRequests(task)
+                }
+
+                #expect(exception == nil)
+            }
+
+            @Test("Task resume fallback still rewrites standard request-backed tasks")
```
turnipdabeets, line 281: [suggestion] Could we add a test for the legitimate-nil case — a task whose `currentRequest` returns nil without throwing (the common not-yet-started case)? Today the guard's nil branch is only reached via the exception→nil mapping, so a regression there would go unseen.

PostHogTests/PostHogSessionReplayPluginTest.swift
```diff
@@ -218,5 +221,143 @@
                 ) == testCase.expected
             )
         }
+
+        #if !SWIFT_PACKAGE
+            @Test("Task resume fallback ignores AVFoundation tasks that throw on currentRequest")
```
turnipdabeets, line 226: [suggestion] All four tests call `modifyTaskRequests` directly, so nothing verifies the swizzled composition — that after a caught throw, `resume()` still falls through to the original implementation and the download actually starts. One integration-style test (swizzle + stubbed request + `defer` unswizzle) would lock that in — though no need to grow this PR for it; we're happy to take it as a follow-up on our side.

turnipdabeets (commented):
LGTM -  would appreciate @ioannisj opinion on this PR too.

A couple of house keeping items:

- We need a changeset to release. Please see RELEASING.md
- All of your commits need to be signed to merge

Package.swift
wowrumal: Just misunderstood vendors definition. Reverted back to the first option

PostHog/Replay/Plugins/Network/URLSessionSwizzler.swift
wowrumal: Added logs to the failure paths

PostHog/Replay/Plugins/Network/URLSessionSwizzler.swift
wowrumal: Added comment

PostHogTests/PostHogSessionReplayPluginTest.swift
wowrumal: Good catch. I added `CwlCatchException` as a direct dependency of `PostHogTests` so the tests no longer rely on Nimble bringing it in transitively

PostHogTests/PostHogSessionReplayPluginTest.swift
wowrumal: Added a dedicated nil-path test as well, so the `currentRequest == nil` branch is covered independently of the exception-to-nil path

PostHogTests/PostHogSessionReplayPluginTest.swift
wowrumal: Tightened the tests to assert the failure paths degrade to a no-op, not just that they avoid crashing

PostHogTests/PostHogSessionReplayPluginTest.swift
wowrumal: Agreed. The current tests only cover the request-rewrite fallback directly. I left the full swizzled `resume()` integration case out to keep this PR scoped, but it would be useful follow-up coverage to verify we still fall through to the original implementation after a caught exception

turnipdabeets:
One more request: could we rename PHObjCTryCatch to match what it actually does, e.g. PHURLSessionTaskSafeAccess? It currently reads as a generic exception guard, but we'd like to keep it narrow rather than generalize.

wowrumal:
Sure. Renamed it to `PHURLSessionTaskSafeAccess` and updated the internal packaging references

turnipdabeets:
> Commits must have verified signatures.

Mind doing this so we can merge? :)

wowrumal:
Squashed and force pushed signed commit :)

turnipdabeets (approved):

haritowa:
@turnipdabeets Give that man a changelog mention, he is crying to my DM 😭

Issue #702, opened 2026-07-07
Event-queue peek/pop positional misalignment can duplicate event delivery when a file is skipped


tsushanth:
### Description
`PostHogFileBackedQueue.loadFiles` (backing `peek`) silently drops files it can't read -- a missing file (`continue`) or one that throws on `Data(contentsOf:)`, which it treats as corrupted, deletes, and skips -- but does not remove those entries from the in-memory `items` array. The consumer then pops by the count of `Data` actually returned, assuming a 1:1 positional correspondence with the head of `items`. When a skipped file precedes returned items within the scanned prefix, `pop` under-removes and leaves already-sent records in the queue, which are re-sent on the next flush.

### Root cause
- `PostHog/PostHogFileBackedQueue.swift`, `loadFiles`: missing file -> `continue`; corrupt file -> `deleteSafely(itemURL)` then continue. Neither removes the item from `items`.
- `PostHog/PostHogQueue.swift`: `let items = self.fileQueue.peek(count)`, then on success `self.fileQueue.pop(items.count)` (i.e. `deleteFiles` from index 0).

Trace: `items = [A(corrupt), B, C, D]`, cap 2: `peek(2)` deletes A, returns `[B, C]`, `items` still `[A, B, C, D]`; batch sends B and C; `pop(2)` removes A and B -> `items = [C, D]`; C was already sent but stays queued -> re-sent (duplicate). Real trigger: locked-device file protection (`NSFileReadNoPermission`) makes `Data(contentsOf:)` throw for otherwise-valid files (cf. #213).

Note: this is a deterministic single-threaded static trace of the code paths, not a full runtime reproduction.

### Suggested fix
Make `peek` return identities (filenames) alongside data and have the consumer `pop` those exact filenames; or have `loadFiles` remove skipped/corrupt entries from `items` as it deletes them.

PR #709, opened 2026-07-09
fix(api): don't crash when a flags/remote-config response isn't HTTPURLResponse

turnipdabeets:

`PostHogApi` force-casts the URL response in two handlers:

```swift
let httpResponse = response as! HTTPURLResponse
```

- `PostHog/PostHogApi.swift:325` — feature flags handler
- `PostHog/PostHogApi.swift:412` — remote config handler

If `response` is ever `nil` (or not an `HTTPURLResponse`), the force-cast **crashes the host app** with `Fatal error: Unexpectedly found nil while unwrapping an Optional value`. An SDK must never crash the app that embeds it. This surfaced as an intermittent crash in the flags handler during CI (`PostHogApi.swift:325`).

- `swift build -Xswiftc -DTESTING` compiles cleanly.
- SwiftLint / SwiftFormat clean on the changed files.
- The existing `PostHogApi` suites (flags, remote config, request headers, content-encoding, host-scoping) exercise the happy path and continue to pass.
- Added regression tests (`PostHogApiTests` → "Non-HTTPURLResponse handling"): a custom `URLProtocol` injected through the public `config.urlSessionConfiguration` seam replies with a plain `URLResponse`, and both `flags` and `remoteConfig` now complete gracefully with `(nil, nil)`. Verified the flags test crashes the test runner when the force-cast is reinstated.

## Changes

Both sites now guard instead of force-cast, logging and returning gracefully:

```swift
guard let httpResponse = response as? HTTPURLResponse else {
    hedgeLog("Error parsing the flags response: unexpected response type")
    return completion(nil, nil)
}
```

## :pencil: Checklist

## If releasing new changes

## 🤖 Agent context

Spun out of a review of the #702 queue fix, where CI surfaced this pre-existing force-unwrap crash in the networking layer. Kept as its own PR (not folded into #702) since it is an unrelated crash-safety fix. Grepped the file for every `as!`, found both response force-casts, and replaced them with graceful guards matching the existing error-handling style.

### DIFF
--- a/.changeset/brave-apis-guard.md
+++ b/.changeset/brave-apis-guard.md
@@ -1,0 +1,5 @@
+---
+'posthog-ios': patch
+---
+
+Avoid a crash in the flags and remote-config API handlers when the URL response is not an `HTTPURLResponse` (previously force-cast); the SDK now logs and returns gracefully instead.
--- a/PostHog/PostHogApi.swift
+++ b/PostHog/PostHogApi.swift
@@ -322,7 +322,10 @@ class PostHogApi {
                 return completion(nil, nil)
             }
 
-            let httpResponse = response as! HTTPURLResponse
+            guard let httpResponse = response as? HTTPURLResponse else {
+                hedgeLog("Error parsing the flags response: unexpected response type")
+                return completion(nil, nil)
+            }
 
             if !(200 ... 299 ~= httpResponse.statusCode) {
                 let jsonBody = fromJSONData(data, options: .allowFragments)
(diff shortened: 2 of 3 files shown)


dustinbyrne (approved):
worth a test or two

turnipdabeets:
Added regression tests in 6ea20c2 (@dustinbyrne): a custom `URLProtocol` injected via the public `config.urlSessionConfiguration` seam returns a plain `URLResponse` (not `HTTPURLResponse`), and both the flags and remote-config handlers now complete gracefully with `(nil, nil)`. Sanity-checked that the flags test crashes the test runner if the force-cast is reinstated.

PR #714, opened 2026-07-13
feat(replay): internal session-replay snapshot hook for out-of-engine capture

turnipdabeets:

First-party wrapper SDKs (starting with posthog-flutter) need to record native screens that cover their engine-rendered UI — paywalls and other view controllers presented on top of a Flutter surface. The wrapper runs its own out-of-engine occlusion detection and needs to drive a one-shot capture of the current native window on its own cadence.

This adds an SPI (`@_spi(PostHogInternal)`) hook for that. It is **not public API**: it carries no source/binary stability guarantees and shares snapshot state with the normal timer-driven capture, so only a coordinating first-party SDK should call it.

Consumed by PostHog/posthog-flutter#473. PostHog/posthog-flutter#151 is the end-user issue this enables fixing.

## :hammer: Changes

- `PostHogSDK.captureSessionReplaySnapshot(afterScreenUpdates:)` — SPI entry point; callable from any thread, hops to main for the capture, returns whether a frame was captured so the caller can retry.
- `PostHogReplayIntegration.captureBridgeSnapshot(settlePresentation:)` — renders and enqueues the current key window's screenshot. `settlePresentation` passes `afterScreenUpdates: true` (used only on an episode's first frame — it flickers secure text fields, so steady-state ticks avoid it).
- Bridged frames carry the covering screen's name: the capture resolves the window's topmost presented view controller and passes its name as the meta `href`, and an episode's first frame re-arms the meta event so every episode opens with one. Native (non-bridge) capture is unaffected — the re-arm is gated on the bridge-only first-frame flag.
- `renderAndEnqueueScreenshot` now returns `Bool` and bails (`return false`) on a failed capture, so nothing is enqueued and the caller can retry.
- First-frame re-arm also resets `lastImageHash`, so a recurring native screen always re-sends its opening frame under the always-on screenshot dedup (#711).

- Exercised end-to-end from posthog-flutter's example app on the iOS simulator: native paywall presented over Flutter, both standard presentation and Superwall-style own-window presentation. The episode held for the full duration and captured the native screen.
- swiftformat clean; compiled via the posthog-flutter example app build against this branch.
- There is no unit coverage for the bridge path (it needs a host window); the driving logic is tested on the Flutter side (`native_occlusion_episode_test.dart` over there).

## :pencil: Checklist

## If releasing new changes

## 🤖 Agent context

- Implemented with Claude Code (Opus 4.8 and Fable 5 sessions) driven by @turnipdabeets, alongside the matching posthog-android hook and the posthog-flutter feature that consumes both.
- Design decisions: SPI over public API (no stability commitment for a wrapper-only hook); reuse of the existing screenshot path + masking rather than a parallel pipeline; `afterScreenUpdates` only on episode-first frames because it visibly flickers secure text fields; per-episode meta re-arm added after review so replay names the covering native screen instead of keeping the stale wrapper screen name.
- A multi-agent review pass over the combined feature verified the episode/dedup interactions and confirmed the meta re-arm cannot affect native (non-bridge) capture.

### DIFF
--- a/PostHog/PostHogSDK.swift
+++ b/PostHog/PostHogSDK.swift
@@ -2378,11 +2378,12 @@ let maxRetryDelay = 30.0
         /// timer-driven capture. Returns false if no frame was captured, so the
         /// caller can retry.
         ///
-        /// Pass [afterScreenUpdates] until the episode's first frame has been
-        /// *captured* (returned true) — not just on the first attempt: it also
-        /// re-arms the per-window meta and dedup hash, so a retried opening
-        /// frame keeps its reset. Drop it for steady-state frames (it flickers
-        /// secure fields).
+        /// Pass [episodeFirstFrame] until the episode's first frame has been
+        /// *captured* (returned true) — not just on the first attempt: it
+        /// renders with `afterScreenUpdates` so a freshly-presented screen
+        /// isn't captured black, and re-arms the per-window meta and dedup
+        /// hash, so a retried opening frame keeps its reset. Drop it for
+        /// steady-state frames (it flickers secure fields).
         ///
         /// Prefer the main thread. An off-main call blocks on a synchronous
         /// main-queue hop for the duration of the capture, so it must not come
@@ -2390,13 +2391,13 @@ let maxRetryDelay = 30.0
         ///
         /// SPI, not public API: no stability guarantees.
         @_spi(PostHogInternal) @discardableResult public func captureSessionReplaySnapshot(
-            afterScreenUpdates: Bool
+            episodeFirstFrame: Bool
         ) -> Bool {
             if !isEnabled() {
                 return false
             }
 
-            return replayIntegration?.captureBridgeSnapshot(settlePresentation: afterScreenUpdates) ?? false
+            return replayIntegration?.captureBridgeSnapshot(episodeFirstFrame: episodeFirstFrame) ?? false
         }
     #endif
 
--- a/PostHog/PostHogSDK.swift
+++ b/PostHog/PostHogSDK.swift
@@ -2375,9 +2375,18 @@ let maxRetryDelay = 30.0
         /// Captures the current native window for a first-party wrapper SDK
         /// (e.g. posthog-flutter) that drives session-replay capture on its own
         /// cadence. Not for app use — it shares snapshot state with the normal
-        /// timer-driven capture. Pass [afterScreenUpdates] only for an episode's
-        /// first frame (it flickers secure fields). Returns false if no frame
-        /// was captured, so the caller can retry.
+        /// timer-driven capture. Returns false if no frame was captured, so the
+        /// caller can retry.
+        ///
+        /// Pass [afterScreenUpdates] until the episode's first frame has been
+        /// *captured* (returned true) — not just on the first attempt: it also
+        /// re-arms the per-window meta and dedup hash, so a retried opening
+        /// frame keeps its reset. Drop it for steady-state frames (it flickers
+        /// secure fields).
+        ///
+        /// Prefer the main thread. An off-main call blocks on a synchronous
+        /// main-queue hop for the duration of the capture, so it must not come
+        /// from a queue the main thread can be waiting on.
         ///
         /// SPI, not public API: no stability guarantees.
         @_spi(PostHogInternal) @discardableResult public func captureSessionReplaySnapshot(
--- a/PostHog/PostHogSDK.swift
+++ b/PostHog/PostHogSDK.swift
@@ -2371,6 +2371,24 @@ let maxRetryDelay = 30.0
 
             replayIntegration.stop()
         }
+
+        /// Captures the current native window for a first-party wrapper SDK
+        /// (e.g. posthog-flutter) that drives session-replay capture on its own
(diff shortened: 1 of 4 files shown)


PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -1054,34 +1068,43 @@
...
                 }
+                wireframe.image = image
+                captureSnapshot(
+                    wireframe,
+                    window: window,
+                    windowSize: windowSize,
+                    screenName: screenName,
+                    postHog: postHog,
+                    timestampDate: timestampDate,
+                    isFirstFrame: afterScreenUpdates
```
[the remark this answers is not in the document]
turnipdabeets: Good catch on the contract being under-specified — the coordinated caller (posthog-flutter) keys afterScreenUpdates on "episode started" (set only when a call returned true), not on attempt count, so a failed first attempt retries with the flag still true and the re-arm rides along. The doc did undersell that; tightened in 6a4a61012 to say "until the episode's first frame has been captured".

PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -1347,6 +1371,48 @@
...
+        /// [settlePresentation] renders with `afterScreenUpdates` so a
+        /// freshly-presented screen isn't captured black — pass it only for an
+        /// episode's first frame (it flickers secure fields). Returns false if
+        /// no frame was captured, so the caller can retry.
+        @discardableResult
+        func captureBridgeSnapshot(settlePresentation: Bool) -> Bool {
+            guard Thread.isMainThread else {
+                return DispatchQueue.main.sync {
+                    captureBridgeSnapshot(settlePresentation: settlePresentation)
+                }
```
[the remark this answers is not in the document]
turnipdabeets: The only first-party caller invokes this from the main thread (its occlusion ticker is a main-run-loop timer), so the sync hop is the off-main escape hatch rather than the hot path — and the return value is load-bearing for the caller's retry/fallback logic, which rules out going async here. Documented the constraint (prefer main; never call from a queue the main thread can wait on) in 6a4a61012.

PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -1347,6 +1371,48 @@
...
+            }
+            guard let postHog, postHog.isSessionReplayActive() else {
+                return false
+            }
+            guard let window = UIApplication.getCurrentWindow() else {
+                return false
```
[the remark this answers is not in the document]
turnipdabeets: Fair in principle for multi-scene apps. Two mitigations today: the wrapper's occlusion detector is scoped to the Flutter window's own scene (it only reports covers there), and Flutter embeds are single-scene in practice, where getCurrentWindow() and the detected cover coincide. We kept the SPI parameterless deliberately — no stability guarantees means we can add a window-targeting parameter the moment a multi-scene wrapper needs it, rather than committing to a UIWindow-shaped contract now. Tracking it as the planned extension.

PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -1347,6 +1371,50 @@
...
+            }
+            guard tryStartScreenshotRender() else {
+                return false
+            }
+            // Name the covering screen, not the window's root: bridged frames
+            // show the presented native screen, and the replay meta should say
+            // so. Falls back to the root for SDK-owned windows that install
+            // their screen as rootViewController directly.
+            var coveringController = window.rootViewController
```
ioannisj: I think this will miss nav controllers. Can we reuse `UIViewController.ph_topViewController` here?

PostHog/PostHogSDK.swift
```diff
@@ -2371,6 +2371,33 @@ let maxRetryDelay = 30.0
...
+        ///
+        /// SPI, not public API: no stability guarantees.
+        @_spi(PostHogInternal) @discardableResult public func captureSessionReplaySnapshot(
+            afterScreenUpdates: Bool
+        ) -> Bool {
+            if !isEnabled() {
+                return false
+            }
+
+            return replayIntegration?.captureBridgeSnapshot(settlePresentation: afterScreenUpdates) ?? false
```
ioannisj: nit: This bool param changes 3 names as it's propagated through afterScreenUpdates -> settlePresentation -> isFirstFrame which is a bit confusing

ioannisj (approved):
LG left a couple of comments

PostHog/Replay/PostHogReplayIntegration.swift
turnipdabeets: Good call — the manual walk would name the nav controller instead of its visible child. Swapped it for `UIViewController.ph_topViewController(base: window.rootViewController)`, which keeps the window scoping and adds the nav/tab descent. Fixed in 147e80b.

PostHog/PostHogSDK.swift
turnipdabeets: Agreed — the flag really means "episode's first frame" and the settle render is derived from it, so I renamed the whole chain to `episodeFirstFrame` (SPI param included — no stability guarantee and the only caller is ours in posthog-flutter, updated there too) and kept `afterScreenUpdates` only at the `toImage` call. Fixed in 147e80b.

PR #715, opened 2026-07-14
feat: add bootstrap config to pre-seed identity and feature flags

ioannisj:

The SDK has no way to seed identity or feature flags before `setup()` finishes and the first `/flags` response lands. So events captured synchronously during init (`Application Installed`/`Application Updated`, pre-identify lifecycle events) go out with the SDK-generated UUID instead of a distinct ID the caller controls, and flag reads fall back to not-loaded defaults during cold start.

This adds a `bootstrap` config surface that mirrors the `bootstrap` option in posthog-js. You set `config.bootstrap = PostHogBootstrapConfig(...)` before `setup()` to seed a distinct or anonymous ID (with an `isIdentifiedId` flag) plus flag values and payloads. Bootstrapped identity is first-session only: it never overrides a persisted anonymous ID or an already-identified user, and it keeps the identified ID out of `$device_id`. Bootstrapped flags are a temporary base layer served until the first `/flags` response: a complete response replaces them entirely (bootstrapped-only keys are not retained), while a partial or errored response merges and keeps the keys it didn't recompute. The base layer is cleared on `reset()`. `$feature_flag_called` picks up `$feature_flag_bootstrapped_response`/`_payload` and `$used_bootstrap_value`.

Worth a follow-up to extract the bootstrap flag state out of `PostHogRemoteConfig.swift`, which is getting large.

Added unit tests in `PostHogFeatureFlagsTest`, `PostHogStorageManagerTest`, and `PostHogSDKTest`. They cover immediate flag serving before the first load, sanitizing non-serializable values (NaN/Infinity) so setup can't crash the host, payload serving, loaded values overriding the bootstrap, a complete `/flags` load dropping bootstrapped-only keys, first-session-only behavior after `reset()`, identity reconciliation (merging an anonymous user into an identified bootstrap, and preserving a different already-identified user), fresh-install seeding where an identified bootstrap keeps a random device ID, early lifecycle events carrying the reconciled identity, the intentional `personProfiles == .never` parity with posthog-js (fresh install applies an identified bootstrap, a returning anonymous user drops it), and the `$feature_flag_called` enrichment. The bootstrap suites (`PostHogFeatureFlagsTest`, `PostHogStorageManagerTest`, `PostHogSDKTest`) pass. Also regenerated the public API snapshot with `make apiUpdate`.

## :pencil: Checklist

## If releasing new changes

## 🤖 Agent context

**DRI:** @ioannisj


PostHog/PostHogSDK.swift
```diff
@@ -240,6 +240,36 @@ let maxRetryDelay = 30.0
             DispatchQueue.main.async {
                 NotificationCenter.default.post(name: PostHogSDK.didStartNotification, object: nil)
             }
+
+            // Reconcile inside the setup lock so the merge is atomic with the rest of setup:
+            // no other thread can observe the pre-merge identity or race identify()'s
+            // check-then-act. setupLock is recursive, so identify()'s own setupLock reads are
+            // safe re-entrancy; its heavier work (queueing, reloadFeatureFlags) is dispatched
+            // async, so the critical section isn't extended by network/queue work. A fresh
+            // install is already seeded in PostHogStorageManager; this only reconciles an
```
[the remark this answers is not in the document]
ioannisj: Good catch, this is real for the existing-anonymous install + identified bootstrap case, where the lifecycle publisher replays `didFinishLaunching` synchronously during `installIntegrations`. Fixed in d3f679c0e2: moved `reconcileBootstrapIdentityIfNeeded()` to before `installIntegrations()` (still inside `setupLock`), so the `identify()` merge runs first and the early `Application Installed`/`Application Updated` events carry the bootstrapped identity.

PostHog/PostHogRemoteConfig.swift
```diff
@@ -119,6 +132,9 @@ class PostHogRemoteConfig {
                         hedgeLog("hasFeatureFlags is false, clearing flags and skipping loading flags")
                         // Server responded with explicit hasFeatureFlags: false, meaning no active flags on the account
                         clearFeatureFlags()
+                        // bootstrapped flags are a caller-provided base layer, not server state;
+                        // re-seed them so they stay available until a real /flags response overlays them
+                        seedBootstrapFlagsIfNeeded()
                         // need to notify cause people may be waiting for flags to load
```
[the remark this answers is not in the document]
ioannisj: Yeah, that's inconsistent. Fixed in b8023dc9d5: the `hasFeatureFlags == false` branch now notifies with the served flags (`getFeatureFlags()`) instead of `[:]`, so `onFeatureFlagsLoaded` matches what `getFeatureFlag()` returns.

PostHog/PostHogRemoteConfig.swift
```diff
@@ -413,13 +429,19 @@ class PostHogRemoteConfig {
...
+                        // overlapping keys, while bootstrapped-only keys remain available
+                        let mergedFeatureFlags = self.bootstrappedFlags.merging(featureFlags) { _, loaded in loaded }
+                        let mergedFeatureFlagPayloads = self.bootstrappedPayloads.merging(featureFlagPayloads) { _, loaded in loaded }
+                        loadedFeatureFlags = mergedFeatureFlags
                         if let flagsV4 {
                             self.setCachedFlags(flagsV4)
                         }
-                        self.setCachedFeatureFlags(featureFlags)
-                        self.setCachedFeatureFlagPayload(featureFlagPayloads)
+                        self.setCachedFeatureFlags(mergedFeatureFlags)
```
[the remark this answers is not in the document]
ioannisj: This one is by design. Bootstrapped-only keys get persisted into the flag cache the same way loaded flags do, so they survive a cold start until `reset()` clears them or the next `/flags` response overlays them. It mirrors posthog-js, where `bootstrap.featureFlags` get written into the persisted flags and served until the first `/flags` load, so a stale bootstrapped value on a later launch without the config is the intended offline-caching behavior, not a leak.

PostHog/PostHogRemoteConfig.swift
```diff
@@ -413,13 +429,19 @@ class PostHogRemoteConfig {
...
+                        self.setCachedFeatureFlagPayload(mergedFeatureFlagPayloads)
                     }
+
+                    self.setFlagsLoadedFromRemoteLocked()
                 }
 
                 self.notifyFeatureFlagsAndRelease(loadedFeatureFlags)
```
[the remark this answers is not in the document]
ioannisj: Good point. Fixed in b8023dc9d5: `setFlagsLoadedFromRemoteLocked()` now only runs on a complete load, not the partial `errorsWhileComputingFlags` path, so `$used_bootstrap_value` stays true while keys are still served from the bootstrap layer. A proper per-key version (checking whether each flag was in the last response) is a bigger change, worth a follow-up if we want that precision.

PostHog/PostHogRemoteConfig.swift
```diff
@@ -413,12 +430,21 @@ class PostHogRemoteConfig {
...
+                        let mergedFeatureFlags = self.bootstrappedFlags.merging(featureFlags) { _, loaded in loaded }
+                        let mergedFeatureFlagPayloads = self.bootstrappedPayloads.merging(featureFlagPayloads) { _, loaded in loaded }
+                        loadedFeatureFlags = mergedFeatureFlags
                         if let flagsV4 {
                             self.setCachedFlags(flagsV4)
                         }
-                        self.setCachedFeatureFlags(featureFlags)
-                        self.setCachedFeatureFlagPayload(featureFlagPayloads)
+                        self.setCachedFeatureFlags(mergedFeatureFlags)
+                        self.setCachedFeatureFlagPayload(mergedFeatureFlagPayloads)
```
dustinbyrne: deviation from posthog-js and android: a complete `/flags` response should replace existing flags and payloads completely, while only partial or errorsWhileComputingFlags (?) responses are merged

e.g.,

bootstrap flags:
```txt
beta-ui = "variant-a"
legacy = true
```

bootstrap payloads:
```txt
beta-ui = { "color": "blue" }
```

complete remote response:
```txt
beta-ui = "variant-b"
payloads = {}
```

With this merge, iOS returns:
```txt
beta-ui = "variant-b"
beta-ui payload = { "color": "blue" } // belongs to variant-a
legacy = true                         // absent from the complete response
```

posthog-js returns `variant-b`, no payload, and removes `legacy`.

PostHog/PostHogSDK.swift
```diff
@@ -243,6 +253,28 @@ let maxRetryDelay = 30.0
...
+        }
+
+        if storageManager.isIdentified() {
+            hedgeLog("Bootstrap distinctId differs from an already-identified user. The existing " +
+                "identity is preserved. Call reset() before reinitializing to switch users.")
+        } else {
+            // Existing anonymous user — merge it into the identified bootstrap ID.
+            identify(bootstrapId)
+        }
+    }
```
dustinbyrne, line 299: small deviation in behavior from posthog-js: 

given persisted state:
```txt
anonymousId = "user-123"
distinctId resolves to "user-123"
isIdentified = false
```

and bootstrap:
```txt
distinctId = "user-123"
isIdentifiedId = true
```

the guard returns because the distinct ids match, leaving the user anonymous.

posthog-js handles this by [keeping the same ID but marking the user as identified](https://github.com/PostHog/posthog-js/blob/c3749156170d93e6c30bf12ad61206e675305dfd/packages/browser/src/posthog-core.ts#L785-L794). no `identify()` is necessary because the distinct id did not change.

PostHog/PostHogSDK.swift
```diff
@@ -243,6 +253,28 @@ let maxRetryDelay = 30.0
...
+        else {
+            return
+        }
+
+        if storageManager.isIdentified() {
+            hedgeLog("Bootstrap distinctId differs from an already-identified user. The existing " +
+                "identity is preserved. Call reset() before reinitializing to switch users.")
+        } else {
+            // Existing anonymous user — merge it into the identified bootstrap ID.
+            identify(bootstrapId)
```
dustinbyrne, line 297: identify returns immediately if the user is opted out, so this scenario will never reconcile:

given persisted state:
```txt
anonymousId = "anon-123"
optOut = true
```

and bootstrap:
```txt
distinctId = "user-456"
isIdentifiedId = true
```

if they later opt in, this would be lost. posthog-js and the android  implementation both maintain local identity while opted out to avoid this.

PostHog/PostHogRemoteConfig.swift
```diff
@@ -413,12 +430,21 @@ class PostHogRemoteConfig {
...
+
+                        // Only a complete load flips the latch: a partial (errorsWhileComputingFlags)
+                        // response still serves some keys from the bootstrap layer, so
+                        // $used_bootstrap_value must stay true until a full /flags response overlays them.
+                        self.setFlagsLoadedFromRemoteLocked()
```
dustinbyrne: [P2] `$used_bootstrap_value` is a global “has this SDK received a successful remote response?” marker in posthog-js, not per-key provenance.

posthog-js sets `_flagsLoadedFromRemote = true` after any HTTP 200 (`packages/browser/src/posthog-featureflags.ts:666-678`), including a response with `errorsWhileComputingFlags=true`. Android currently matches that behavior. Here the marker is set only in the complete-response branch.

Example: bootstrap `legacy=true`, then receive HTTP 200 with `foo=true` and `errorsWhileComputingFlags=true`. The SDK preserves `legacy`, but a later `$feature_flag_called` reports `$used_bootstrap_value=false` in posthog-js/Android and `true` in iOS.

Please set the latch after any successfully received `/flags` HTTP 200 if this event property is intended to match posthog-js.

PostHog/PostHogRemoteConfig.swift
```diff
@@ -598,6 +624,54 @@ class PostHogRemoteConfig {
...
 
+    /// Seeds `config.bootstrap` feature flags and payloads into the served cache as a base
+    /// layer, so reads return them before the first `/flags` response. Existing cached values
+    /// (from a prior load) win over bootstrapped ones; bootstrapped-only keys are added.
+    private func seedBootstrapFlagsIfNeeded() {
+        featureFlagsLock.withLock {
+            guard !bootstrappedFlags.isEmpty else { return }
+
+            let existingFlags = getCachedFeatureFlags() ?? [:]
+            let mergedFlags = bootstrappedFlags.merging(existingFlags) { _, existing in existing }
```
dustinbyrne: [P2/design] Can we confirm that persisted cache intentionally wins over bootstrap?

For example, if the persisted cache from the previous launch contains `checkout-redesign=false` and a fresh caller-provided bootstrap contains `checkout-redesign=true`, iOS serves `false`. posthog-js applies bootstrap as a complete initial snapshot and serves `true`.

If this mobile-specific precedence is intentional, the public documentation should say so rather than saying this mirrors posthog-js. Otherwise the merge precedence should be reversed.

dustinbyrne (commented):
Compared the bootstrap implementation against posthog-js browser semantics. Leaving inline comments for the material behavioral differences and two mobile-specific design decisions that should be documented if intentional.

PostHog/PostHogRemoteConfig.swift
ioannisj: You're right, and I checked it against posthog-js. `receivedFeatureFlags` (`posthog-featureflags.ts:121-161`) replaces flags and payloads on a complete response, and only merges for `partialResponse` or `errorsWhileComputingFlags`. The spec had it backwards ("bootstrapped-only keys survive"), so I corrected the spec (PostHog/sdk-specs#12) and reworked iOS: a complete `/flags` now replaces the served flags and payloads, and partial/errored responses still merge. Fixed in bcafa132.

PostHog/PostHogSDK.swift
ioannisj: Confirmed. `posthog-core.ts:781-792` falls into the `else` branch and sets `USER_STATE_IDENTIFIED` when the bootstrap id matches the existing one, so iOS returning early on the `!=` guard was wrong. Fixed in bcafa132: reconcile now upgrades a matching anonymous id to identified, without re-linking or a redundant `$identify`. Added to the spec in PostHog/sdk-specs#12.

PostHog/PostHogSDK.swift
ioannisj: Right. posthog-js `identify()` (`posthog-core.ts:2511-2558`) persists the distinct id and identified state before any capture, with no opt-out early return, so local identity survives opt-out. iOS's `identify()` bails on `isOptOutState()` before persisting. Fixed in bcafa132: reconcile persists the merged identity directly when opted out, and the event stays suppressed. Spec updated in PostHog/sdk-specs#12.

PostHog/PostHogRemoteConfig.swift
ioannisj: Good catch, and this one reverses a change I made earlier in the PR. `posthog-featureflags.ts:647` sets `_flagsLoadedFromRemote = !errorsLoading` on any successful 200 (`errorsWhileComputingFlags` is tracked separately at `:663`), and `$used_bootstrap_value` is a global marker. So I reverted my complete-only change: the latch is set after any successful `/flags` response now. Spec tightened in PostHog/sdk-specs#12. bcafa132.

PostHog/PostHogRemoteConfig.swift
ioannisj: Confirmed, this is a real deviation. posthog-js applies bootstrap through a complete `receivedFeatureFlags` in `initialize()` (`posthog-featureflags.ts:316-338`), so bootstrap replaces the persisted cache (bootstrap wins) on every init, while iOS had cache-wins. Fixed in bcafa132: the seed applies bootstrap as a replace snapshot now. It was a genuine spec gap, so I added it to PostHog/sdk-specs#12 rather than treating it as iOS-only.

PostHog/PostHogRemoteConfig.swift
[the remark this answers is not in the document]
ioannisj: Correction to my earlier reply here. I was defending the base-layer/survive model, but checking against posthog-js showed a complete `/flags` response actually replaces the served flags, it doesn't keep bootstrapped-only keys. So this isn't by design. Reworked iOS in bcafa132 so a complete load replaces, and bootstrap doesn't linger in the durable cache past a complete load. Spec corrected in PostHog/sdk-specs#12.

PostHog/PostHogRemoteConfig.swift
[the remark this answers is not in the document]
ioannisj: Correcting my earlier reply. I'd changed the latch to complete-load-only, but posthog-js sets it after any successful 200 including `errorsWhileComputingFlags`, and `$used_bootstrap_value` is a global "received a remote response" marker rather than per-key. Reverted to match in bcafa132: the latch is set after any `/flags` response. Spec tightened in PostHog/sdk-specs#12.

ioannisj:
While reworking the flag and identity model to match posthog-js, I did a wider sweep of the js source and found three more differences beyond the five in the threads above.

### 1. Bootstrap now serves only enabled flags (fixed)

posthog-js filters bootstrap flags through `!!value` in `initialize()` (`posthog-featureflags.ts:318-337`), so a `false`/disabled flag gets dropped and payloads are kept only for the enabled flags. Its own tests pin this down (`posthog-core-also.test.ts:936-985`): `getFeatureFlag('disabled')` is `undefined` and `getFlagVariants()` leaves out the disabled keys. iOS was serving every value, `false` included.

Fixed in 2d71309d: `PostHogRemoteConfig.init` now keeps only the enabled bootstrap flags (a `true` boolean or a non-empty variant string) and the payloads for those keys. I tightened the spec to match in PostHog/sdk-specs#12, with an enabled-only requirement and a "disabled flag is not served" scenario.

### 2. Anonymous bootstrap over an existing anonymous id (kept iOS's non-overwrite, on purpose)

posthog-js's identity `else` branch runs `register({ distinct_id: bootstrapDistinctId })` (`posthog-core.ts:781-792`), which would overwrite an existing persisted anonymous id when an anonymous bootstrap is supplied. No js test covers that path though (the bootstrap-identity tests only cover fresh install, identified, matching id, first visit, and preserve-identified), so I read it as an untested edge rather than intended behavior.

iOS keeps the safer non-overwrite: `applyBootstrapIdentityIfNeeded` returns early when an anonymous id is already persisted, so a returning user's anon id never gets re-linked. I left this as a deliberate divergence rather than match an untested overwrite. Happy to switch to strict parity if we'd rather.

### 3. Bootstrap now fires the flags-loaded callback (fixed)

posthog-js fires `onFeatureFlags` the moment bootstrap is applied: `initialize()` calls `receivedFeatureFlags(...)` with the bootstrapped flags (`posthog-featureflags.ts:316-338`), which sets `_hasLoadedFlags = true` and runs the flag callbacks (`posthog-featureflags.ts:983-1005`). So a listener waiting on flags is unblocked straight from bootstrap, before any `/flags` response. iOS seeded the bootstrap snapshot at `PostHogRemoteConfig.init` but only fired `notifyFeatureFlags` after the remote round-trip, so listeners stayed blocked on the network even though the served values were already there.

Fixed in 4c6a0afd: `seedBootstrapFlagsIfNeeded` now reports whether it seeded, and init fires `notifyFeatureFlags` with the served flags when it does. Android had the same gap (bootstrap applied in `PostHogRemoteConfig`'s init, callback only fired after the network load), fixed in posthog-android c746d049: setup fires the flags-loaded callbacks when the snapshot was applied, before the reload. Spec tightened in PostHog/sdk-specs#12 with a requirement and a setup-time scenario.

PostHog/PostHogSDK.swift
```diff
@@ -243,6 +253,48 @@ let maxRetryDelay = 30.0
...
+                storageManager.setIdentified(true)
+            }
+            return
+        }
+
+        // Differing id.
+        if storageManager.isIdentified() {
+            hedgeLog("Bootstrap distinctId differs from an already-identified user. The existing " +
+                "identity is preserved. Call reset() before reinitializing to switch users.")
+        } else if isOptOutState() {
```
dustinbyrne, line 282: [P2] The opted-out reconciliation path bypasses the `personProfiles == .never` identity gate.

For the same returning anonymous user and identified bootstrap:

```
personProfiles = .never
anonymousId = "anon-123"
bootstrap identified ID = "user-456"
```

With opt-out disabled, this reaches `identify()`, which respects `.never` and leaves the SDK at `anon-123`. With opt-out
enabled, this branch directly writes `user-456` and marks the user identified.

Merely enabling opt-out therefore changes whether bootstrap identity is applied. posthog-js and Android route the
differing-identity case through the person-processing gate in both situations.

Consider preserving the `.never` decision while suppressing only event emission. A regression test could run the
scenario with `personProfiles = .never` and `optOut = true`, asserting the result matches the existing non-opted-out
`.never` test.

PostHog/PostHogConfig.swift
```diff
@@ -192,6 +192,24 @@ public typealias BeforeSendBlock = (PostHogEvent) -> PostHogEvent?
     /// - Returns: The UUID to persist as the anonymous ID.
     @objc public var getAnonymousId: ((UUID) -> UUID) = { uuid in uuid }
 
+    /// Pre-seeded identity and feature-flag state applied on the very first SDK launch
```
dustinbyrne: [P2/docs] This public documentation no longer matches the implementation in three places.

1. It says bootstrap is applied only on the first launch, but feature-flag bootstrap is applied on every initialization
   and wins over persisted flag cache.
2. It says bootstrap is ignored once an anonymous ID exists, but an identified bootstrap reconciles an existing
   anonymous user.
3. It says loaded values overlay bootstrap for overlapping keys, but a complete `/flags` response now replaces the
   entire snapshot and removes bootstrapped-only keys.

For example, a returning anonymous `anon-123` with an identified bootstrap `user-456` is reconciled, and a complete
response containing only `beta-ui` removes a bootstrapped `legacy` key.

Consider updating both this block and `PostHogBootstrapConfig.swift:3-42`: separate fresh-install identity seeding from
identified-bootstrap reconciliation for a returning anonymous user, then describe feature flags as a per-initialization
snapshot that replaces cache and is replaced by a complete response while partial/errored responses merge.

dustinbyrne (commented):
Follow-up review after the latest bootstrap changes.

dustinbyrne (approved):

PostHog/PostHogSDK.swift
ioannisj: Good catch, fixed in fac2cb55a: the opted-out branch now applies the same `personProfiles == .never` gate `identify()` does, so flipping opt-out no longer changes whether the bootstrap identity gets written. Added the regression test you suggested too (`.never` + optOut, asserting the same result as the non-opted-out `.never` case).

PostHog/PostHogConfig.swift
ioannisj: Fixed in b66b24d0e: reworked both the `PostHogConfig.bootstrap` doc and the `PostHogBootstrapConfig` class doc. Split identity seeding from identified-bootstrap reconciliation, and the flags are now described as a per-init snapshot that wins over the persisted cache, gets replaced by a complete `/flags` response, and merged by a partial/errored one.

PR #718, opened 2026-07-15
feat(flags): add $feature_flag_has_experiment to $feature_flag_called events

haacked:

Part of a cross-SDK effort to shrink `$feature_flag_called` events. This first phase adds a `$feature_flag_has_experiment` boolean property to every `$feature_flag_called` event, reflecting the server's `has_experiment` signal: `metadata.has_experiment` in the `/flags?v=2` response and `has_experiment` on `/api/feature_flag/local_evaluation` flag definitions. When the server does not report the field (older deployments, bootstrapped flags), the property is omitted, so it is tri-state: `true`, `false`, or absent (unknown).

This lets ingestion distinguish experiment-linked flag events (which need the full property set for exposure analysis) from the rest, and lets us measure the split before a later phase strips the expensive properties (`$feature/<key>`, `$feature_flag_payload`, per-event system metadata) from non-experiment flag events. This PR intentionally minimizes nothing: no properties are removed, no config options are added, and dedupe behavior is unchanged.

## Changes

- `reportFeatureFlagCalled` in `PostHogSDK.swift` (the single capture site for `$feature_flag_called`) now sets `$feature_flag_has_experiment` from the flag's `metadata.has_experiment`, defaulting to `false` (including v3 responses with no flag details). No parsing changes were needed; the SDK already caches the raw v2 flags payload.
- The SDK compliance adapter's hand-built `$feature_flag_called` also includes the property, read from the `/flags?v=2` JSON it already fetches.
- Patch changeset.

v4 mock-server fixture updated (`has_experiment: true` / `false` / absent per flag); assertions added to the two existing v4 `$feature_flag_called` tests plus a new absent-defaults-to-false test (which also covers v3 responses). Local caveat: the full test suite requires Xcode, which is not on this machine. `swift build` of the PostHog target and the adapter package succeed, changed files pass `swiftc -parse`, and `swiftformat --lint` is clean; CI should be the authoritative test run.

## :pencil: Checklist

## If releasing new changes

### DIFF
--- a/.changeset/feature-flag-has-experiment.md
+++ b/.changeset/feature-flag-has-experiment.md
@@ -1,0 +1,5 @@
+---
+"posthog-ios": patch
+---
+
+Add a `$feature_flag_has_experiment` boolean property to `$feature_flag_called` events, sourced from the flag's `metadata.has_experiment` in the flags response (false when the server does not report it).
--- a/.changeset/feature-flag-has-experiment.md
+++ b/.changeset/feature-flag-has-experiment.md
@@ -2,4 +2,4 @@
 "posthog-ios": patch
 ---
 
-Add a `$feature_flag_has_experiment` boolean property to `$feature_flag_called` events, sourced from the flag's `metadata.has_experiment` in the flags response (false when the server does not report it).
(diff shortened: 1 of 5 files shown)


turnipdabeets (approved):
LGTM! The two-layer default (seeded false, overwritten from cached v2 metadata with `as? Bool ?? false`) handles every response shape safely, and the fixtures cover true/explicit-false/absent.

PR #724, opened 2026-07-16
feat(flags): minimize $feature_flag_called events for non-experiment flags

haacked:

`$feature_flag_called` events carry the full enriched properties (context envelope, super properties, `$active_feature_flags`, the `$feature/<key>` enumeration) on every flag call, even for flags not linked to an experiment. This PR trims those events to a strict allowlist **iff** the server-controlled gate is on (top-level `minimalFlagCalledEvents` in the v2 `/flags` response) **and** the flag's `has_experiment` is exactly `false`. Any missing signal (legacy response, cached/bootstrap state without the field, `has_experiment` unknown) sends the full legacy shape unchanged. Server-gated because rolling this out unconditionally could break existing insights; announcement/comms come before enablement.

iOS specifics:

- The gate is persisted alongside the cached flags in storage (`posthog.minimalFlagCalledEvents`), so it survives app restarts; it is removed when a response omits the field and cleared on `reset()`.
- The allowlist is applied after `buildProperties()` builds the full property set, so the strip is structural — new context properties added later are stripped automatically. `beforeSend` hooks and the legacy `propertiesSanitizer` run after the filter and may re-add keys (accepted escape hatch, codified in the contract).
- Experiment-linked flags always send the full envelope — experiment exposure analysis reads it.
- `sendMinimalFlagCalledEvents`, `lastRequestId`, and `lastEvaluatedAt` all memoize their value in memory after the first disk read, so a flag check only touches storage once per process instead of on every call.

Part of a cross-SDK rollout; reference implementation and fuller context: PostHog/posthog-python#748

New test suites were authored but could **not** be executed in the authoring environment (no Xcode available — only `swift build -Xswiftc -DTESTING` type-checking and `swiftformat --lint` were run), so **CI is the validation gate for this PR**:

- `PostHogSDKTest` (Quick/Nimble): gated + no experiment → event's key set equals the exact expected minimal set; gated + registered group → `$groups` survives minimization and carries the group; gated + experiment → full event; gated + unknown `has_experiment` → full event; ungated → full event.
- `PostHogRemoteConfigTest` (swift-testing): gate defaults to false when absent; persists across a simulated restart (fresh instance re-reads it from storage); clears when a later response omits the field; `clear()` drops it.

Behavioral note: the v4 flag details (including `metadata.has_experiment`) persist with the cached flags, so minimization resumes from cache after a cold start once a gated response has been seen. Caches written before this change (or responses without flag details) lack `has_experiment`, so events stay full until the first fresh `/flags` response — fail-safe.

## :pencil: Checklist

## If releasing new changes

---

### DIFF
--- a/.changeset/minimal-flag-called-events.md
+++ b/.changeset/minimal-flag-called-events.md
@@ -2,4 +2,4 @@
 "posthog-ios": minor
 ---
 
-Send minimal `$feature_flag_called` events when the server opts the project in (top-level `minimalFlagCalledEvents` in the flags response) and the evaluated flag has no experiment. Minimal events keep only a strict allowlist of flag-evaluation and linkage properties; the device/OS context envelope, super properties, `$active_feature_flags`, and the `$feature/<key>` enumeration are stripped. Experiment-linked flags, ungated projects, and any response missing the signals keep sending the full event.
+Send minimal `$feature_flag_called` events when the server opts the project in (top-level `minimalFlagCalledEvents` in the flags response) and the evaluated flag has no experiment. Minimal events keep only a strict allowlist of flag-evaluation and linkage properties plus `$os_name`, `$os_version`, and `$app_version` for OS- and version-segmented insights; the rest of the device/OS context envelope, super properties, `$active_feature_flags`, and the `$feature/<key>` enumeration are stripped. Experiment-linked flags, ungated projects, and any response missing the signals keep sending the full event.
--- a/.changeset/minimal-flag-called-events.md
+++ b/.changeset/minimal-flag-called-events.md
@@ -1,0 +1,5 @@
+---
+"posthog-ios": minor
+---
+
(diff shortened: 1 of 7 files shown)


turnipdabeets (approved):
LGTM

Two non-blocking items:

Before the server gate ever turns on, since each SDK carved out a different debug exception: python keeps $os/$os_version/$python_runtime for platform breakdowns, browser js keeps $current_url/$pathname as debug location, but iOS (and flutter) keep neither analog — $os_name/$os_version/$app_version/$screen_name are all stripped. Once the gate goes live, insights that break down $feature_flag_called by OS or app version will silently lose mobile rows. Fine to ship as is, but I think we should either add the mobile analogs to the shared list or explicitly note the mobile divergence in the contract before enablement — worth settling once since it affects flutter too.

Also ReviewHog's memoization point is valid but minor (the read is deduped per flag and the pattern predates this PR via lastRequestId/lastEvaluatedAt). The suggested ReferenceWritableKeyPath fix is a nice 3-liner that fixes all three getters.

haacked:
Thanks @turnipdabeets. I addressed both of your suggestions.

Issue #335, opened 2025-04-17
Session Replay - Snapshot mode v2


ioannisj:
## Problem

The current snapshot mode implementation for Session Replay involves:
	1.	Capturing a full-window screenshot.
	2.	Traversing the view hierarchy to locate maskable elements and record their global coordinates and size.
	3.	Overlaying black masks on top of the captured screenshot based on these coordinates.

While effective in many cases, this approach has a few limitations:

 -	**Occlusion Ambiguity:** It’s difficult to determine whether a view is fully or partially visible due to the absence of a concrete z-index system. E.g., a modal may partially cover a view, but be semi-transparent.
-	**Transition Glitches:** UI transitions are problematic for snapshot consistency. Our current workaround disables snapshots during transitions, which can lead to gaps in the session replay.
-	**Syncing Issues:** Rapid UI changes (e.g., fast scrolling) can desync the base screenshot and the masking overlays. Attempting to resolve this will likely block the main thread even longer, which we want to avoid.

⸻

## Proposal: Snapshot v2 – View-by-View Compositing

A more reliable and flexible snapshot mode would:

- Traverse the full view hierarchy (as currently done to find maskable views anyway).
- For each (sub)view:
    - Capture a snapshot of the actual view or a masked counterpart if the view is sensitive (e.g Text *****)
    - Composite each snapshot into a shared context/canvas, positioning based on global coordinates.
- Output a single image representing the full window snapshot.

### Benefits

✅ Improved Visual Accuracy: Since views are drawn in hierarchy order, this will naturally account for occlusion, z-order, and partial visibility.
✅ No Sensitive Data Leaks: Masked views are never captured unmasked, preventing accidental exposure of text/images.
✅ Better Transition Handling: Layered drawing helps avoid the need to pause snapshotting during transitions.
✅ More Robust Scrolling Support: View-specific snapshots are less prone to visual desync.

### Risks & Considerations
- Performance: Needs profiling – compositing many views may prove to be slower or heavier on memory?
- Edge Cases: Views using custom rendering (e.g. OpenGL layers, blur effects) might not snapshot properly

### Tasks
- [x]	Create a proof-of-concept for compositing snapshots via the view hierarchy. 
- [ ] Benchmark performance across key devices (low-end, mid-range, high-end).
- [ ] Validate masking logic and ensure nothing leaks through in edge cases.
- [ ] Test against fast transitions and scroll-heavy views.

### Proof of concept
[...]

PR #728, opened 2026-07-17
fix: eliminate per-layout SwiftUI masking cost (iOS 26 launch/scroll hangs from per-row postHogMask)

### LINKED ISSUE #727: Session replay: per-row postHogMask() in SwiftUI lazy lists causes launch and scroll hangs on iOS 26
## Version

3.64.4 (analyzed in depth; the masking machinery is unchanged on `main` as of 3.66.1, so the issue applies there equally)

## Steps to Reproduce

1. iOS 26 device (Xcode 26 SwiftUI rendering engine), session replay enabled with `screenshotMode = true`.
2. A `ScrollView` + `LazyVStack` list with a few hundred rows where each row applies `postHogMask()` (in our production app: a bill tracker masking each row plus a nested category label — ~1,000 mask applications). A plain SwiftUI `List` is far less affected — the pathology needs all rows sharing one hosting view.
3. Cold-launch and programmatically scroll deep into the list (realizes hundreds of masked rows at once), then flick-scroll for ~10 s.

Minimal self-contained repro app (links posthog-ios as a local sibling checkout so any SDK revision can be profiled — setup in its README; no token needed): https://github.com/ibru/PostHogMaskingReproMinimal. Its default screen is exactly the pathological pattern; sibling screens (container mask, no masks, `List` variants) isolate the cost. A real-token recording of the repro app, for reference: https://us.posthog.com/shared/FGtto_gzyEGs4PYv8dQ-PtkToZq81Q?t=17

Note: the dominant cost reproduces **with a placeholder token and no recording** — the `postHogMask()` tag machinery runs regardless. The global `layoutSublayers` swizzle also installs with an invalid token because the surveys integration (`config.surveys` defaults to `true`) subscribes to the layout publisher unconditionally during `setup()`.

## Expected Result

Masking a few hundred rows should have negligible cost: masking metadata is only consumed at snapshot time (~1/s as a list of redaction rects), so nothing should need per-layout, per-mask hierarchy work.

## Actual Result

Multi-hundred-millisecond main-thread hangs at launch and sustained hitching while scrolling, with the cost concentrated in the SwiftUI tagging machinery.

**Measured in the repro app** (iPhone 12, iOS 26, Release, current `main` @ b0fa428; Time Profiler + Hangs): the Hangs instrument flags **repeated hangs throughout launch and scrolling**, and the heaviest stack shows **2.87 s (55.4%) of the sampled main-thread window under `PostHogFrameCaptureUIView.init(id:handler:)`**, with `getTargetViews(from:)` at 2.81 s (54.3%) and the `recursiveSequence`/`AnyIterator`/`LazyDropWhileSequence` iterator chains at ~2.68 s (51.6%):

Time Profiler + Hangs, repro app on iPhone 12, pre-fix

Full `.trace` files available on request.

**In the production app where we found this** (bill tracker, ~1,000 mask applications; separate traces, figures for corroboration): a 555 ms startup hang with ~97% under `PostHogTagUIView.setupAncestorObserver` → `getTargetViews`, and ≥1.4 s of a 10 s scroll capture under `PostHogFrameCaptureUIView.init` → `getTargetViews` iterator chains (~1.24 s inside `recursiveSequence`). Independent of masking count, every `layoutSublayers` call app-wide allocates a fresh `DispatchQueue` (`PostHogMulticastCallback.throttleQueue` is a computed property).

Recurring hot frames in all traces: `getTargetViews(from:)`, `closure #1 in recursiveSequence<A>(_:children:)`, `UIView.descendants.getter`, `specialized AnyIterator.next()`, `Array._conditionallyBridgeFromObjectiveC`, `swift_dynamicCast`.

### Why it degenerates on iOS 26

`postHogMask()` injects an anchor/tagger pair (plus a frame-capture overlay on iOS 26) and resolves the masked UIKit views by walking descendants of their nearest common **UIView** ancestor. On iOS 26, SwiftUI renders row content as CALayers, so the injected views' common ancestor is the hosting view — essentially the whole screen. Consequences, per masked row:

- every resolution enumerates the entire screen's hierarchy through lazy `recursiveSequence`/`AnyIterator` chains (large constant overhead per element);
- each mask KVO-observes the root layer's `sublayers`, so one row appearing during scroll re-fires **all N** masks' traversals — O(N × hierarchy) per scroll frame, effectively O(N²);
- the iOS 26 layer scan computes its reference frame from the *hosting view*, so every mask collects (and claims) content layers across the whole screen. A side effect of that over-collection: masks claim each other's layers, and since the mask flag is a plain Boolean, one mask's teardown (a lazy row scrolling offscreen) can clear flags on layers still owned by other masks — a latent privacy bug, not just a perf one.

The SDK's own doc comment on `postHogMask()` acknowledges the iOS 26 mapping limitation. Related: #335 (Snapshot mode v2) reworks the *capture* side — it doesn't address this tagging-side cost, but the fix direction we're proposing (masking decided at capture time from live state) is complementary to it; PR with fixes to follow.

ibru:

Fixes https://github.com/PostHog/posthog-ios/issues/727.

Per-row `postHogMask()` in SwiftUI lazy lists causes launch hangs and scroll hitching on iOS 26: masking metadata is computed at **layout frequency** (whole-screen traversals + root-scoped KVO per mask) but only consumed at **capture frequency** (~1/s, as redaction rects). On iOS 26 the injected tag views' common ancestor degenerates to the hosting view, making the work effectively O(N²) while scrolling. Traces in the linked issue; self-contained repro/benchmark app: https://github.com/ibru/PostHogMaskingReproMinimal.

Six stacked commits — the first five make the existing mechanism correct and affordable, the last replaces it for `postHogMask()` (details in each commit message):

1. `test(replay)`: characterization tests pinning current resolution behavior + DEBUG rect-introspection hooks (no behavior change).
2. `fix`: stop allocating a `DispatchQueue` per `layoutSublayers` call; skip idle dispatches entirely.
3. `fix`: iterative DFS traversal replaces the `recursiveSequence`/`AnyIterator` chains; `nearestCommonAncestor` O(depth²) → O(depth).
4. `fix`: flag ownership becomes ref-counted (one mask's teardown could unmask another's targets — latent privacy bug) and re-tagging coalesces to one drain per run-loop turn.
5. `fix`: iOS 26 layer scan bounded to the masked view's own extent, with clipped-subtree pruning.
6. `feat(replay)`: `postHogMask()` reports its region via a single passive overlay in a weak registry; the capture path reads live rects at snapshot time. Zero traversals/KVO/per-layout work, rects can never go stale, and it's OS-rendering agnostic — resolving the modifier's documented iOS 26 caveat. `postHogNoMask()`/`postHogNoRageClick()` stay on the tag-view path.

Relation to #335 (Snapshot mode v2): complementary — v2 reworks the capture side, this removes the tagging-side cost. The registry exposes live sensitive regions with no rendering decisions baked in, so a v2 compositor can consume it directly.

Three deliberate behavior changes, all failing closed: masks no longer collect content **outside their own extent** (previously over-collection); an explicit `postHogMask()` now wins inside a `postHogNoMask()` area; and in recordings a masked view is redacted across its **full extent** — previously on iOS 26 only the text/image layers the scan happened to discover inside it were redacted (an OS-specific artifact that could also under-redact whenever the layer mapping missed). Redaction now matches the API contract — what you mask is what's redacted — and is identical across OS versions.

**Known limitation**: full-extent redaction is coarser than before — a `postHogMask()` on a container (e.g. a list row) now blacks out the whole container in replays, not just the text/images inside it, which reduces replay fidelity for apps that masked containers. Workaround: apply `postHogMask()` to the sensitive subviews instead — the registry makes per-subview masks cheap, and redaction then hugs exactly those elements. A possible SDK-side follow-up is a snapshot-time pass that intersects recognized text/image layers with reporter extents and falls back to the full extent when nothing is recognized (fail-closed, unlike the old scan); and under #335's view-by-view compositor consuming this registry, content-shaped redaction inside masked regions comes back naturally, making this a transitional artifact.

**Remaining scope**: `postHogNoMask()` and `postHogNoRageClick()` intentionally stay on the tag-view traversal — they're subtractive, per-element flags (exempt *this* element from heuristics / rage-click detection), which an additive rect registry can't express without region subtraction. The traversal they ride is now iterative, extent-bounded, and coalesced to one drain per run-loop turn, but on iOS 26 it still resolves against the shared hosting view, so a screen unrealistically dense with these two modifiers would see a much smaller version of the original cost. If that ever matters in practice, there's a natural follow-up that finishes the job inside the current architecture: deduplicate the ancestor KVO observers (N observers on the same layer → one shared observer) and have the coalescer do a single shared walk per ancestor per drain, slicing each resolver's targets from one pass — O(subtree + N) per turn instead of O(N × subtree). Happy to do that as a separate PR if you want it.

## :shield: Regression safety

This machinery had no existing test coverage and protects user privacy, so the stack is built on proving behavior preservation:

- **Characterization first**: golden tests pinning current resolution semantics (DFS order, common-ancestor quirks, iOS 26 layer collection, flag consumption) landed before any change; every later edge-case pin was run green against the **old** implementation before the rewrite had to pass the identical suite.
- **Output-level verification**: a DEBUG hook exposes the production redaction-rect collection; the repro app dumps normalized rect lists with deterministic digests — byte-identical through commits 2–5. Commit 6 intentionally changes rect shapes (full mask extents), so there the proof is coverage: overlay screenshots per commit show equal-or-stronger redaction, zero rects on the unmasked baseline.
- **Fail-closed defaults**: first-attach resolution stays synchronous (fresh rows are flagged before they can be captured); the coalescer batches re-resolutions but never drops them; owner-set flags can only over-mask during teardown races; the weak registry self-heals a missed dismantle to "gone", never to "stale rect".
- **Bisectable stack**: each commit was verified before the next was built on it (targeted suites per commit, full suite at the riskiest points).

- Unit suites (Swift Testing): masking characterization (15), ownership & coalescing (7), mask registry (5), throttled-multicast gate regressions (3). **Full suite: 653 tests, green.**
- Repro app: DEBUG overlay + rect dumps on every commit, as above.
- Performance (iPhone 12, iOS 26, Release), Time Profiler + Hangs on the repro app's pathological screen (cold launch + sustained scrolling), before (upstream `main` @ `b0fa428`) vs. after (this branch):
  - **Before**: a dozen-plus hangs flagged across the session; heaviest stack has **2.87 s (55.4%) of the sampled main-thread window under `PostHogFrameCaptureUIView.init(id:handler:)`**, `getTargetViews(from:)` at 2.81 s (54.3%), `recursiveSequence`/`AnyIterator` iterator chains at ~2.68 s (51.6%).
  - **After**: **zero hangs recorded**; the heaviest stacks are SwiftUI's own layout (`AG::Graph::UpdateStack::update()` 850 ms, `LazyStack.place` 538 ms) — **no PostHog frame appears in the heaviest stack trace at all**. Screenshots: [before](https://raw.githubusercontent.com/ibru/PostHogMaskingReproMinimal/main/docs/repro-before.png) / [after](https://raw.githubusercontent.com/ibru/PostHogMaskingReproMinimal/main/docs/repro-after.png); full `.trace` files available on request.
- Real replay (real project token, iPhone 12, same app, `main` vs. this branch): recordings verified in the PostHog player. Before: only the discovered text layers render redacted ([recording](https://us.posthog.com/shared/FGtto_gzyEGs4PYv8dQ-PtkToZq81Q?t=17), [screenshot](https://raw.githubusercontent.com/ibru/PostHogMaskingReproMinimal/main/docs/replay-before-fix.png)); after: each masked view renders redacted across its full extent (the behavior change above) — everything redacted before remains redacted ([recording](https://us.posthog.com/shared/7uztRb6AnIlPZ7YYLT2ToAnubFYQCw?t=3), [screenshot](https://raw.githubusercontent.com/ibru/PostHogMaskingReproMinimal/main/docs/replay-after-fix.png)). Masked rows stay redacted throughout the session, including while scrolling.

## :pencil: Checklist

- [x] I updated the docs if needed. *(modifier doc comments updated — the iOS 26 caveat is resolved)*
- [ ] No breaking change or entry added to the changelog. *(the behavior notes above should get a changelog entry — happy to add in your preferred format)*

## If releasing new changes

- [ ] Ran `pnpm changeset` to generate a changeset file *(will add before merge)*

## 🤖 Agent context

Investigation, fix design, and implementation were directed by the PR author with a human in the loop throughout. Implementation was driven with **Claude Code (Fable)** (trace analysis, characterization tests, fix implementation, repro-app tooling); the analysis and fix plan were independently reviewed with **ChatGPT (Sol)**. Key decisions — characterization-first, ref-counted ownership as a prerequisite for coalescing, synchronous first-attach resolution, the capture-agnostic registry design composing with #335 — were made and reviewed by the author. All verification steps described above were actually run; profiling numbers were measured by the author.

### DIFF
--- a/PostHog/PostHogSDK.swift
+++ b/PostHog/PostHogSDK.swift
@@ -2948,7 +2948,8 @@ let maxRetryDelay = 30.0
 #endif
 
 #if os(iOS) && DEBUG
-    /// Debug-only session-replay introspection for verification harnesses.
+    /// Debug-only session-replay introspection for verification harnesses. SPI, not
+    /// part of the supported public API (mirrors `captureSessionReplaySnapshot`).
     public extension PostHogSDK {
         /// Debug-only: the redaction rects session replay would apply to `window` right
         /// now, computed through the production masking path (`findMaskableWidgets`,
@@ -2957,7 +2958,7 @@ let maxRetryDelay = 30.0
         /// Intended for verification harnesses that compare masking output across SDK
         /// changes. Works even when replay isn't recording (invalid token or replay
         /// flag off): a transient integration is used to read the local config.
-        func debugSessionReplayMaskableRects(in window: UIWindow) -> [CGRect] {
+        @_spi(PostHogInternal) func debugSessionReplayMaskableRects(in window: UIWindow) -> [CGRect] {
             guard isEnabled() else { return [] }
             let integration = replayIntegration ?? PostHogReplayIntegration.debugTransient(for: self)
             return integration.debugMaskableRects(in: window)
@@ -2966,7 +2967,7 @@ let maxRetryDelay = 30.0
         /// Objective-C-visible bridge for `debugSessionReplayMaskableRects(in:)`, so
         /// verification harnesses can discover the hook at runtime (`responds(to:)`)
         /// and keep compiling against SDK revisions that don't have it.
-        @objc(debugSessionReplayMaskableRectsIn:)
+        @_spi(PostHogInternal) @objc(debugSessionReplayMaskableRectsIn:)
         func debugSessionReplayMaskableRectValues(in window: UIWindow) -> [NSValue] {
             debugSessionReplayMaskableRects(in: window).map { NSValue(cgRect: $0) }
         }
--- a/PostHog/PostHogSDK.swift
+++ b/PostHog/PostHogSDK.swift
@@ -2947,32 +2947,4 @@ let maxRetryDelay = 30.0
     }
 #endif
 
-#if os(iOS) && DEBUG
-    /// Debug-only session-replay introspection for verification harnesses. SPI, not
-    /// part of the supported public API (mirrors `captureSessionReplaySnapshot`).
-    public extension PostHogSDK {
-        /// Debug-only: the redaction rects session replay would apply to `window` right
-        /// now, computed through the production masking path (`findMaskableWidgets`,
-        /// including the iOS 26 layer scan). Main thread only.
-        ///
-        /// For verification harnesses; works even when replay isn't recording.
-        /// Returns nil when no frame would be captured (a masked view isn't laid out
-        /// yet, or the SDK isn't set up) — distinct from an empty list (frame captured
-        /// with nothing masked). Main queue only; enforced also in Release builds.
-        @_spi(PostHogInternal) func debugSessionReplayMaskableRects(in window: UIWindow) -> [CGRect]? {
-            guard isEnabled() else { return nil }
-            let integration = replayIntegration ?? PostHogReplayIntegration.debugTransient(for: self)
-            return integration.debugMaskableRects(in: window)
-        }
-
-        /// Objective-C-visible bridge for `debugSessionReplayMaskableRects(in:)`, so
-        /// verification harnesses can discover the hook at runtime (`responds(to:)`)
-        /// and keep compiling against SDK revisions that don't have it.
-        @_spi(PostHogInternal) @objc(debugSessionReplayMaskableRectsIn:)
-        func debugSessionReplayMaskableRectValues(in window: UIWindow) -> [NSValue]? {
-            debugSessionReplayMaskableRects(in: window)?.map { NSValue(cgRect: $0) }
-        }
-    }
-#endif
-
 // swiftlint:enable file_length cyclomatic_complexity
--- a/PostHog/PostHogSDK.swift
+++ b/PostHog/PostHogSDK.swift
@@ -2947,4 +2947,30 @@ let maxRetryDelay = 30.0
     }
 #endif
 
+#if os(iOS) && DEBUG
+    /// Debug-only session-replay introspection for verification harnesses.
+    public extension PostHogSDK {
+        /// Debug-only: the redaction rects session replay would apply to `window` right
+        /// now, computed through the production masking path (`findMaskableWidgets`,
+        /// including the iOS 26 layer scan). Main thread only.
+        ///
+        /// Intended for verification harnesses that compare masking output across SDK
+        /// changes. Works even when replay isn't recording (invalid token or replay
+        /// flag off): a transient integration is used to read the local config.
+        func debugSessionReplayMaskableRects(in window: UIWindow) -> [CGRect] {
+            guard isEnabled() else { return [] }
+            let integration = replayIntegration ?? PostHogReplayIntegration.debugTransient(for: self)
+            return integration.debugMaskableRects(in: window)
+        }
+
+        /// Objective-C-visible bridge for `debugSessionReplayMaskableRects(in:)`, so
+        /// verification harnesses can discover the hook at runtime (`responds(to:)`)
+        /// and keep compiling against SDK revisions that don't have it.
+        @objc(debugSessionReplayMaskableRectsIn:)
+        func debugSessionReplayMaskableRectValues(in window: UIWindow) -> [NSValue] {
+            debugSessionReplayMaskableRects(in: window).map { NSValue(cgRect: $0) }
+        }
+    }
+#endif
+
 // swiftlint:enable file_length cyclomatic_complexity
--- a/PostHog/PostHogSDK.swift
+++ b/PostHog/PostHogSDK.swift
@@ -2955,11 +2955,12 @@ let maxRetryDelay = 30.0
         /// now, computed through the production masking path (`findMaskableWidgets`,
         /// including the iOS 26 layer scan). Main thread only.
         ///
-        /// Intended for verification harnesses that compare masking output across SDK
-        /// changes. Works even when replay isn't recording (invalid token or replay
-        /// flag off): a transient integration is used to read the local config.
-        @_spi(PostHogInternal) func debugSessionReplayMaskableRects(in window: UIWindow) -> [CGRect] {
-            guard isEnabled() else { return [] }
+        /// For verification harnesses; works even when replay isn't recording.
+        /// Returns nil when no frame would be captured (a masked view isn't laid out
+        /// yet, or the SDK isn't set up) — distinct from an empty list (frame captured
+        /// with nothing masked). Main queue only; enforced also in Release builds.
+        @_spi(PostHogInternal) func debugSessionReplayMaskableRects(in window: UIWindow) -> [CGRect]? {
+            guard isEnabled() else { return nil }
             let integration = replayIntegration ?? PostHogReplayIntegration.debugTransient(for: self)
             return integration.debugMaskableRects(in: window)
         }
@@ -2968,8 +2969,8 @@ let maxRetryDelay = 30.0
         /// verification harnesses can discover the hook at runtime (`responds(to:)`)
         /// and keep compiling against SDK revisions that don't have it.
         @_spi(PostHogInternal) @objc(debugSessionReplayMaskableRectsIn:)
-        func debugSessionReplayMaskableRectValues(in window: UIWindow) -> [NSValue] {
-            debugSessionReplayMaskableRects(in: window).map { NSValue(cgRect: $0) }
+        func debugSessionReplayMaskableRectValues(in window: UIWindow) -> [NSValue]? {
+            debugSessionReplayMaskableRects(in: window)?.map { NSValue(cgRect: $0) }
         }
     }
 #endif
--- a/PostHog/Replay/PostHogReplayIntegration.swift
+++ b/PostHog/Replay/PostHogReplayIntegration.swift
@@ -961,12 +961,6 @@
                 }
             }
 
-            // manually masked views through `.postHogMask()` view modifier
-            if view.postHogNoCapture {
-                maskableWidgets.append(view.toAbsoluteRect(window))
-                return
-            }
-
             // on RN, lots get converted to RCTRootContentView, RCTRootView, RCTView and sometimes its just the whole screen, we dont want to mask
             // in such cases
             if view.isNoCapture() || maskChildren {
@@ -996,19 +990,14 @@
         /// Recursively iterate through layer hierarchy to find maskable layers (iOS 26+)
         ///
         /// On iOS 26, SwiftUI primitives (Text, Image, Button) are rendered as CALayer sublayers
-        /// of parent views rather than having their own backing UIView. When `.postHogMask()` is applied,
-        /// the flag is set directly on the CALayers via the PostHogTagViewModifier.
+        /// of parent views rather than having their own backing UIView, so the text/image
+        /// heuristics below inspect layers as well. (`.postHogMask()` regions are collected
+        /// separately via the mask-reporter registry.)
         @available(iOS 26.0, *)
         private func findMaskableLayers(_ layer: CALayer, _ view: UIView, _ window: UIWindow, _ maskableWidgets: inout [CGRect]) {
             for sublayer in layer.sublayers ?? [] {
                 // Skip layers tagged with .postHogNoMask()
                 if sublayer.postHogNoMask {
-                    continue
-                }
-
-                // Check if layer is manually tagged with .postHogMask()
-                if sublayer.postHogNoCapture {
-                    maskableWidgets.append(sublayer.toAbsoluteRect(window))
                     continue
                 }
 
--- a/PostHog/Replay/PostHogReplayIntegration.swift
+++ b/PostHog/Replay/PostHogReplayIntegration.swift
@@ -1046,15 +1046,23 @@
                 return nil
             }
 
-            var maskableWidgets: [CGRect] = []
-            var maskChildren = false
-
-            findMaskableWidgets(window, window, &maskableWidgets, &maskChildren)
-
             let wireframe = createBasicWireframe(window)
-            wireframe.maskableWidgets = maskableWidgets
+            wireframe.maskableWidgets = collectMaskableRects(in: window)
             wireframe.type = "screenshot"
             return wireframe
+        }
+
+        /// All rects to redact in `window`: heuristic + flag-tagged widgets from the
+        /// hierarchy walk, plus the live rects of `postHogMask()` reporter views
+        /// (computed at capture time from the registry — never cached, so a masked
+        /// view that moved since the last layout callback is still redacted at its
+        /// current position).
+        func collectMaskableRects(in window: UIWindow) -> [CGRect] {
+            var maskableWidgets: [CGRect] = []
+            var maskChildren = false
+            findMaskableWidgets(window, window, &maskableWidgets, &maskChildren)
+            maskableWidgets.append(contentsOf: PostHogSessionReplayMaskRegistry.shared.maskedRects(in: window))
+            return maskableWidgets
         }
 
         // To be called from main thread
@@ -1615,16 +1623,13 @@
 
         extension PostHogReplayIntegration {
             /// Debug-only: computes the redaction rects that `snapshot()` would produce
-            /// for `window` right now, via the exact production `findMaskableWidgets`
-            /// path (including the iOS 26 layer scan). Main thread only.
+            /// for `window` right now, via the exact production collection path (the
+            /// `findMaskableWidgets` walk plus the mask-reporter registry). Main thread only.
             ///
             /// Used by verification harnesses to compare masking output before/after
             /// SDK changes — see the masking characterization tests.
             func debugMaskableRects(in window: UIWindow) -> [CGRect] {
-                var rects: [CGRect] = []
-                var maskChildren = false
-                findMaskableWidgets(window, window, &rects, &maskChildren)
-                return rects
+                collectMaskableRects(in: window)
             }
 
             /// Debug-only: a transient integration bound to `postHog` so that
--- a/PostHog/Replay/PostHogReplayIntegration.swift
+++ b/PostHog/Replay/PostHogReplayIntegration.swift
@@ -1610,6 +1610,34 @@
         }
     #endif
 
+    #if DEBUG
+        // MARK: - Debug verification hooks
+
+        extension PostHogReplayIntegration {
+            /// Debug-only: computes the redaction rects that `snapshot()` would produce
+            /// for `window` right now, via the exact production `findMaskableWidgets`
+            /// path (including the iOS 26 layer scan). Main thread only.
+            ///
+            /// Used by verification harnesses to compare masking output before/after
+            /// SDK changes — see the masking characterization tests.
+            func debugMaskableRects(in window: UIWindow) -> [CGRect] {
+                var rects: [CGRect] = []
+                var maskChildren = false
+                findMaskableWidgets(window, window, &rects, &maskChildren)
+                return rects
+            }
+
+            /// Debug-only: a transient integration bound to `postHog` so that
+            /// `debugMaskableRects(in:)` can read the session-replay config even when
+            /// replay isn't installed/recording (e.g. placeholder token, flag off).
+            static func debugTransient(for postHog: PostHogSDK) -> PostHogReplayIntegration {
+                let integration = PostHogReplayIntegration()
+                integration.postHog = postHog
+                return integration
+            }
+        }
+    #endif
+
 #endif
 
 // swiftlint:enable cyclomatic_complexity file_length
--- a/PostHog/Replay/PostHogReplayIntegration.swift
+++ b/PostHog/Replay/PostHogReplayIntegration.swift
@@ -1619,6 +1619,7 @@
     #endif
 
     #if DEBUG
+
         // MARK: - Debug verification hooks
 
         extension PostHogReplayIntegration {
--- a/PostHog/Replay/PostHogReplayIntegration.swift
+++ b/PostHog/Replay/PostHogReplayIntegration.swift
@@ -1623,30 +1623,6 @@
         }
     #endif
 
-    #if DEBUG
-
-        // MARK: - Debug verification hooks
-
-        extension PostHogReplayIntegration {
-            /// Debug-only: the redaction rects `snapshot()` would produce for `window`
-            /// right now, via the production collection path. Main thread only.
-            /// nil = production would skip this frame; [] = captured with nothing
-            /// redacted — harnesses must not conflate the two.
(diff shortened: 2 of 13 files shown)


ibru:
Hello guys, I discovered an issue #727 when using PostHog SDK in one of my apps.
So I thought I'll try to provide fix for it as well.
I tried to not break anything. There is little bit of change to original behavior which I discuss in the description text.
If this PR doesnt land well feel free to discard it or handle by your own.

PostHog/SwiftUI/PostHogMaskViewModifier.swift
```diff
@@ -40,32 +40,168 @@
...
+
     extension UIView {
+        /// Whether at least one mask currently claims this view.
         var postHogNoCapture: Bool {
-            get { objc_getAssociatedObject(self, &AssociatedKeys.phNoCapture) as? Bool ?? false }
-            set { objc_setAssociatedObject(self, &AssociatedKeys.phNoCapture, newValue, .OBJC_ASSOCIATION_RETAIN_NONATOMIC) }
+            isPostHogFlagOwned(&AssociatedKeys.phNoCapture)
+        }
+
+        func setPostHogNoCapture(_ enabled: Bool, owner: ObjectIdentifier) {
```
ioannisj: `setPostHogNoCapture(_:owner:)` is dead in production now and the two branches that read it can't fire: `if view.postHogNoCapture` at `PostHogReplayIntegration.swift:965` and `if sublayer.postHogNoCapture` at `:1010` both stay false forever.

PostHog/SwiftUI/PostHogTagViewModifier.swift
```diff
@@ -337,22 +378,53 @@
...
+        // ancestor, starting directly at the anchor (everything before it was only
+        // ever enumerated to be dropped) and terminating at the tagger. Replaces the
+        // recursiveSequence/AnyIterator lazy chains whose per-element overhead
+        // (closure boxes, protocol-witness dispatch, per-node subviews bridging)
+        // dominated launch and scroll traces.
+        //
+        // Seed the stack with the anchor plus the not-yet-visited right siblings
+        // along the path from the anchor up to the common ancestor — exactly the
+        // remainder of the flat pre-order sequence from the anchor onwards.
+        var siblingChains: [ArraySlice<UIView>] = []
```
ioannisj, line 407: Scope note rather than a defect: this traversal is the bit the fix doesn't fully kill. `postHogMask()` bypasses `getTargetViews` now via the registry, but `postHogNoMask()` and `postHogNoRageClick()` still ride it, and they can't move to the additive-rect registry because they're subtractive flags on specific backing views.

So the shared-hosting-view walk that caused the original hang still survives for those two on iOS 26. It's much cheaper now (bounded, coalesced, iterative) but still O(depth) per resolver per turn, so a screen with a few hundred `postHogNoMask()` rows could hit a smaller version of the same thing. Realistically those get used way less densely than `postHogMask()`, so I'm not worried about it.

Might be worth a sentence in the PR or changeset

ioannisj:
I still want to do a manual regression test on this just to make sure nothing else breaks. I'll hold off approval until i circle back on this soon. Till then some CI failures need to be addressed?

PostHog/SwiftUI/PostHogMaskViewModifier.swift
ibru: Thanks. fixed in a685833 (accessors + both read branches removed; `PostHogFlagOwners` stays since `postHogNoMask()`/`postHogNoRageClick()` still use it, and the ownership tests now run against `postHogNoMask`).

PostHog/SwiftUI/PostHogTagViewModifier.swift
ibru: Agreed — added a paragraph to the PR description saying exactly this. If the residual cost ever becomes real, there's a clean follow-up inside the current architecture: dedupe the ancestor KVO observers (one shared observer per ancestor instead of N) and let the drain do one shared walk per ancestor, slicing each resolver's targets from that single pass — O(subtree + N) per turn instead of O(N × subtree). Happy to open that as a separate PR.

ibru:
> I still want to do a manual regression test on this just to make sure nothing else breaks. I'll hold off approval until i circle back on this soon. Till then some CI failures need to be addressed?

Sure, no problem. CI should be sorted with the latest three commits:

- **SwiftFormat**: missing blank line around a MARK comment — fixed in `98e6b4d`; `swiftformat --lint` is clean repo-wide.
- **Public API snapshot**: the two DEBUG-only introspection hooks were showing up as new public symbols. Moved them behind `@_spi(PostHogInternal)` in `59dfd9c` (same convention as `captureSessionReplaySnapshot`), so the committed snapshot stays untouched — verified by running the `check-public-api.sh` extraction locally.
- **compliance-tests**: this one looks like a false alarm on the job level — the harness itself passed 45/45; the job then fails in its "post results as PR comment" step with `Resource not accessible by integration`, because `pull_request` runs from forks get a read-only `GITHUB_TOKEN` regardless of the workflow's `permissions:` block. So it'll likely keep "failing" for any fork PR until that step is guarded (e.g. skip the comment when `github.event.pull_request.head.repo.fork`).

The new commits are awaiting workflow approval whenever you have a moment.

For your manual regression pass: the repro app linked in the description switches SDK revisions with a plain `git checkout` and has a DEBUG overlay + deterministic rect-dump harness that reads the production `collectMaskableRects` output live — handy for eyeballing masking while scrolling. Shared replay recordings of the same screen before/after are linked in the description too. Happy to address anything the regression pass turns up.

ioannisj:
@ibru couple of things while I finish this up. Please add a changeset entry (see https://github.com/PostHog/posthog-ios/blob/9fc2443747c72960651d3808b5d2260e7c0a0411/RELEASING.md) and if you can please re-sign your commits?

PostHog/SwiftUI/PostHogMaskViewModifier.swift
```diff
@@ -40,32 +40,146 @@
...
+            let liveReporters = lock.withLock {
+                reporters = reporters.filter { $0.value.value != nil }
+                return reporters.values.compactMap(\.value)
+            }
+
+            var rects: [CGRect] = []
+            for reporter in liveReporters {
+                guard reporter.window === window, reporter.isVisible() else { continue }
+                let rect = reporter.toAbsoluteRect(window)
+                if !rect.isEmpty {
```
ioannisj, line 103: [suggestion] `maskedRects()` skips a reporter whose `toAbsoluteRect` is empty (`if !rect.isEmpty`). A reporter registered in `didMoveToWindow` before the first layout pass has a `.zero` frame, so a snapshot taken in that window silently drops its region and masks nothing for that view on that frame. Narrow, but it's fail-open. Worth confirming a pre-layout snapshot can't happen here, or falling back to the view's own bounds when the rect is empty?

PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -1610,6 +1607,32 @@
         }
     #endif
 
+    #if DEBUG
```
ioannisj: [suggestion] Let's move these behind `#if TESTING` 

Also, `debugMaskableRects` is just a thin alias for the internal `collectMaskableRects`,  which should be accessible through `@testable import` already?

PostHog/PostHogSDK.swift
```diff
@@ -2862,4 +2862,31 @@ let maxRetryDelay = 30.0
...
+    /// part of the supported public API (mirrors `captureSessionReplaySnapshot`).
+    public extension PostHogSDK {
+        /// Debug-only: the redaction rects session replay would apply to `window` right
+        /// now, computed through the production masking path (`findMaskableWidgets`,
+        /// including the iOS 26 layer scan). Main thread only.
+        ///
+        /// Intended for verification harnesses that compare masking output across SDK
+        /// changes. Works even when replay isn't recording (invalid token or replay
+        /// flag off): a transient integration is used to read the local config.
+        @_spi(PostHogInternal) func debugSessionReplayMaskableRects(in window: UIWindow) -> [CGRect] {
```
ioannisj: [suggestion] Did you find this useful to have or was it a convenience for during implementation? I lean towards dropping it since it will leak into host app debug builds

PostHog/SwiftUI/PostHogMaskViewModifier.swift
```diff
@@ -40,32 +40,146 @@
...
+        }
+    }
+
+    /// Ref-counted flag ownership. A target can be claimed by several masks at once —
+    /// overlapping target sets are the norm on iOS 26, where masks resolve against the
+    /// shared hosting view — so a plain Boolean flag would let one mask's teardown
+    /// (e.g. a lazy row scrolling offscreen) unmask targets still owned by another.
+    /// A target stays flagged for as long as at least one owner claims it.
+    final class PostHogFlagOwners {
+        var owners: Set<ObjectIdentifier> = []
```
ioannisj: Can we mirror what PostHogSessionReplayMaskRegistry already does for reporters here? hold the owner as a Weak ref rather than a bare ObjectIdentifier, and prune entries whose weak ref has gone nil on read. Then a reused address can't resolve to a dead owner.

ioannisj (commented):
Hey, tested this out manually just to make sure masking is not affected. Looks really good, thank you for this contribution! Could not identify any regressions other that one pointed out on whole row masking, but that feels that it does behave like it should and the workaround for people who don't want that behavior is straight forward. 

Left a couple of comments + changeset entry + signed comments and this should be good . 

Thank you for all the effort here!

PostHog/SwiftUI/PostHogMaskViewModifier.swift
ibru: Done in 6a7e229 — owners are now weak refs keyed by identity and pruned on read (mirroring the registry), typed `UIView` since every real owner is an injected tag view, with a dead-owner self-heal test covering it.

PostHog/PostHogSDK.swift
ibru: Yes, I agree. It was scaffolding for an external verification harness I used while developing the fix, and the characterization tests now pin the same behavior in-repo, so I've dropped it entirely in 02603c3 (both hooks, the ObjC bridge, and the transient-integration helper).

Happy I had a chance to learn about `@_spi` annotation. Didn't know it before.

PostHog/Replay/PostHogReplayIntegration.swift
ibru: It was indeed a thin wrapper and unnecessary. I've removed it as per other your comment. The tests hit `collectMaskableRects` directly via `@testable import`.

PostHog/SwiftUI/PostHogMaskViewModifier.swift
ibru: Good catch — you're right that it was fail-open, and it's concretely reachable with `screenshotModeBackgroundCapture`, where the pixels render after the rects are collected. The "fall back to bounds" idea unfortunately doesn't work: pre-layout `bounds` are zero too, so there's simply no geometry to use yet.

Fixed in 89cf2e2 by failing closed instead: the reporter tracks whether it has completed a layout pass, and if a registered one hasn't, the capture path skips that tick entirely (same pattern as the transition bail) rather than silently masking nothing. Three details to make that safe: a laid-out zero-size view doesn't stall capture (it's legitimately empty, not unsettled), unsettled reporters are marked layout-dirty on attach so nothing can stay unsettled forever, and the flag only resets on genuine cross-window moves so cell reuse during scrolling doesn't cost ticks.

Covered with tests, including a hosted-SwiftUI zero-size case — which actually caught a fail-dead variant of an earlier attempt at this fix, so it's already earned its keep.

ibru:
Both done — changeset added in `2790b0b`, and the whole branch is re-signed (also rebased onto latest `main`).

ibru:
@ioannisj Thanks for the review feedback. I did couple of smaller changes. So if the recent commits look good to you I guess it's worth quick smoke test to make sure, it's still not introducing any regressions. Will do that myself too.

PostHog/SwiftUI/PostHogTagViewModifier.swift
```diff
@@ -317,14 +332,40 @@
...
+    /// flag ownership on targets that dropped out of the resolution — without it, every
+    /// re-resolution (content replacement, hierarchy churn) would leak stale ownership
+    /// on views SwiftUI has recycled elsewhere.
+    func resolveTagTargets(
+        from taggerView: PostHogTagUIView,
+        coordinator: PostHogTagView.Coordinator,
+        onChange: PostHogTagHandler
+    ) {
+        let targets = getTargetViews(from: taggerView)
+        guard !targets.isEmpty else { return }
```
ioannisj: I think we should move this guard below, right before onChange() call. Thinking about recycled `UIView`s that keep a stale `postHogNoMask`/`postHogNoRageClick` claim from a prior owner here. If we don't release ownership, a recycled view now backing sensitive content gets treated as safe to record. Correct me if my thinking here is wrong please

PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -1046,17 +1035,41 @@
                 return nil
             }
 
-            var maskableWidgets: [CGRect] = []
-            var maskChildren = false
-
-            findMaskableWidgets(window, window, &maskableWidgets, &maskChildren)
+            // nil = a mask reporter has no geometry yet; capturing now could show that
+            // content unmasked. Skip the tick (fail closed), like the transition bail.
+            guard let maskableWidgets = collectMaskableRects(in: window) else {
```
ioannisj, line 1040: Nothing caps how long a reporter may hold the window unsettled here. If a reporter attaches, sets hasCompletedFirstLayout = false, and then never fires again, hasUnsettledReporters stays true forever and  `collectMaskableRects` keeps returning nil. Every snapshot for that window is dropped for the rest of the session? Code reads that this is plausible but I don't know how likely. Just to be on the safe side, should we  treat a reporter as settled after a reasonable timeout as a fallback? wdyt?

PostHog/SwiftUI/PostHogTagViewModifier.swift
```diff
@@ -174,10 +177,21 @@
...
+                let previous = coordinator.cachedLayers
+                let dropped = previous.filter { previousLayer in
+                    !layers.contains(where: { $0 === previousLayer })
                 }
+                if !dropped.isEmpty {
+                    coordinator.onRemoveHandler(owner, [], dropped)
+                }
+
+                coordinator.cachedLayers = layers
+                changeHandler(owner, [], layers)
```
ioannisj: [nit] I think we do something similar here in resolveTagTargets, so may be worth extracting into a helper function (something like func reconcileAndEmit())

ioannisj (commented):
Thank @ibru left a couple of minor comments, mainly on some edge cases that we may want to handle. I'll do another regression pass myself just to be safe. Apologies for the slow progress here but replay masking is important and I want us to nail this as much as possible before merging. Appreciate all the work so far here

ibru:
@ioannisj thanks for picking good areas. I'll work on them and try to deliver improvements today. After that we could do (hopefully :) ) final round of regressions.

PostHog/SwiftUI/PostHogTagViewModifier.swift
ibru: You're right — with the guard first, an empty re-resolution skipped the release and a live tagger kept stale claims on recycled views. Fixed in `4c50c17`: the reconciliation now always runs and only `onChange` is gated on non-empty. The same pattern existed in the iOS 26 layer handler, so both paths got the fix (via the shared helper from your other comment). The direction is safe fail-closed: releasing a stale no-mask claim can only increase masking.

PostHog/SwiftUI/PostHogTagViewModifier.swift
ibru: Done — both paths now go through a shared `reconcileAndEmit` in the same commit, so they can't drift apart again.

PostHog/Replay/PostHogReplayIntegration.swift
ibru: I'm not that familiar with this specific piece and too busy to dive deeply into this piece myself today, but both Fable and Sol agreed that the code can stay as is. Here is explanation from them. Does it sound reasonable to you?

> Permanent starvation is prevented by construction — the two mechanisms close the loop:
>
> - **An unsettled reporter is always layout-dirty.** When it attaches without a completed layout, `didMoveToWindow` calls `setNeedsLayout()`, and Core Animation lays out all dirty layers on every transaction commit (hidden subtrees included). So the reporter settles at the very next commit.
> - **Snapshots only fire *because of* layout activity.** `snapshot()` isn't on a timer — its only production trigger is the throttled `onViewLayout` subscription fed by the swizzled `UIView.layoutSublayers(of:)`.
>
> Put together, the scenario is self-contradicting:
>
> - If the window stops laying out → no layout events → **no snapshot ticks are generated at all** → there are no frames to skip.
> - If layout is happening → the commit that produced the tick also services the reporter's dirty flag, and the throttled callback runs on the main queue *after* that commit → by the time rects are read, the reporter is settled.
>
> So a skip lasts at most a tick in exotic interleavings, never the rest of the session — and every skip is visible via the `hedgeLog` line, so a pathological case in the wild would show up in debug logs rather than silently eating the session.
>
> A timeout would also be the wrong tool for the remaining risk: after it expires, the never-laid-out reporter still has a zero frame and contributes no rect, so the frame captures **without** the mask — the exact fail-open this fix removed, just delayed. Adversarial edge cases probed (hidden ancestors, inactive tab/nav branches, `screenshotModeBackgroundCapture` — which only moves pixel rendering off-main, adding no independent polling): all of them stay fail-closed (a frame may be dropped, never captured unmasked).

If you'd still like a safety net despite the above, the shape I'd suggest is a forced `window.layoutIfNeeded()` in the skip branch instead of a deadline — it deterministically settles the reporter on the same tick rather than guessing when to give up. Happy to add that if you prefer.

ioannisj: Agreed, let's keep this as-is. The timeout fallback was not a viable alternative anw

ioannisj (approved):
Went through some regression checks. All seems good here! Thank you!

PR #729, opened 2026-07-20
feat: add push notification support for PostHog Workflows

ioannisj:

Adds push notification support for PostHog Workflows, on by default and gated behind `enableSwizzling`:

- Registers the device's APNs token with PostHog (`POST /api/push_subscriptions`) so Workflows knows where to deliver. A token already delivered for the current distinct id isn't re-POSTed on every launch.
- Captures `$push_notification_opened` when a user taps a notification.
- On `reset()` (logout), unregisters the token for the old identity and re-registers it under the new anonymous id, so notifications stop following a logged-out user.

Each is independently opt-out via `capturePushNotificationSubscriptions` and `capturePushNotificationOpened`.

When the SDK isn't swizzling for you (`enableSwizzling` off, or a capability turned off), register, unregister, and capture by hand from your own delegate:

- `registerPushNotificationToken(_:)`, which uses the bundle identifier as the app id, or `registerPushNotificationToken(_:appId:)` to register a token.
- `unregisterPushNotificationToken()` to unregister the current token, for example from your own logout flow.
- `capturePushNotificationOpened(response:)` to record an open, or `capturePushNotificationOpened(title:subtitle:body:payload:action:)` when there's no `UNNotificationResponse` to hand over (matches the Android signature, so cross-platform layers can relay raw fields).

The `appId` overload also relays a non-APNs token (an FCM token under a Firebase `project_id`), so both direct-APNs and FCM delivery work; the backend routes on `app_id`, not platform.

Registration landed in PostHog/posthog#65807. Unregister sends `DELETE /api/push_subscriptions`, reusing the register `POST`'s path and five-field body so the backend can `$unset` the `$device_push_subscription_<app_id>` property. The backend `DELETE` route lands in PostHog/posthog#72490, which needs to merge first; until then the client ships ahead of it. An unregister is best-effort, and any non-2xx (including today's 405) is logged and ignored, so nothing breaks before it lands. SDK docs are tracked separately in PostHog/posthog.com#16177.

This PR also carries an unrelated one-target fix: `PostHogExample`'s runpath was missing `@executable_path/Frameworks`, so the example crashed on launch on a device under Xcode 16's debug-dylib builds.

`make test` covers the registration and unregister request shapes, retry and backoff with `Retry-After`, opt-out, re-registration when the distinct id changes, the skip when the same token is already delivered for the current distinct id, the `reset()` move (unregister the old identity, re-register under the new anonymous id, and skip the unregister when `reuseAnonymousId` keeps the same id), and every `$push_notification_opened` property vector, including the FCM path where the `posthog` attribution payload arrives as a JSON string rather than a dictionary. The unregister path runs against a mock PostHog server, since the backend `DELETE` isn't implemented yet.

I also ran the register-and-open flow end to end on a physical iPhone, against a live PostHog project and Workflow, on both providers:

- APNs direct: the token auto-registers via swizzling under the bundle id, and a Workflow-delivered push, once tapped, produced `$push_notification_opened` with the correct `$notification_title` and `$notification_body`, omitting subtitle and action when empty or default.
- FCM relay: the FCM token registers via `registerPushNotificationToken(_:appId:)` under the Firebase `project_id`, and a Workflow push delivered FCM to APNs, once tapped, produced the same spec-conformant event.

## :pencil: Checklist

## If releasing new changes

## 🤖 Agent context

**DRI:** @ioannisj


PostHog/PostHogSDK.swift
```diff
@@ -2824,6 +2884,143 @@ let maxRetryDelay = 30.0
...
+    /// rely on automatic swizzling (for example when `enableSwizzling` is `false`).
+    ///
+    /// The token is linked to the current distinct id, so it follows the user across `identify(_:)`.
+    ///
+    /// - Note: Registration is iOS-only in this version; calling this from another platform still
+    ///   sends a request labeled `platform: "ios"`.
+    ///
+    /// - Parameter deviceToken: The APNs token as a lowercase-hex string. Convert a `Data` token via
+    ///   `token.map { String(format: "%02x", $0) }.joined()`.
+    @objc public func handlePushNotificationDeviceToken(_ deviceToken: String) {
```
turnipdabeets: [suggestion] posthog-android#642 ships this as `registerPushNotificationToken(deviceToken, appId)`, and we already pair it with `unregisterPushNotificationToken()` here — could we rename to `registerPushNotificationToken` before this surface ships? Cross-SDK docs get much simpler with one verb, especially for AI.

PostHog/PostHogSDK.swift
```diff
@@ -2824,6 +2884,143 @@ let maxRetryDelay = 30.0
...
+    /// Use the `appId` overload when relaying a token obtained through another provider (for example a
+    /// Firebase `project_id`). When `appId` is `nil` the app's bundle identifier is used.
+    ///
+    /// - Note: Registration is iOS-only in this version; calling this from another platform still
+    ///   sends a request labeled `platform: "ios"`.
+    ///
+    /// - Parameters:
+    ///   - deviceToken: The push token string (APNs lowercase-hex, or an FCM token verbatim).
+    ///   - appId: The app identifier the token belongs to, or `nil` to use the bundle identifier.
+    @objc public func handlePushNotificationDeviceToken(_ deviceToken: String, appId: String?) {
```
turnipdabeets: [question] These token APIs are public on tvOS/watchOS/visionOS too (always sending `platform: "ios"`), while `capturePushNotificationOpened(response:)` is gated to iOS/macOS. Since registration is iOS-only in v1 and public API is hard to take back, should we gate these behind `#if os(iOS) || os(macOS)` until the backend accepts more platforms?

PostHog/PostHogSDK.swift
```diff
@@ -2824,6 +2884,143 @@ let maxRetryDelay = 30.0
...
+        /// Manually captures a `$push_notification_opened` event for a notification the user tapped.
+        ///
+        /// Use this when you're not relying on the automatic swizzling installed by
+        /// `capturePushNotificationOpened`, for example when `enableSwizzling` is `false` or when you
+        /// manage your own `UNUserNotificationCenterDelegate`. Call it from your
+        /// `userNotificationCenter(_:didReceive:withCompletionHandler:)` implementation.
+        ///
+        /// - Parameter response: The `UNNotificationResponse` received from the system.
+        @available(iOS 14.0, macOS 11.0, *)
+        @objc public func capturePushNotificationOpened(response: UNNotificationResponse) {
```
turnipdabeets, line 2959: [question] The swizzled `didReceive` fires for local notifications too, so a reminder-style `UNTimeIntervalNotificationTrigger` tap also emits `$push_notification_opened`. Android only auto-captures FCM remote taps — is including local notifications deliberate, or should we check `response.notification.request.trigger is UNPushNotificationTrigger`?

PostHog/PostHogSDK.swift
```diff
@@ -2824,6 +2884,143 @@ let maxRetryDelay = 30.0
...
+
+        pushSubscriptionHandler?.send(deviceToken: deviceToken, appId: appId)
+    }
+
+    /// Unregisters this device's push token from PostHog so Workflows stop targeting it — for example
+    /// from your logout flow.
+    ///
+    /// Sends a best-effort `DELETE /api/push_subscriptions` for the current distinct id (the backend
+    /// unsets the subscription property) and forgets the locally stored token. Unlike registration this
+    /// is not retried. It's called for you on `reset()` when `capturePushNotificationSubscriptions` is
```
turnipdabeets: [nit] reset()'s unregister is record-based, not flag-gated (the test at PostHogPushNotificationTest.swift:233 asserts a manually-registered token still moves — which seems right). This sentence promises flag gating, though; Android's `PostHogInterface` KDoc has the same sentence. Worth fixing in both.

PostHog/PostHogApi.swift
```diff
@@ -236,6 +236,71 @@ class PostHogApi {
...
+
+    private func sendPushSubscription(
+        httpMethod: String,
+        endpointName: String,
+        distinctId: String,
+        deviceToken: String,
+        appId: String,
+        completion: @escaping (PostHogUploadInfo) -> Void
+    ) {
+        guard let url = getEndpointURL("/api/push_subscriptions", relativeTo: config.host) else {
```
turnipdabeets: [question] Android #642 hits `/api/push_subscriptions/` (trailing slash) while we send `/api/push_subscriptions` — both presumably work today, but a POST/DELETE that hits a slash-redirect can drop the body on some stacks. Should the two SDKs align on whatever PostHog/posthog#65807 defines as canonical?

PostHog/PushNotifications/PostHogPushSubscriptionHandler.swift
```diff
@@ -0,0 +1,298 @@
...
+    }
+
+    /// Public-API unregister: DELETE for the current distinct id, then forget the local record so a
+    /// later launch won't re-send it.
+    func unregisterCurrentToken() {
+        guard let record = loadRecord() else {
+            hedgeLog("Push unregister skipped: no registered token.")
+            return
+        }
+        unregister(distinctId: distinctIdProvider(), deviceToken: record.deviceToken, appId: record.appId)
```
turnipdabeets, line 196: [nit] If the record was delivered for a previous distinct id and the resend hasn't landed yet, this DELETEs for the current id and the old id's subscription survives. Narrow window and the backend unsets dead tokens, but `record.deliveredForDistinctId ?? distinctIdProvider()` would close it.

PostHog/PushNotifications/PostHogPushNotificationSubscriptionIntegration.swift
```diff
@@ -0,0 +1,61 @@
...
+            uninstallIfNeeded(from: postHog, installedPostHog: self.postHog, state: Self.integrationInstallState) {
+                stop()
+                self.postHog = nil
+            }
+        }
+
+        func start() {
+            // Remote-notification registration needs a real app bundle; skip in test runners / CLI tools.
+            let bundleExtension = Bundle.main.bundleURL.pathExtension
+            guard bundleExtension == "app" || bundleExtension == "appex" else {
```
turnipdabeets: [question] Allowing `.appex` here means the publisher can hit `UIApplication.shared` in an extension process, where it's unavailable. Extensions can't register for remote notifications anyway — should this integration be app-only?

PostHog/PostHogSDK.swift
ioannisj: Right. I think the work to support all platform in the backend should be cheap enough. Checked the code and the only blocker is this [here](https://github.com/PostHog/posthog/blob/18e296b3ebceb0eb036f47d4dfba18d2eec71115/products/messaging/backend/api/push_subscriptions.py#L23). But I'd opt in on shipping on iOS first (since autocapture token is currently only developed and tested on that platform) and expand the public api later. I'll platform guard the apis

PostHog/PushNotifications/PostHogPushSubscriptionHandler.swift
ioannisj: This one's deliberate, it matches Android. `unregisterCurrent()` there also DELETEs for the current distinct id via `distinctIdProvider()`, not the delivered one. I'd rather keep the two aligned than close a narrow window the backend already handles by unsetting dead tokens on the next send. If we do want the `deliveredForDistinctId ?? distinctIdProvider()` fallback, could we do it in both SDKs together so they don't drift?

PostHog/PushNotifications/PostHogPushSubscriptionHandler.swift
```diff
@@ -0,0 +1,298 @@
...
+              let delivered = record.deliveredForDistinctId,
+              delivered != currentDistinctId
+        else {
+            return
+        }
+
+        resetRetryState()
+
+        attemptIfAllowed(deviceToken: record.deviceToken, appId: record.appId)
+    }
```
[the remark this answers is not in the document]
ioannisj: This is handled after the concurrency pass, an undelivered token isn't dropped on an identity change. `retryIfNeeded` reloads it and sends with the current distinct id, and a resend that races an in-flight send now gets coalesced (`pendingResend`) and replayed on completion. `resendIfDistinctIdChanged` only fires for already-delivered tokens on purpose: it re-targets a live subscription under the old id, whereas an undelivered one just sends fresh under the new id.

PostHog/PushNotifications/PostHogPushSubscriptionHandler.swift
```diff
@@ -0,0 +1,298 @@
...
+            return
+        }
+
+        // Delivered to a different user (identity changed while the app was closed) — treat as fresh.
+        if let delivered = record.deliveredForDistinctId, delivered != distinctId {
+            resetRetryState()
+        }
+
+        attemptIfAllowed(deviceToken: record.deviceToken, appId: record.appId)
+    }
```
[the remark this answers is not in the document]
ioannisj: By design, retries piggyback on `flush()` and the next launch rather than a dedicated timer. The record's persisted, so a backed-off attempt resumes on the next flush or relaunch. For a best-effort registration I didn't think a standalone retry scheduler earned the extra lifecycle surface in v1.

PostHog/PushNotifications/PostHogPushSubscriptionHandler.swift
```diff
@@ -0,0 +1,298 @@
...
+            return
+        }
+        api.deletePushSubscription(distinctId: distinctId, deviceToken: deviceToken, appId: appId) { info in
+            if let statusCode = info.statusCode, 200 ... 299 ~= statusCode {
+                hedgeLog("Push subscription unregistered successfully.")
+            } else {
+                hedgeLog("Push unregister failed (status \(info.statusCode.map(String.init) ?? "none")); ignoring (best-effort).")
+            }
+        }
+    }
```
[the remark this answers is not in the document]
ioannisj: Acknowledged, narrow window, and Android's reset has the same DELETE-then-reregister shape. The local record writes are serialized under a lock now, and if an in-flight old-id POST lands after the DELETE, the backend unsets dead tokens on the next send. Fully closing it would need the server to sequence the requests, so I'd rather track it as a follow-up than block v1.

turnipdabeets (approved):

PR #730, opened 2026-07-20
feat(surveys): add displaySurvey to show a survey on demand

### LINKED ISSUE #648: Support manual API displaySurvey
### Description

Web supports that already https://posthog.com/docs/surveys/implementing-custom-surveys
Mobile does not

- iOS (this issue)
- RN (https://github.com/PostHog/posthog-js/issues/3817)
- Flutter (https://github.com/PostHog/posthog-flutter/issues/225)

bs1180:

Closes #648

The web SDK exposes `posthog.displaySurvey()` to display a survey on demand (see [implementing custom surveys](https://posthog.com/docs/surveys/implementing-custom-surveys)), but there is no mobile equivalent, so surveys can only be shown when the automatic display conditions are met, and API-type surveys can't be shown at all. This adds the iOS counterpart:

```swift
PostHogSDK.shared.displaySurvey("survey-id")
```

- Looks the survey up in the loaded surveys and routes it through the existing display path (translations, active-survey tracking, `survey shown`/`sent`/`dismissed` events), so all downstream behavior is unchanged.
- Bypasses display conditions (targeting flags, event triggers, seen/wait-period checks), matching the web SDK's default behavior.
- Also enables API-type surveys, which `showNextSurvey()` intentionally never auto-displays.
- No-op (with a log) when the survey isn't found or another survey is already being displayed.

`showNextSurvey()` was refactored to share the new `presentSurvey(_:)` helper; its behavior is unchanged.

Sibling PRs: posthog-android and posthog-flutter (Flutter delegates survey logic to the native SDKs, so it needs this API first). RN is tracked in PostHog/posthog-js#3817.

Added `PostHogSurveyDisplaySurveyTest` covering: displaying an API-type survey by ID, bypassing an unfired event trigger, unknown ID as a no-op, and ignoring the call while another survey is active. Written to match the existing survey test harness, but not executed locally (authored in a Linux environment without an Apple toolchain) — relying on CI here.

## :pencil: Checklist

## If releasing new changes

## 🤖 Agent context

Authored with PostHog Code (Claude Code) from a request to bring the JS SDK's `displaySurvey` to the Flutter SDK; since Flutter's survey pipeline lives in the native SDKs, this iOS API is the prerequisite. Design choice: reuse the internal display path via a shared `presentSurvey(_:)` helper rather than a parallel code path, and keep v1 to a single `surveyId` parameter (no `ignoreConditions`/prefill options yet) to match the tracking issue's scope. Tests assert on integration state via the existing TESTING extension instead of rendering UI.

---

### DIFF
--- a/PostHog/Surveys/PostHogSurveyIntegration.swift
+++ b/PostHog/Surveys/PostHogSurveyIntegration.swift
@@ -295,24 +295,61 @@
                 // Check if there is a new popover surveys to be displayed
                 getActiveMatchingSurveys { activeSurveys in
                     if let survey = activeSurveys.first(where: self.canRenderSurvey) {
-                        let language = self.resolveDisplayLanguage()
-                        let translations = resolveSurveyTranslations(survey: survey, targetLanguage: language)
-                        self.setActiveSurvey(survey: survey, language: translations.matchedKey, questionTranslations: translations.questions)
+                        self.presentSurvey(survey)
+                    }
+                }
+            #endif
+        }
 
-                        DispatchQueue.main.async { [weak self] in
-                            if let self {
-                                // render the survey
-                                self.postHog?.config.surveysConfig.surveysDelegate.renderSurvey(
-                                    survey.toDisplaySurvey(
-                                        surveyTranslation: translations.survey,
-                                        questionTranslations: translations.questions
-                                    ),
-                                    onSurveyShown: self.handleSurveyShown,
-                                    onSurveyResponse: self.handleSurveyResponse,
-                                    onSurveyClosed: self.handleSurveyClosed
-                                )
-                            }
-                        }
+        /// Displays the survey with the given ID on demand.
+        ///
+        /// Unlike automatic display, this bypasses display conditions (targeting flags, event
+        /// triggers, and the seen/wait-period checks), so it also works for API-type surveys,
+        /// which are never auto-displayed. If another survey is already being displayed, the
+        /// call is ignored.
+        func displaySurvey(surveyId: String) {
+            #if os(iOS)
+                guard #available(iOS 15.0, *) else {
+                    hedgeLog("[Surveys] Surveys can be rendered only on iOS 15+")
+                    return
+                }
+            #endif
+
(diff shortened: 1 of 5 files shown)


bs1180:
Hi @ioannisj @turnipdabeets 👋 — tagging you both since you maintain the surveys code here. This adds the manual `displaySurvey` API tracked in #648, as the first piece needed for Flutter support (PostHog/posthog-flutter#225). No rush at all, and happy to adjust the API shape (e.g. add an `ignoreConditions` option or restrict allowed survey types) if you'd prefer something different. Thanks for taking a look!

PostHog/Surveys/PostHogSurveyIntegration.swift
```diff
@@ -295,24 +295,75 @@
...
+                    return
+                }
+                // Honor the documented "must be running" precondition. We deliberately bypass
+                // targeting flags, event triggers, and seen/wait-period checks, but a survey
+                // that has been stopped on the dashboard (endDate set) must not be shown.
+                guard survey.isActive else {
+                    hedgeLog("[Surveys] Cannot display survey \(surveyId) - survey is not running")
+                    return
+                }
+                self.presentSurvey(survey)
```
turnipdabeets, line 336: [suggestion] Bypassing `canRenderSurvey` is what enables `.api` — but it also lets `.unknown(type:)` through to a popover UI built for other types (`canRenderSurvey`'s own comment says popover-only for now). Could we allow-list `.popover`/`.api`/`.widget` and no-op with a log on unknown? Whether `.widget` should render as a popover here feels like a Team Surveys call, but failing closed on unknown seems right either way.

turnipdabeets (commented):
Two notes without a diff anchor: CI never ran on the head commit (only the Wiz checks fired) — it needs a re-trigger before merge; and the two ReviewHog threads have no reply even though both fixes are in 3a94107 — a one-liner on each would save reviewers re-deriving that.

marandaneto:
moving to draft until comments/conflicts are resolved

PR #734, opened 2026-07-27
fix: rage click interrupt

frederoni:
We noticed that popovers such as PencilKit’s PaperMarkupViewController’s color pickers were dismissed immediately.

bug

Bisected to [b9861d880](https://github.com/PostHog/posthog-ios/commit/b9861d880) feat: skip rage clicks on keyboard, text inputs, steppers

 Root cause: PostHogRageClickIntegration.handleApplicationEvent(_:) runs on every
 UIApplication.sendEvent(_:) via swizzling. That commit added a fallback hit-test:

 ```swift
   let hitView = touch.view ?? window.hitTest(touchCoordinates, with: event)
 ```

 Passing the live, in-flight UIEvent into hitTest(_:with:) while UIKit is still mid-delivery of
 that same event re-enters UIKit's hit-testing/gesture machinery and corrupts its internal touch
 state. For a presented markup dialog (PaperMarkupViewController, which is a popover/sheet),
 UIKit then treats the opening tap as an outside touch and dismisses it immediately.

 Fix: Perform a stateless hit-test by passing nil instead of the live event:

 ```swift
   let hitView = touch.view ?? window.hitTest(touchCoordinates, with: nil)
 ```

 This still recovers the tapped view for rage-click eligibility checks without perturbing
 UIKit's active event delivery. The event parameter is still used elsewhere in the method
 (event.type, event.allTouches), so nothing else changes.

### DIFF
--- a/.changeset/rage-click-interrupt.md
+++ b/.changeset/rage-click-interrupt.md
@@ -1,0 +1,5 @@
+---
+"posthog-ios": patch
+---
+
+Fix rage click detection dismissing presented UI (e.g. popovers and sheets) on the opening tap. The hit-test fallback used to recover a tapped view now runs as a stateless query (`hitTest` with `nil`) instead of re-entering UIKit's hit-testing/gesture machinery with the in-flight event, which corrupted internal touch/gesture state.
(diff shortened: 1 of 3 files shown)


ioannisj:
@frederoni this looks really good. Missing a changeset entry though and I think it should be good to go. Thank you for this!

frederoni:
Thanks @ioannisj I haven't used changesets before but I think I got it right. Let me know if there is anything else.

ioannisj (approved):

ioannisj:
@frederoni one last thing before I can merge this. Can you sign your commit please (branch protection rules require that) 🙏

PR #735, opened 2026-07-27
feat: attach identity verification token to push subscription requests

ioannisj:

[PostHog/posthog#72488](https://github.com/PostHog/posthog/pull/72488) (draft) adds opt-in identity verification to `/api/push_subscriptions`: the backend mints a short-lived HS256 JWT (`sub` = distinct_id, `app_id`, `aud` = `posthog:push_identity`, `exp`) signed with the project's secret API key, and verifies it on POST and DELETE. Integration modes are `disabled` (default), `optional`, `required`. Without SDK support there is no way to attach the token, so `required` mode is unusable from iOS. The provider shape was agreed with Dan in #team-messaging.

Adds the SDK half:

- New `config.pushIdentityProvider: (distinctId, appId, completion) -> Void`. The app hands the backend-minted token (or nil) to `completion`, from any thread, exactly once.
- The token goes out as `identity_token` on the register POST and unregister DELETE. When the key is absent it is omitted entirely, so the body stays byte-identical to today's contract for apps that don't opt in.
- Minting is on demand and cached in memory per exact `(distinctId, appId)`: reused across in-session backoff retries, re-minted on identity change, refreshed on 401, and never cached once minted after opt-out.

`make test` green (707 tests), including cross-SDK vectors 9-14 (token on both legs, key omission when absent, reset-leg token counts, 500-retry cache reuse, 401 refresh then terminal, 401 with no provider) and exactly-once foreign-thread completion tests. `make apiCheck` and `make lint` clean.

## :pencil: Checklist

- [x] No breaking change entry to add to the changelog.

## If releasing new changes

- [ ] Ran `pnpm changeset` to generate the changeset file (deliberately omitted: feature branch, not releasing yet)

## 🤖 Agent context

**DRI:** @ioannisj

### DIFF
--- a/PostHog/PushNotifications/PostHogPushSubscriptionHandler.swift
+++ b/PostHog/PushNotifications/PostHogPushSubscriptionHandler.swift
@@ -222,9 +222,13 @@ final class PostHogPushSubscriptionHandler {
         unregister(distinctId: distinctIdProvider(), deviceToken: record.deviceToken, appId: record.appId)
     }
 
-    /// Opt-out drops the cached identity credential so a later opt-in re-mints it.
+    /// Opt-out drops the cached identity credential so a later opt-in re-mints it, and clears the
+    /// per-cycle 401 retry flag so a consumed retry doesn't stay stuck and suppress the next one.
     func onOptOut() {
-        stateLock.withLock { cachedIdentityToken = nil }
+        stateLock.withLock {
+            cachedIdentityToken = nil
+            didAuthRetry = false
+        }
     }
 
     // MARK: - Private
@@ -292,7 +296,13 @@ final class PostHogPushSubscriptionHandler {
             guard let self else { return }
             // Opt-out may land during a slow mint; re-check before sending.
             guard isAllowedProvider() else {
-                stateLock.withLock { self.isSending = false }
+                // Also drop any registration that folded into `pendingResend` during the mint. We're
+                // opted out, so it must not fire, and leaving it set would make the next send's
+                // `handleResult` service a stale resend. A later opt-in re-registers normally.
+                stateLock.withLock {
+                    self.isSending = false
+                    self.pendingResend = false
+                }
                 return
             }
             api.pushSubscription(
@@ -343,15 +353,16 @@ final class PostHogPushSubscriptionHandler {
         }
         provider(distinctId, appId) { [weak self] token in
             guard let self else { return }
-            // A mint can complete after opt-out cleared the cache; caching it would resurrect a
-            // stale credential on a later opt-in. The 401 refresh covers the residual race window.
-            let allowed = isAllowedProvider()
             let isFirst = self.stateLock.withLock { () -> Bool in
                 if completed {
                     return false
                 }
                 completed = true
-                if let token, allowed {
+                // Sample opt-out under the same lock onOptOut() clears the cache with, so an opt-out
+                // can't land between the check and the write and leave a stale token cached. If opt-out
+                // wins, isAllowedProvider() is false here and nothing is cached; if we win, onOptOut()
+                // clears our write right after we release the lock.
+                if let token, self.isAllowedProvider() {
                     self.cachedIdentityToken = CachedIdentityToken(token: token, distinctId: distinctId, appId: appId)
                 }
                 return true
--- a/PostHog/PushNotifications/PostHogPushSubscriptionHandler.swift
+++ b/PostHog/PushNotifications/PostHogPushSubscriptionHandler.swift
@@ -320,12 +320,15 @@ final class PostHogPushSubscriptionHandler {
         var completed = false
         provider(distinctId, appId) { [weak self] token in
             guard let self else { return }
+            // A mint can complete after opt-out cleared the cache; caching it would resurrect a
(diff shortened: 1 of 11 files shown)


PostHog/PushNotifications/PostHogPushSubscriptionHandler.swift
```diff
@@ -254,8 +280,61 @@ final class PostHogPushSubscriptionHandler {
...
+        provider(distinctId, appId) { [weak self] token in
+            guard let self else { return }
+            let isFirst = self.stateLock.withLock { () -> Bool in
+                if completed {
+                    return false
+                }
+                completed = true
+                if let token {
+                    self.cachedIdentityToken = CachedIdentityToken(token: token, distinctId: distinctId, appId: appId)
+                }
```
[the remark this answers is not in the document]
ioannisj: Good catch. Fixed in e1d46af63c, the completion now skips the cache write when opt-out landed mid-mint, so a later opt-in re-mints instead of reusing the stale token. Same race existed on the Android side, fixed there too (PostHog/posthog-android#656).

PostHog/PushNotifications/PostHogPushSubscriptionHandler.swift
```diff
@@ -254,8 +286,81 @@ final class PostHogPushSubscriptionHandler {
...
+            // A mint can complete after opt-out cleared the cache; caching it would resurrect a
+            // stale credential on a later opt-in. The 401 refresh covers the residual race window.
+            let allowed = isAllowedProvider()
+            let isFirst = self.stateLock.withLock { () -> Bool in
+                if completed {
+                    return false
+                }
+                completed = true
+                if let token, allowed {
+                    self.cachedIdentityToken = CachedIdentityToken(token: token, distinctId: distinctId, appId: appId)
```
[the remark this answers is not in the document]
ioannisj: Good catch, fixed in 1e9522a46c. The `isAllowedProvider()` check now happens inside the same `stateLock` that `onOptOut()` clears the cache under, so opt-out can't slip in between the check and the write: either it wins and nothing caches, or the mint wins and `onOptOut()` clears the write right after. The 401 refresh isn't the only thing bounding it anymore.

Two neighbouring opt-out gaps went with it in the same commit: the mid-mint abort now drains `pendingResend` (it was left stranded, so the next send's `handleResult` would fire a spurious duplicate), and `onOptOut()` clears `didAuthRetry` so a consumed 401 retry doesn't stay stuck and suppress the next one.

PostHog/PushNotifications/PostHogPushSubscriptionHandler.swift
```diff
@@ -145,11 +170,18 @@ final class PostHogPushSubscriptionHandler {
             hedgeLog("Push unregister skipped: missing distinct id, token, or app id.")
             return
         }
-        api.deletePushSubscription(distinctId: distinctId, deviceToken: deviceToken, appId: appId) { info in
-            if let statusCode = info.statusCode, 200 ... 299 ~= statusCode {
-                hedgeLog("Push subscription unregistered successfully.")
-            } else {
-                hedgeLog("Push unregister failed (status \(info.statusCode.map(String.init) ?? "none")); ignoring (best-effort).")
+        // Same one-verification-state as registration: the DELETE resolves a token for the distinct id
+        // it carries (the old identity on the reset() path). Still best-effort — no 401 refresh here.
```
marandaneto: **[P1] Refresh expired tokens before giving up on unregister**

`resolveIdentityToken` prefers an in-memory token with no expiry check, while the backend mints these with a 5-minute default TTL. A normal logout/reset later in the same process can therefore send an expired token here; required mode returns 401, this path deliberately does not refresh, and `unregisterCurrentToken()` has already deleted the local record. The old person's server-side subscription is then left active, so this device can keep receiving their notifications after logout. Please invalidate/remint and retry the DELETE once on 401 (or always mint fresh for DELETE) before dropping this best-effort operation.

marandaneto (dismissed):
The registration path looks well covered, but the unregister path can leave a user's subscription active once the cached JWT expires. Also, the current head has red test checks (`test` and `test-ios-simulator`); the simulator job specifically reports the newly added vector-11 reset test failing.

PostHog/PushNotifications/PostHogPushSubscriptionHandler.swift
marandaneto: The same data-loss path occurs when the device is offline: unlike registration, unregister does not check connectivity or persist a pending operation, and `unregisterCurrentToken()` removes the local record before attempting the DELETE. A logout/reset while offline therefore gets one transport failure, is never retried on `flush()` or next launch, and leaves the device associated with the old user on the backend. We should persist a separate pending-unregister intent and retry it later (or otherwise retain enough state to retry) rather than dropping it.

ioannisj: Both fixed in f1bd05e. Unregister now persists a pending intent (`PendingUnregister`) before the DELETE, so an offline or failed attempt gets retried on `flush()` and next launch instead of dropped. And on a 401 it clears the cached token and re-mints once before giving up, same as the send path.

Need to close the same gap on Android, will do on upcoming release PRs

marandaneto (approved):

PR #739, opened 2026-07-29
test: golden-image and geometric regression suite for session replay masking

ioannisj:

Session replay masking has no rendered-pixel regression coverage. #728 reworked mask collection and we only caught its regressions by eyeballing the example apps. This adds a CI safety net that fails when masked output changes in either direction: a secret leaking (under-masking) or a mask growing to cover the whole screen (over-masking).

Three layers:

- `PostHogMaskScenarioTest`: hosts SwiftUI views, collects redaction rects via `collectMaskableRects(in:)`, and asserts the privacy property geometrically (leaf masks hug the label, container masks cover their extent, rects track scrolling). Runs in the normal iOS suite.
- `PostHogMaskSnapshotTest`: 24 golden-image scenarios rendered as `[caption | ORIGINAL | MASKED]` composites and compared pixel-wise against committed goldens. Covers the heuristic config surface (`maskAllTextInputs`, `maskAllImages`), explicit `postHogMask`/`postHogNoMask`, SecureField fail-closed behavior, SF Symbol vs raster image handling, and placeholder-only text fields. Each caption embeds the live config and the expectation, so a failure heatmap is self-explanatory.
- A `mask-snapshots` CI job that runs `make maskSnapshots` on the pinned iOS 26.2 simulator runtime; on failure the actual + diff heatmap PNGs upload as a `mask-snapshot-diffs` artifact.

A few design decisions worth knowing about:

- The harness forces every device-shaped render input (light appearance, no safe area, `displayScale` 3, fixed content size, locale, layout direction, idiom, gamut), so any iPhone on the pinned OS runtime renders byte-identical output. I verified this by re-recording on a different device model. The OS version is the only pin; when the CI runner image rotates, `checkMaskSnapshotRuntime` fails with instructions to bump `MASK_SNAPSHOT_OS` and re-record.
- The snapshot suite only compiles under `TEST_MASK_SNAPSHOTS` (passed by the `maskSnapshots`/`recordMaskSnapshots` make targets), so Xcode ⌘U and `make testOniOSSimulator` never run it and OS drift on a dev machine can't fail unrelated work. Re-recording is `make recordMaskSnapshots` (adds `RECORD_MASK_SNAPSHOTS`), then review and commit the PNGs.
- Goldens are deterministic, so the diff threshold is tight (0.1% of pixels, box-downsampled to absorb antialiasing): small enough that a single leaf mask regressing on one row is caught. `verifyGoldens` also fails on orphaned goldens so removed scenarios can't leave stale PNGs behind.

Only SwiftUI is covered; UIKit golden scenarios are worth a follow-up.

This PR is tests. Verified locally on the iOS 26.2 simulator:

- `make maskSnapshots`: snapshot suite green against the committed goldens. I confirmed via the raw xcodebuild output that the test actually ran, since the compile-flag gating makes a 0-test pass look green in xcpretty.
- `xcodebuild test -only-testing:PostHogTests/PostHogMaskScenarioTest`: 9 geometric tests green.
- Cases 1 to 21 re-recorded on a different iPhone model produced byte-identical PNGs.
- Each new scenario was rendered first and its caption rewritten to the observed masking truth (two initial predictions were wrong; the captions document what actually happens).

## :pencil: Checklist

## If releasing new changes

## 🤖 Agent context

**DRI:** @ioannisj


PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -977,6 +977,8 @@
 
             if !view.subviews.isEmpty {
                 for child in view.subviews {
+                    // INTENTIONAL BUG (mutation test, remove this line): first subview at every level is never traversed
+                    if child === view.subviews.first { continue }
```
ioannisj: Just demo, should be removed before merge

marandaneto (approved):
nice, approving to unblock, ci is failing tho

.github/workflows/test.yml
```diff
@@ -87,6 +87,71 @@ jobs:
...
+            **[Download the diff artifact](https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}/artifacts/${{ steps.diff-upload.outputs.artifact-id }})** — per failing case: `expected` (committed golden), `actual` (this run), `diff` (heatmap of changed pixels).
+
+            If the change is intended, refresh with `make recordMaskSnapshots` and commit the updated PNGs.
+          edit-mode: replace
+      # Goldens match again: drop the stale divergence comment so a green PR reads clean.
+      - name: Clear mask-diff comment on success
+        if: ${{ success() && needs.detect-markdown-only.outputs.markdown_only != 'true' && github.event.pull_request.head.repo.full_name == github.repository && steps.find-mask-comment.outputs.comment-id != '' }}
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: gh api -X DELETE "repos/${{ github.repository }}/issues/comments/${{ steps.find-mask-comment.outputs.comment-id }}"
```
marandaneto: blocking: This direct interpolation is what the failing Semgrep check flags. Please pass `steps.find-mask-comment.outputs.comment-id` through an environment variable and quote the shell variable in `gh api`.

Makefile
```diff
@@ -108,6 +108,35 @@ testOniOSSimulator:
...
+	  exit 1; }
+	@[ -n "$(MASK_SNAPSHOT_UDID)" ] || { \
+	  echo "error: no available iPhone simulator on the iOS $(MASK_SNAPSHOT_OS) runtime; create one via Xcode or 'xcrun simctl create'."; \
+	  exit 1; }
+
+maskSnapshots: checkMaskSnapshotRuntime
+	set -o pipefail && xcrun xcodebuild test -scheme PostHog \
+	  -destination "platform=iOS Simulator,id=$(MASK_SNAPSHOT_UDID)" \
+	  SWIFT_ACTIVE_COMPILATION_CONDITIONS='$$(inherited) TEST_MASK_SNAPSHOTS' \
+	  "-only-testing:PostHogTests/PostHogMaskSnapshotTest/verifyGoldens()" | xcpretty
```
marandaneto: suggestion: Can we make this target assert that `verifyGoldens` actually executed? The current Actions log says `Executed 0 tests` even though the generated diff artifacts prove Swift Testing ran. If the selector or compilation gate breaks later, a real zero-test run could look green. TEEing the raw log or retaining/parsing the `.xcresult` for an execution marker would close that gap.

.github/workflows/test.yml
```diff
@@ -87,6 +87,71 @@ jobs:
...
+    steps:
+      - name: Complete markdown-only PR check
+        if: needs.detect-markdown-only.outputs.markdown_only == 'true'
+        run: echo "Only Markdown files changed; no additional work is required for this check."
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        if: needs.detect-markdown-only.outputs.markdown_only != 'true'
+      - uses: maxim-lobanov/setup-xcode@ed7a3b1fda3918c0306d1b724322adc0b8cc0a90 # v1.7.0
+        if: needs.detect-markdown-only.outputs.markdown_only != 'true'
+        with:
+          xcode-version: latest-stable
```
marandaneto: suggestion: Since these goldens are byte-sensitive, should Xcode be pinned alongside the iOS 26.2 runtime? `latest-stable` can change the compiler/SDK while the simulator OS remains fixed, which makes the comment that the OS is the only pin stronger than the workflow guarantees.

PostHogTests/PostHogMaskSnapshotTest.swift
```diff
@@ -0,0 +1,647 @@
...
+            }
+
+            private func paintingBlack(_ rects: [CGRect], over image: UIImage, in bounds: CGRect) -> UIImage {
+                let format = UIGraphicsImageRendererFormat()
+                format.scale = 1
+                return UIGraphicsImageRenderer(bounds: bounds, format: format).image { ctx in
+                    image.draw(in: bounds)
+                    UIColor.black.setFill()
+                    for rect in rects {
+                        ctx.fill(rect)
```
marandaneto: suggestion: This validates rectangle discovery by painting square blocks, but production uses `RRWireframe.maskImage()` with rounded paths and then base64/WebP encoding. Could we reuse/extract the production masking helper here, or narrow the suite documentation to rectangle-collection coverage and track an end-to-end output-path test separately?

PostHogTests/PostHogMaskSnapshotTest.swift
```diff
@@ -0,0 +1,647 @@
...
+        /// Heuristics off — scenarios demonstrating only the explicit
+        /// `.postHogMask()`/`.postHogNoMask()` modifiers. With the defaults on,
+        /// `maskAllTextInputs` masks ALL text (not just inputs), which would redact the
+        /// "visible" reference labels and muddy what the explicit modifier contributes.
+        private static func explicitMasksOnly(_ config: PostHogConfig) {
+            config.sessionReplayConfig.maskAllTextInputs = false
+            config.sessionReplayConfig.maskAllImages = false
+        }
+
+        private var scenarios: [Scenario] {
```
marandaneto, line 158: question: The PR description says the suite covers `maskAllSandboxedViews`, but none of these scenarios configures or exercises it. Is a scenario missing, or should that claim be removed from the description?

.github/workflows/test.yml
```diff
@@ -87,6 +87,71 @@ jobs:
...
 
+  # Golden-image masking regression. Any iPhone on the Makefile-pinned OS runtime renders
+  # identically; on failure the heatmap PNGs upload as an artifact.
+  mask-snapshots:
+    name: Session replay masking snapshots
+    needs: detect-markdown-only
+    runs-on: macos-15
+    permissions:
+      contents: read
+      pull-requests: write
```
marandaneto: suggestion: Could the PR-comment maintenance move to a separate no-checkout job? As written, same-repository PR code runs in a job with `pull-requests: write`. Keeping the test/artifact job read-only and granting write only to a follow-up comment job reduces the token exposure.

marandaneto (commented):
A few additional inline notes. I did not duplicate the existing intentional-mutation and empty-artifact-ID threads.

.github/workflows/test.yml
ioannisj: Good catch, done in bab6c1d. The comment id goes through an env var now and the `gh api` call quotes it, so Semgrep's green again.

Makefile
ioannisj: Added a guard for this in c723b9e: the target tees the raw log and asserts a Swift Testing test actually passed, so a 0-test run fails instead of reading green off the xcpretty summary. Amusingly my first cut of the guard grepped the function name and Swift Testing prints the `@Test` display name instead, so it false-failed once, fixed in 994e4d4.

.github/workflows/test.yml
ioannisj: Pinned it in c723b9e: `xcode-version: 26.3` on the mask job, plus a `MASK_SNAPSHOT_XCODE` assert in the Makefile so a compiler drift fails loud the same way the OS pin does. Left `latest-stable` everywhere else.

PostHogTests/PostHogMaskSnapshotTest.swift
ioannisj: Went with reusing the production painter, 0991b0c. Extracted `RRWireframe.maskImage` into a shared helper and the snapshot suite redacts through it now (rounded corners, scale-1), so the goldens are real replay output rather than hand-drawn squares. Still pre-WebP though, so a true end-to-end output-path test is worth a follow-up.

PostHogTests/PostHogMaskSnapshotTest.swift
ioannisj: You're right, none of the scenarios configure `maskAllSandboxedViews`. Dropped it from the description rather than claim coverage that isn't there. Could add a scenario for it later.

.github/workflows/test.yml
ioannisj: Did the split, ad6d3cd. The mask job is `contents: read` now and just exposes the artifact id as a job output, and a separate no-checkout job holds `pull-requests: write` and does the comment. Verified both on CI.

PR #743, opened 2026-07-30
feat: add push notification support for PostHog Workflows

ioannisj:

Ships push notification support for PostHog Workflows: APNs token registration (swizzled by default, manual via `registerPushNotificationToken(_:appId:)`), `$push_notification_opened` capture, identity-verified subscription requests, and durable register/unregister with passive retry.

The pieces of this branch were already reviewed as they landed (#729 push notification support, #735 identity verification token), but the branch has diverged a bit since: convergence with the Android implementation, including a fix for a race where a queued logout DELETE could cancel a re-registered subscription for the same identity. I think it's worth a fresh full review before release.

Push notification unit suite (60 tests on the iOS simulator: offline deferral, backoff, identity change, durable unregister, the DELETE/POST supersede), plus swiftlint/swiftformat.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

**DRI:** @ioannisj


PostHog/PushNotifications/PostHogPushNotificationOpenIntegration.swift
```diff
@@ -0,0 +1,66 @@
...
+            guard bundleExtension == "app" || bundleExtension == "appex" else {
+                hedgeLog("Push notification opened integration: skipping setup - not running in an app context")
+                return
+            }
+            token = DI.main.pushNotificationPublisher.onNotificationResponse.subscribe { [weak self] response in
+                // Auto-capture remote pushes only, matching Android's FCM-only auto-capture. A local
+                // notification (calendar/interval/location trigger, or none) is the app's own, not a
+                // delivered push; users who want to capture those can call
+                // `capturePushNotificationOpened(response:)` themselves — it stays unfiltered.
+                guard response.notification.request.trigger is UNPushNotificationTrigger else {
```
turnipdabeets: The trigger check gates on remote-vs-local but not on PostHog attribution, and `capturePushNotificationOpened(response:)` then forwards title/subtitle/body of any remote push. Android's auto path never sends content — it passes title/body as null and flattens only the `posthog` extra — so after an SDK upgrade an iOS app starts shipping third-party notification content (message previews, OTPs) to PostHog by default. I think we need to gate content capture on `userInfo["posthog"]` being present, or capture attribution-only events for unattributed pushes, before this ships.

PostHog/PushNotifications/PostHogPushNotificationOpenIntegration.swift
```diff
@@ -0,0 +1,66 @@
...
+                    return
+                }
+                self?.postHog?.capturePushNotificationOpened(response: response)
```
turnipdabeets, line 52: [suggestion] A dismiss can land here too: with a `.customDismissAction` category, swiping a push away invokes `didReceive` with `UNNotificationDismissActionIdentifier`, and we'd capture it as an open. Android can't observe dismissals at all, so filtering keeps both the event's meaning and parity: `guard response.actionIdentifier != UNNotificationDismissActionIdentifier else { return }`.

PostHog/PushNotifications/PostHogPushSubscriptionHandler.swift
```diff
@@ -0,0 +1,665 @@
...
+        stateLock.withLock {
+            cachedIdentityToken = nil
+            didAuthRetry = false
+        }
+    }
+
+    // MARK: - Private
+
+    private func resendIfDistinctIdChanged(currentDistinctId: String) {
+        guard let record = loadRecord(),
```
turnipdabeets, line 372: [suggestion] This subscriber runs synchronously on the main thread for every screen-name change (`screen()` → `notifyContextDidChange`), and `loadRecord()` hits the file system + JSON-decodes before we can tell the id didn't change. Android keeps the record in a `@Volatile` field hydrated from disk once. Could we cache the record (or at least `deliveredForDistinctId`) in memory so screen tracking doesn't pay disk I/O?

PostHog/PushNotifications/PostHogPushSubscriptionHandler.swift
```diff
@@ -0,0 +1,665 @@
...
+
+        attemptIfAllowed(deviceToken: record.deviceToken, appId: record.appId)
+    }
+
+    /// Unregister: `DELETE /api/push_subscriptions` for `distinctId`. The intent is persisted first, so an
+    /// offline or failed attempt is retried on `flush()`/next launch (see `PendingUnregister`) rather than
+    /// dropped — otherwise a logout while offline, or with an expired identity token, would leave the
+    /// device associated with the logged-out user on the backend. Cleared on a 2xx or a terminal 4xx.
+    func unregister(distinctId: String, deviceToken: String, appId: String) {
+        guard isAllowedProvider() else {
```
turnipdabeets: [suggestion] Android persists the intent first and gates only the network attempt — here the guard runs before `writePendingUnregister`, so a `reset()` while opted out loses the durable logout DELETE for good (the old identity stays subscribed even after a later opt-in). Could we move this guard below the persist, matching `send()`/`reregisterAfterReset()`?

PostHog/PushNotifications/PostHogPushSubscriptionHandler.swift
```diff
@@ -0,0 +1,665 @@
...
+    /// hold an expired cached token (the provider mints a short TTL). `isRetry` caps that to one re-mint.
+    private func attemptUnregister(_ pending: PendingUnregister, isRetry: Bool = false) {
+        guard isAllowedProvider() else { return }
+        guard isConnectedProvider() else {
+            hedgeLog("Push unregister deferred: no network connection. Will retry on flush/next launch.")
+            return
+        }
+        // Same one-verification-state as registration: the DELETE resolves a token for the distinct id it
+        // carries (the old identity on the reset() path).
+        resolveIdentityToken(distinctId: pending.distinctId, appId: pending.appId) { [weak self] identityToken in
```
turnipdabeets, line 254: [suggestion] The registration path re-checks its record after the mint, but this DELETE fires post-mint (up to ~10s with a slow provider) with only the allowed-guard. Log out of A → log back in as A during that window: the fresh POST lands, `markDelivered` clears the intent, and this DELETE still fires and unsubscribes the device. Worth re-checking `loadPendingUnregister() == pending` here before sending? (Same window exists on the Android twin.)

PostHog/PushNotifications/PostHogPushSubscriptionHandler.swift
```diff
@@ -0,0 +1,665 @@
...
+    /// Called from `PostHogSDK.flush()`.
+    func retryIfNeeded() {
+        // Drain any pending unregister first (offline logout, transport failure) — independent of the
+        // send record, usually absent after a logout; the in-memory identity-token cache is empty here,
+        // so the retry re-mints a fresh token and sidesteps the expired-token 401. If a same-identity
+        // registration is queued (logged out then back in), drop the DELETE instead — an in-flight
+        // DELETE completing after the POST would kill the subscription just delivered.
+        let record = loadRecord()
+        if let pending = loadPendingUnregister() {
+            if record != nil, pending.distinctId == distinctIdProvider() {
```
turnipdabeets: [nit] Dropping the DELETE on a distinctId match alone is right for token rotation (the backend keys per (person, app_id), so the queued POST overwrites) — but a pending unregister for a *different appId* gets dropped too, leaving that app's subscription behind. Matching on appId as well would close it. (Android's drain at PostHogPushSubscriptionManager.kt:153 has the same match.)

PostHog/PostHogApi.swift
```diff
@@ -236,6 +236,81 @@ class PostHogApi {
...
+        sendPushSubscription(
+            httpMethod: "POST", endpointName: "push subscription",
+            distinctId: distinctId, deviceToken: deviceToken, appId: appId,
+            identityToken: identityToken, completion: completion
+        )
+    }
+
+    /// Unregisters a device token: `DELETE /api/push_subscriptions/` with the same 5-field body as
+    /// registration (the backend `$unset`s `$device_push_subscription_<app_id>`). Best-effort — the
+    /// SDK fires this once and does not retry a failure.
```
turnipdabeets: [nit] "fires this once and does not retry a failure" is stale now — the handler persists a durable intent and retries on `flush()`/next launch, plus one 401 re-mint.

PostHog/PushNotifications/PostHogPushSubscriptionHandler.swift
```diff
@@ -0,0 +1,665 @@
...
+        recordLock.withLock { clearPendingUnregisterLocked(matching: pending) }
+    }
+
+    private func clearPendingUnregisterLocked(matching pending: PendingUnregister) {
+        guard loadPendingUnregisterLocked() == pending else { return }
+        storage.remove(key: .pushPendingUnregister)
+    }
+
+    /// Transport error (no status), 429, or 5xx is retryable; everything else (4xx) is terminal.
+    private func isRetryable(_ info: PostHogUploadInfo) -> Bool {
```
turnipdabeets, line 749: [nit] Of the three retryable inputs here, only 5xx is exercised by the suite — 429 and the nil-status transport-error branch have no test. Two small cases would pin the partition.

PostHogTests/PostHogIntegrationInstallationTest.swift
```diff
@@ -284,4 +303,39 @@ class PostHogIntegrationInstallationTest {
...
+        @Test("push notification opened integration skipped when swizzling is disabled")
+        func pushNotificationOpenedIntegrationSkippedWithoutSwizzling() async {
+            guard #available(iOS 14.0, macOS 11.0, *) else { return }
+            let sut = getSut(projectToken: "test_project_token", enableSwizzling: false, capturePushNotificationOpened: true)
+
+            #expect(sut.getPushNotificationIntegration() == nil)
+
+            sut.close()
+        }
+    #endif
```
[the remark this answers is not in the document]
ioannisj: Added the accessor and tests in 11276ac4d9.

PostHog/PostHogSwizzler.swift
```diff
@@ -7,8 +7,80 @@
...
+            )
+        } else {
+            let noopBlock: @convention(block) () -> Void = {}
+            class_addMethod(
+                targetClass,
+                swizzled,
+                imp_implementationWithBlock(noopBlock),
+                method_getTypeEncoding(swizzledMethod)
+            )
+        }
```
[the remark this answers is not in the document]
ioannisj: Added in 11276ac4d9.

PostHog/PushNotifications/PostHogPushSubscriptionHandler.swift
```diff
@@ -0,0 +1,665 @@
...
+        // registration is queued (logged out then back in), drop the DELETE instead — an in-flight
+        // DELETE completing after the POST would kill the subscription just delivered.
+        let record = loadRecord()
+        if let pending = loadPendingUnregister() {
+            if record != nil, pending.distinctId == distinctIdProvider() {
+                clearPendingUnregister(matching: pending)
+            } else {
+                attemptUnregister(pending)
+            }
+        }
```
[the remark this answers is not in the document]
ioannisj: Fixed in 47188e75cb, the drain matches on appId now.

PostHog/PushNotifications/PostHogPushSubscriptionHandler.swift
```diff
@@ -0,0 +1,665 @@
...
+            return
+        }
+        guard !distinctId.isEmpty, !deviceToken.isEmpty, !appId.isEmpty else {
+            hedgeLog("Push unregister skipped: missing distinct id, token, or app id.")
+            return
+        }
+        let pending = PendingUnregister(distinctId: distinctId, deviceToken: deviceToken, appId: appId)
+        writePendingUnregister(pending)
+        attemptUnregister(pending)
+    }
```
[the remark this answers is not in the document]
ioannisj: Fixed in 11276ac4d9, the intent is persisted before the gate now, matching Android.

PostHog/PushNotifications/PostHogPushNotificationOpenIntegration.swift
```diff
@@ -0,0 +1,66 @@
...
+            }
```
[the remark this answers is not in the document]
ioannisj: Fixed in 11276ac4d9, dismiss actions are excluded now.

PostHog/PostHogStorage.swift
```diff
@@ -254,6 +254,8 @@ class PostHogStorage {
         case errorTracking = "posthog.errorTracking"
         case capturePerformance = "posthog.capturePerformance"
         case deviceId = "posthog.deviceId"
+        case pushSubscription = "posthog.pushSubscription"
+        case pushPendingUnregister = "posthog.pushPendingUnregister"
```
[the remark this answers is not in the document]
ioannisj: Added a comment in 11276ac4d9 explaining why it's deliberately kept across reset().

PostHog/PostHogSDK.swift
```diff
@@ -612,6 +633,7 @@ let maxRetryDelay = 30.0
         queue?.flush()
         replayQueue?.flush()
         logsQueue?.flush()
+        pushSubscriptionHandler?.retryIfNeeded()
     }
 
     /// Resets local identity, super properties, feature flag cache, and session state.
```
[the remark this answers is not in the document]
ioannisj: Fixed in 11276ac4d9.

PostHog/PushNotifications/PostHogPushSubscriptionHandler.swift
```diff
@@ -0,0 +1,665 @@
...
+        // Retryable (transport error, 429, 5xx): keep the intent for flush/next launch.
+        if isRetryable(info) {
+            hedgeLog("Push unregister failed (status \(statusString(info))); will retry on flush/next launch.")
+            return
+        }
+
+        // Terminal (post-retry 401, 404 already gone, other 4xx): drop the intent, best-effort ceiling.
+        clearPendingUnregister(matching: pending)
+        hedgeLog("Push unregister failed (status \(statusString(info))); dropping (best-effort).")
+    }
```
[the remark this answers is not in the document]
ioannisj: Leaving this as is, Android's unregister path deliberately has no backoff either, so this is cross-platform parity rather than a gap.

PostHogTests/PostHogPushNotificationTest.swift
```diff
@@ -0,0 +1,1319 @@
...
+                let sut = getSDK(optOut: true)
+
+                sut.registerPushNotificationToken("deadbeef", appId: "com.example.app")
+
+                try await Task.sleep(nanoseconds: 300_000_000)
+                #expect(server.pushSubscriptionRequests.isEmpty)
+
+                sut.close()
+            }
+        #endif
```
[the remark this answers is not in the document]
ioannisj: Added in 11276ac4d9.

PostHog/PushNotifications/PostHogPushSubscriptionHandler.swift
```diff
@@ -0,0 +1,665 @@
...
+                    }
+                    return true
+                }
+                guard isFirst else { return }
+                hedgeLog(token != nil
+                    ? "Attaching freshly minted identity token to push request."
+                    : "No identity token attached to push request (provider completed nil).")
+                completion(token)
+            }
+        }
```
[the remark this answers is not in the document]
ioannisj: Correction: my earlier reply here was wrong, the winner gate was the problem, not the defense. It correctly drops a late completion's delivery, but the cache write sat inside the same gate, so a token minted after the watchdog was thrown away. A provider persistently slower than the 10s watchdog would then re-mint and time out on every send, forever token-less.

Fixed in 0d1e3205f7: the cache write now happens regardless of who wins the race (still under the same lock the opt-out clear uses), and only completion delivery stays gated. Regression test parks the mint past the watchdog and asserts the next send uses the cached token without re-minting.

PostHogTests/PostHogPushNotificationTest.swift
```diff
@@ -0,0 +1,1319 @@
...
+            func sdkRegistrationNoRequestWhenOptedOut() async throws {
+                let sut = getSDK(optOut: true)
+
+                sut.registerPushNotificationToken("deadbeef", appId: "com.example.app")
+
+                try await Task.sleep(nanoseconds: 300_000_000)
+                #expect(server.pushSubscriptionRequests.isEmpty)
+
+                sut.close()
+            }
```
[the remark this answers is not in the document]
ioannisj: Fixed in 11276ac4d9, those are `defer`'d now.

PostHog/PushNotifications/PostHogPushNotificationOpenIntegration.swift
ioannisj: Fixed in 11276ac4d9, dismiss actions are excluded now.

PostHog/PushNotifications/PostHogPushSubscriptionHandler.swift
ioannisj: Fixed in 11276ac4d9, persist now happens before the gate.

PostHog/PushNotifications/PostHogPushSubscriptionHandler.swift
ioannisj: Fixed in 47188e75cb, attemptUnregister re-checks the pending intent and applies the same supersede rule after the mint.

PostHog/PushNotifications/PostHogPushSubscriptionHandler.swift
ioannisj: Fixed in 47188e75cb, it matches on appId too now.

PostHog/PostHogApi.swift
ioannisj: Fixed in 47188e75cb.

PostHog/PushNotifications/PostHogPushSubscriptionHandler.swift
ioannisj: Added 429 and transport-error retry tests in 47188e75cb.

PostHog/PostHogStorage.swift
```diff
@@ -422,6 +424,8 @@ class PostHogStorage {
         deleteSafely(url(forKey: .sessionReplay))
         deleteSafely(url(forKey: .errorTracking))
         deleteSafely(url(forKey: .capturePerformance))
+        // Drop the push token on logout so a new user doesn't inherit the previous user's subscription.
+        deleteSafely(url(forKey: .pushSubscription))
```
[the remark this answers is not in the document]
ioannisj: Good catch, fixed in eb07567fbc, the clear now happens inside recordForReset()'s locked block instead of PostHogStorage.reset(), so a concurrent send() can't lose its record.

PostHog/PushNotifications/PushNotificationPublisher.swift
```diff
@@ -0,0 +1,257 @@
...
+                return
+            }
+
+            guard let appDelegateClass = swizzledAppDelegateClass else { return }
+            // Reverses the exchange. Methods added by swizzleAddingIfNeeded stay on the class — their
+            // IMPs are swapped back but the method entries remain. Known, harmless limitation.
+            swizzle(forClass: appDelegateClass, original: Self.didRegisterSelector, new: Self.swizzledDidRegisterSelector)
+            swizzle(forClass: appDelegateClass, original: Self.didFailSelector, new: Self.swizzledDidFailSelector)
+            swizzledAppDelegateClass = nil
+        }
```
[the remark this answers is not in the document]
ioannisj: Fixed in eb07567fbc, install and uninstall both always hop to the main queue now so they land in request order.

PostHog/PushNotifications/PostHogPushSubscriptionHandler.swift
```diff
@@ -0,0 +1,665 @@
...
+        self.storage = storage
+        self.config = config
+        self.distinctIdProvider = distinctIdProvider
+        self.isConnectedProvider = isConnectedProvider
+        self.isAllowedProvider = isAllowedProvider
+
+        contextChangedToken = onEventContextChanged.subscribe { [weak self] context in
+            guard let distinctId = context["distinct_id"] as? String, !distinctId.isEmpty else { return }
+            self?.resendIfDistinctIdChanged(currentDistinctId: distinctId)
+        }
```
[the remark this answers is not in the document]
ioannisj: Fixed in eb07567fbc, the subscriber now does a cheap in-memory check on the caller's thread and only a genuine identity change hops to a serial queue for the disk-backed re-check.

PostHog/PushNotifications/PostHogPushSubscriptionHandler.swift
ioannisj: Fixed in eb07567fbc, same change as the sync-read thread: in-memory check inline, the disk read moved to a serial queue.

PostHogTests/PostHogPushNotificationTest.swift
```diff
@@ -0,0 +1,1319 @@
...
+            handler.clearBackoffForTesting()
+            handler.retryIfNeeded()
+            try await Task.sleep(nanoseconds: 200_000_000)
+            #expect(server.pushSubscriptionRequests.count == 1)
+
+            // Record kept, not marked delivered.
+            #expect(record(storage)?["deviceToken"] == "tok")
+            #expect(!delivered(storage))
+        }
+
```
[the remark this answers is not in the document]
ioannisj: Covered in 47188e75cb, retries429ThenSucceeds and retriesTransportErrorThenSucceeds hit the 429 and nil-status branches.

PostHog/PushNotifications/PostHogPushNotificationOpenIntegration.swift
ioannisj: Yeah agreed, fixed in c74f1283d8. The response-based path now forwards title/subtitle/body only when userInfo has a posthog key; unattributed pushes still capture the open with no content, matching Android's auto path. The field-based overload stays ungated since there the caller passes content explicitly.

PostHog/PostHogSDK.swift
```diff
@@ -2901,6 +2962,164 @@ let maxRetryDelay = 30.0
...
+        @objc public func capturePushNotificationOpened(response: UNNotificationResponse) {
+            let content = response.notification.request.content
+            capturePushNotificationOpened(
+                title: content.title,
+                subtitle: content.subtitle,
+                body: content.body,
+                payload: content.userInfo,
+                action: response.actionIdentifier
+            )
+        }
```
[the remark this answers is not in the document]
ioannisj: Fixed in c74f1283d8, the auto path now gates title/subtitle/body on the posthog attribution key, mirroring the existing $notification_<key> gate and Android's behavior. No new unit test since UNNotificationResponse isn't constructible in tests; the gate feeds the already-covered field overload.

PostHog/PushNotifications/PushNotificationPublisher.swift
```diff
@@ -0,0 +1,257 @@
...
+            onDeviceToken = PostHogMulticastCallback(onSubscriberCountChanged: { count in
+                guard let self = weakSelf else { return }
+                if count == 1 {
+                    self.swizzleAppDelegateMethods()
+                } else if count == 0 {
+                    self.unswizzleAppDelegateMethods()
+                }
+            })
+            weakSelf = self
+        }
```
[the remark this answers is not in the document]
ioannisj: Deferring this one: `UNNotificationResponse` has no public initializer, so the delegate-forwarding path can't be unit-tested without adding an abstraction seam, and the swizzle mutates the shared test-host app delegate. The publisher is thin glue, the handler and capture paths underneath are covered. Worth a follow-up if we add a seam.

ioannisj:
Resolved all comments above. Thank you 🤖 and good humans

PostHog/PushNotifications/PostHogPushSubscriptionHandler.swift
```diff
@@ -0,0 +1,734 @@
...
+                return
+            }
+            // The identity may have changed while the mint was in flight (identify/reset racing a
+            // slow provider): sending under the stale distinctId would register the token to the
+            // wrong person. Skip — the identify/reset-driven resend already covers the new identity.
+            guard self.distinctIdProvider() == distinctId else {
+                hedgeLog("Push subscription skipped: distinct id changed while minting the identity token.")
+                stateLock.withLock {
+                    self.isSending = false
+                    self.pendingResend = false
```
dustinbyrne, line 453: feels like an edge case, but if `reset()` happens while we're waiting on `pushIdentityProvider`, `reregisterAfterReset()` sets `pendingResend = true`

so the token can be registered for the new anonymous id after this request finishes

this clears it instead. should we preserve the pending reset here?

PostHog/PushNotifications/PostHogPushSubscriptionHandler.swift
```diff
@@ -0,0 +1,734 @@
...
+            if let record = loadRecord(), pending.distinctId == distinctIdProvider(), pending.appId == record.appId {
+                clearPendingUnregister(matching: pending)
+                return
+            }
+            api.deletePushSubscription(
+                distinctId: pending.distinctId, deviceToken: pending.deviceToken, appId: pending.appId,
+                identityToken: identityToken
+            ) { [weak self] info in
+                self?.handleUnregisterResult(info, pending: pending, isRetry: isRetry)
+            }
```
dustinbyrne, line 271: registration and unregistration could still race once both requests are in flight. an older `POST` could finish after reset's `DELETE` and resubscribe the old identity, or a `DELETE` could finish after a newer `POST` and remove it.

the android implementation avoids this by serializing requests

PostHog/PushNotifications/PostHogPushSubscriptionHandler.swift
dustinbyrne: or maybe it's sufficient to just let the registration retry on another flush / next launch?

dustinbyrne (approved):
looking good. there's still a couple of edge cases, but they feel pretty narrow. i'll let you decide how/if you want to handle them

PostHog/PushNotifications/PostHogPushSubscriptionHandler.swift
ioannisj: Good catch, the interleaving is real. reset()'s re-register folds into `pendingResend` while the mint is in flight, and the stale-identity guard was dropping it instead of servicing it.

It did self-heal already (the record is persisted before the drop, and `retryIfNeeded` re-attempts on the next flush/launch independent of `pendingResend`), so worst case was about one flush interval with the new anon id unregistered. But the fix was small enough that I went for it: the guard now services the folded resend the same way `handleResult` does, in 35ac9bc992, plus a regression test that parks the mint, resets mid-mint, and asserts the POST for the new anon id arrives without needing a flush.

PostHog/PushNotifications/PostHogPushSubscriptionHandler.swift
ioannisj: Yeah, the key word here is "in-flight". The intent guards already cover everything up to dispatch (the post-mint re-check drops a superseded DELETE, and a reset during a POST's mint never reaches the wire), so what's left is two requests racing on the wire itself, which makes the window pretty small.

Went for the fix anyway so the two implementations are in parity: 006f912766 serializes the push POST/DELETE calls so only one is ever in flight, same guarantee the Android executor gives. Regression test holds a POST's response mid-flight, fires `reset()`, and asserts the DELETE waits for it.

ioannisj:
Fixed a flaky test and addressed new feedback. Should be good to merge

Issue #746, opened 2026-08-05
Push: opt-out during an in-flight unregister can strand the DELETE, and opting back in never re-registers the token


ioannisj:
### Problem

`unregister` persists its `PendingUnregister` and then defers on `guard isAllowedProvider()` when the SDK is disabled or opted out, and it re-checks consent again after the asynchronous identity-token mint. So `unregisterPushNotificationToken()` followed closely by opt-out, the standard shape of a host app's opt-out flow, can park the DELETE: the pending record survives, but the same consent gates keep deferring it on every retry while opted out. The server-side subscription stays active for the whole opted-out period and Workflows keep delivering to a device that opted out.

On the way back: a successful DELETE clears the stored token, and opting back in only restores consent and the observer, it doesn't refetch the APNs token or re-register. Push stays dead until the next app restart re-runs registration.

Found while reviewing the React Native push integration (PostHog/posthog-js#4415, [thread](https://github.com/PostHog/posthog-js/pull/4415#discussion_r3719629071)): the wrapper orders its bridge calls (unregister before opt-out) but that can't order the handler's internal async work, so the fix has to live here.

### Proposal

- Allow a pending unregister to execute while opted out: a DELETE is data removal, not collection, so the opt-out gate should block registration and sends but not cleanup.
- On opt-in with `capturePushNotificationSubscriptions` enabled, refetch the token and re-register, so opting back in re-arms push without an app restart.

PR #759, opened 2026-08-10
fix: send the push unregister DELETE even after opt-out (#746)

dmarchuk:
## Problem

Opting out did not stop Workflows push. `unregister` deferred on `guard isAllowedProvider()` (false while opted out) and re-checked it after the async identity-token mint. The normal opt-out flow — `unregisterPushNotificationToken()` then `optOut()` — parked the DELETE: the persisted intent kept deferring on every retry while opted out. The server-side subscription stayed active for the whole opted-out period, so Workflows kept delivering to a device that opted out (#746).

## Change

Split the cleanup gate from the send gate. Added an `isEnabledProvider` (SDK enabled, ignoring opt-out) and gated the unregister DELETE on it, instead of `isAllowedProvider` (enabled and not opted out). A DELETE removes data, so opt-out must block registration and sends but not cleanup. The register and send paths keep `isAllowedProvider`.

- New test `PostHogPushOptOutReproTest`: registers a token, parks the unregister's identity mint, flips opt-out mid-flight via the logout path (`unregisterCurrentToken`), then releases the mint. Asserts the DELETE reaches the mock server. Fails on the pre-fix code (reproduces the bug), passes with the fix.
- Split the existing `unregisterGuarded` test into "disabled, no DELETE" and "opted out, DELETE still sent". These run in CI on `test-ios-simulator`.
- Real-device E2E on a physical iPhone against a local stack. Registered a token, then ran the opt-out flow (`unregisterPushNotificationToken()` then `optOut()`) with the identity mint delayed to force the race. The DELETE reached the backend (200). Rebuilt with the pre-fix `isAllowedProvider` guard and the same flow sent no DELETE. The fix is what stops the strand.

## Not in this PR

#746 also reports that opting back in does not refetch the APNs token or re-register. That half is fixed separately in #760.

> [!NOTE]
> Draft. Directed by @dmarchuk; implemented and verified locally by Claude while @ioannisj is out.


PostHogTests/PostHogPushNotificationTest.swift
```diff
@@ -1677,4 +1688,121 @@
         }
     }
 
+    // Regression for the opt-out / unregister race found during the RN push review (posthog-ios#746):
+    // opt-out during an in-flight unregister must still send the DELETE. Lives in this in-target file
+    // (not a standalone one) so `test-ios-simulator` actually compiles and runs it.
+    @Suite("Push opt-out / unregister bug repros", .serialized)
```
marandaneto, line 1694: suggestion: Reuse the existing push notification test fixture

This one-off suite duplicates the existing `MockPostHogServer` lifecycle plus `ParkedMint`, `waitFor`, and `delivered` for a single regression test, increasing maintenance overhead without representing distinct behavior. Please move this test into `PostHogPushNotificationTest` and reuse its fixture and helpers.

marandaneto (approved):

PostHogTests/PostHogPushNotificationTest.swift
dmarchuk: Makes sense, moved it as suggested, thank you!

PR #760, opened 2026-08-10
fix(push): re-register the device token on opt-in (#746)

dmarchuk:
## Problem

The opt-in half of #746. After a logout `unregister` clears the stored push token, `optIn()` only restores consent and the notification observer. It never refetches the APNs token or re-registers the device, so a user who opts out and back in stays unsubscribed and Workflows can't deliver to them until the next app restart re-runs registration.

## Change

On `optIn()`, when `capturePushNotificationSubscriptions` is enabled, the SDK now re-requests the APNs device token. The OS redelivers it to the (re-installed) subscription integration, which re-registers the device. A new `PostHogPushNotificationSubscriptionIntegration.requestTokenRefresh` hook encapsulates the `registerForRemoteNotifications()` call (with the same app-context guard `start()` uses) and is overridable in tests.

Manual push mode (`capturePushNotificationSubscriptions = false`) is unchanged: the host owns the token lifecycle, so opt-in does not auto-refetch.

## How tested

- Unit tests in `PostHogPushNotificationTest`: opt-in re-requests the token when auto-capture is on, and does not when it's off.
- Real-device E2E (iPhone, real backend): register, opt out (which clears the token), then opt in, and observed a re-register `POST /api/push_subscriptions`. With the refetch disabled, the same taps produced no re-register, confirming the fix is what re-arms push.

## Not in this PR

The unregister-DELETE-during-opt-out half of #746 is fixed separately in #759. This PR is independent and targets `main`.

🤖 Draft — opening for review; will un-draft after self-review.

### DIFF
--- a/PostHog/PostHogSDK.swift
+++ b/PostHog/PostHogSDK.swift
@@ -2427,6 +2427,15 @@ let maxRetryDelay = 30.0
             notifyContextDidChange()
             notifyExceptionStepsDidChange()
         }
+
+        #if os(iOS)
+            // A prior logout unregister cleared the push token; opt-in re-installs the subscription
+            // integration above but that alone doesn't refetch the token. Re-request it so the
+            // redelivered token re-registers this device, re-arming push without an app restart (#746).
+            if #available(iOS 14.0, *), config.capturePushNotificationSubscriptions {
+                PostHogPushNotificationSubscriptionIntegration.requestTokenRefresh()
+            }
+        #endif
     }
 
     /// Opts the current user out of data capture.
--- a/PostHog/PostHogSDK.swift
+++ b/PostHog/PostHogSDK.swift
@@ -2432,7 +2432,10 @@ let maxRetryDelay = 30.0
             // A prior logout unregister cleared the push token; opt-in re-installs the subscription
             // integration above but that alone doesn't refetch the token. Re-request it so the
             // redelivered token re-registers this device, re-arming push without an app restart (#746).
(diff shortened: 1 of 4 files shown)


PostHog/PostHogSDK.swift
```diff
@@ -2427,6 +2427,15 @@ let maxRetryDelay = 30.0
             notifyContextDidChange()
             notifyExceptionStepsDidChange()
         }
+
+        #if os(iOS)
+            // A prior logout unregister cleared the push token; opt-in re-installs the subscription
+            // integration above but that alone doesn't refetch the token. Re-request it so the
+            // redelivered token re-registers this device, re-arming push without an app restart (#746).
+            if #available(iOS 14.0, *), config.capturePushNotificationSubscriptions {
```
marandaneto: blocking: When `capturePushNotificationSubscriptions` retains its default `true` but `enableSwizzling` is `false`, `installIntegrations()` skips this subscription integration, yet this branch still calls `registerForRemoteNotifications()`. In this documented manual-registration configuration, `optIn()` unexpectedly triggers the host's APNs lifecycle and cannot automatically forward the resulting token because no SDK observer was installed. Please gate the refresh on `enableSwizzling` as well, or on successful integration installation, and cover this configuration with a test.

marandaneto (approved):
approving to unblock but see comments

PostHog/PostHogSDK.swift
dmarchuk: Good catch, thank you!

PR #767, opened 2026-08-18
feat: skip push registration when the project has no integration for the app_id

dmarchuk:

A device on a project without push configured makes a registration request on every app open that the server can only throw away, and a project that turns push on afterwards never reaches the devices it already has.

- The SDK registers its token at startup whenever push capture is on, against every project it reports to. It has no way to know whether that project opted in.
- `/api/push_subscriptions/` answers those with `200 {"stored": false, "push_enabled": false}` and discards the token.
- Because any 2xx writes `deliveredForDistinctId`, the device then records a success for a token the server never kept and stops asking. No response reaches a client that has stopped asking, so a project that configures push later reaches nothing installed before it — until a token rotation, an `identify()` to a different distinct id, or a reinstall.

Mirrors PostHog/posthog-android#703. Field names and the three client rules are kept identical between the two SDKs, the same way PostHog/posthog-ios#735 and PostHog/posthog-android#656 were.

Server side is merged: PostHog/posthog#83148 publishes the list, and the contract is `docs/internal/push-subscription-registration.md` there.

## Changes

Remote config now carries `push.appIds` — the Firebase project ids and APNs bundle ids a project accepts registrations for.

- **The send is gated in `attemptIfAllowed`**, the choke point every caller already goes through. The record is still persisted when the send is gated, so `onPushAppIdsChanged` has a token to register later rather than waiting for the app to hand us one again.
- **Absent stays permissive.** A server older than the key sends nothing, which an SDK cannot tell apart from a project with push disabled, so `nil` means attempt. Only a published empty list means skip.
- **`onPushAppIdsChanged` clears the delivered marker** for an `app_id` that just became registerable, and re-registers. It fires only on that transition — firing on every config load would put the request back on every launch, which is what the marker exists to prevent.

Two places where iOS legitimately differs from the Android PR, rather than diverging by accident:

- **No per-slice disk cache is needed.** Android caches each config slice separately, so it needed a new `push` preference to survive a cold start. iOS already persists the whole remote config blob, so `preloadPushConfig` reads the slice from it on launch, before the first registration attempt.
- **`processPushConfig` is not re-armed on the `/flags` paths**, where `processErrorTrackingConfig(nil)` is. The in-memory list already holds the cached value there, and a re-arm would compare the list against itself and find no transition.

On the first launch that ever sees the key, the previous list is treated as empty rather than unknown, so every configured `app_id` counts as new. That costs one request per device, once, and it is the only way to reach a device whose project was configured before it updated to a version that reads this key.

New cases in `PostHogPushNotificationTest`, mirroring the Android set one for one:

- An `app_id` outside the list sends nothing but still persists the record. Catches the gate being dropped, and catches the record not being persisted, which would leave `onPushAppIdsChanged` with no token.
- A nil list still registers. Catches the gate failing closed against a server older than the key, which would silently disable push on every such deployment.
- An `app_id` inside the list registers. Catches the gate being inverted.
- An `app_id` becoming registerable clears the delivered marker and re-registers. This is the un-stranding path.
- An unrelated `app_id` becoming registerable does not re-register. Catches over-firing.

### Manual E2E on a simulator against a local server

Ran the six-scenario verification brief on an iPhone 17 Pro simulator (PostHogExample as `com.posthog.pushtest`) against a local PostHog serving real `push.appIds` remote config for a team with a togglable APNs integration.

- Compile fix first: the branch did not build (two `self.` capture errors in `onPushAppIdsChanged`). Fixed in this PR; the 83 `PostHogPushNotificationTest` cases then pass on the simulator.
- Configured launch: one POST, server store, person property `$device_push_subscription_com.posthog.pushtest` set.
- Unconfigured launch: no POST, "skipped: not configured", pending record written without a delivered marker.
- Cold start unconfigured: skip logged before a deliberately delayed `/config` responded, so the disk-cached slice gated it, not the fresh fetch.
- Un-stranding: after the integration was added, the next launch logged "became registerable; re-registering" and sent exactly one POST.
- No repeat: three further relaunches sent nothing.
- Old server: with the `push` key removed from the response, the SDK registered.

Simulated, labeled: the iOS simulator does not mint an APNs token, so the token was fed through `registerPushNotificationToken`, the same `handler.send()` gate the swizzled path uses, with a fixed literal. The swizzled auto-registration itself needs a physical device.

**Real-device follow-up (iPhone 16 Pro, iOS 26.6, real APNs token via the swizzle path):** configured launch registered once and stored the token (person property `$device_push_subscription_com.posthog.pushtest` set); after removing the APNs integration a fresh launch registered nothing server-side. This exercises the swizzled auto-registration the simulator can't, so the token substitution above no longer applies to the configured/unconfigured paths.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Authored with Claude Code, after PostHog/posthog-android#703 landed the reference implementation. Android went first deliberately: its module builds on the JVM, so the design got a compile check there before being mirrored to a platform I cannot build at all.

I confirmed the scope rather than assuming it — a code search across the org for callers of `/api/push_subscriptions` returns only posthog-android and posthog-ios. posthog-flutter forwards to both natives and inherits the gate with no change; posthog-react-native has been emptied and points at posthog-js, which has no caller.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Bbav6vE7Tm24KkdRhZmvNM)_

### DIFF
--- a/PostHog/PostHogRemoteConfig.swift
+++ b/PostHog/PostHogRemoteConfig.swift
@@ -29,6 +29,13 @@ class PostHogRemoteConfig {
     private var recordingMinimumDuration: TimeInterval?
 
     private let errorTrackingLock = NSLock()
+    private let pushLock = NSLock()
+    /// The app_ids this project accepts push registrations for, or nil when no server has told us.
+    /// Nil is not the empty list: it means attempt the registration, because a server older than
+    /// the key cannot be told apart from a project with push disabled.
+    private var pushAppIds: [String]?
+    /// app_ids that became registerable on the most recent /config read, drained by the handler.
+    private var newlyRegisterablePushAppIds: Set<String> = []
     private var autoCaptureExceptions = false
 
     private var flags: [String: Any]?
@@ -139,6 +146,7 @@ class PostHogRemoteConfig {
 
         preloadSessionReplay()
         preloadErrorTrackingConfig()
+        preloadPushConfig()
 
         // Remote config is always loaded (config.remoteConfig is now a no-op)
         preloadRemoteConfig()
@@ -236,6 +244,11 @@ class PostHogRemoteConfig {
 
                 // process error tracking config
                 self.processErrorTrackingConfig(config)
+
+                // process push config. Unlike the others this is not re-armed from cache on the
+                // /flags paths below: the in-memory list already holds the cached value, and a
+                // re-arm would compare the list against itself and find no transition.
+                self.processPushConfig(config)
 
                 // notify
                 DispatchQueue.main.async {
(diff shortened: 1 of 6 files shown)


PostHog/PostHogRemoteConfig.swift
```diff
@@ -652,6 +666,53 @@ class PostHogRemoteConfig {
...
+        }
+    }
+
+    private func preloadPushConfig() {
+        let push = remoteConfigLock.withLock {
+            getCachedRemoteConfig()?["push"] as? [String: Any]
+        }
+        if let appIds = push?["appIds"] as? [String] {
+            pushLock.withLock {
+                pushAppIds = appIds
```
marandaneto, line 714: blocking: Cached app IDs suppress upgrade recovery

`preloadPushConfig()` seeds `pushAppIds` from the cache, so if an older SDK already persisted `push.appIds`, the live response compares the same IDs and produces no newly registerable transition. A previously discarded registration therefore keeps its delivered marker indefinitely after upgrade. Please add durable one-time migration state that resends even when cached and live IDs match.

PostHog/PushNotifications/PostHogPushSubscriptionHandler.swift
```diff
@@ -388,6 +393,36 @@ final class PostHogPushSubscriptionHandler {
...
+            guard let record, newlyRegisterable.contains(record.appId) else {
+                // Either nothing changed for this device, or the app_id was already registerable and
+                // the existing delivered marker is honest. Re-sending here would put the request back
+                // on every launch, which is what the marker exists to prevent.
+                return
+            }
+
+            if record.deliveredForDistinctId != nil {
+                recordLock.withLock {
+                    self.writeRecord(deviceToken: record.deviceToken, appId: record.appId)
```
marandaneto: blocking: Concurrent token updates can be overwritten

The record is read under one lock acquisition, then this second acquisition writes that stale snapshot. If `send()` stores a newer APNs token between them, this overwrites it with the old token; the newer request then cannot mark itself delivered, losing the latest token. Please verify the current record and clear its delivered marker in one atomic read-modify-write, mutating only if it still matches.

marandaneto (commented):
Inline findings from the review.

marandaneto (approved):
left comments, approving to unblock

marandaneto:
lint is failing!

PostHog/PostHogRemoteConfig.swift
dmarchuk: Updated, thank you!

PostHog/PushNotifications/PostHogPushSubscriptionHandler.swift
dmarchuk: Good catch, thanks!

PR #774, opened 2026-08-20
fix: Drop events and logs when Objective-C beforeSend callbacks raise

marandaneto:

Swift `BeforeSendBlock` callbacks are non-throwing, but boxed event and log callbacks supplied from Objective-C can raise `NSException`. Those exceptions previously escaped the Swift callback chain and could crash capture instead of safely rejecting the item.

This change invokes boxed Objective-C callbacks through an Objective-C exception boundary. If a callback raises, the SDK logs the exception, stops the callback chain, and drops the event or log without queueing or sending it. Native Swift callback behavior is unchanged.

- Focused `PostHogSDKTest` run through XcodeBuildMCP using the `Testing` configuration - 97 tests passed
- Focused `PostHogLogsCaptureTests` run through XcodeBuildMCP using the `Testing` configuration - 17 tests passed
- `make swiftLintCheck`
- `make swiftFormatCheck`
- `pnpm changeset status`
- `git diff --check`

The regression tests raise Objective-C exceptions, verify capture returns normally, confirm later callbacks do not run, and check queues and network requests remain empty. A positive event test also proves a working boxed callback executes through the dynamic Objective-C accessor and modifies the sent event.

## :pencil: Checklist

## If releasing new changes

## 🤖 Agent context

Pi coding agent helped audit the Objective-C bridge, add the exception boundary and regression fixtures, run validation, and prepare this PR. Reviewer feedback prompted positive callback coverage, protection for boxed log callbacks, and documentation of why the Objective-C accessor is resolved dynamically. The generated Swift Objective-C header was checked to confirm the exported selector.

### DIFF
--- a/PostHog/ObjCExceptionSupport/PHObjCExceptionCatcher.m
+++ b/PostHog/ObjCExceptionSupport/PHObjCExceptionCatcher.m
@@ -10,6 +10,9 @@ typedef id _Nullable (^PHObjectTransformBlock)(id value);
                          withValue:(id)value
                        onException:(void (^)(NSException *exception))onException {
     @try {
+        // The box types are generated from Swift in this target, so this file cannot import
+        // PostHog-Swift.h. Resolve their @objc block accessor dynamically and invoke it here so
+        // an NSException never unwinds through a Swift frame.
         PHObjectTransformBlock block = ((id (*)(id, SEL))objc_msgSend)(box, NSSelectorFromString(@"block"));
         return block(value);
     } @catch (NSException *exception) {
--- a/PostHog/ObjCExceptionSupport/PHObjCExceptionCatcher.m
+++ b/PostHog/ObjCExceptionSupport/PHObjCExceptionCatcher.m
@@ -1,0 +1,21 @@
+#import "PHObjCExceptionCatcher.h"
+
+#import <objc/message.h>
+
+typedef id _Nullable (^PHObjectTransformBlock)(id value);
+
+@implementation PHObjCExceptionCatcher
+
++ (nullable id)invokeBlockFromBox:(id)box
+                         withValue:(id)value
+                       onException:(void (^)(NSException *exception))onException {
+    @try {
+        PHObjectTransformBlock block = ((id (*)(id, SEL))objc_msgSend)(box, NSSelectorFromString(@"block"));
+        return block(value);
+    } @catch (NSException *exception) {
+        onException(exception);
+        return nil;
+    }
+}
+
+@end
--- a/PostHog/Utils/BoxedBeforeSend.swift
+++ b/PostHog/Utils/BoxedBeforeSend.swift
@@ -50,4 +50,11 @@ import Foundation
     public init(block: @escaping PostHogBeforeSendLogBlock) {
         self.block = block
     }
+
+    func invokeSafely(with record: PostHogLogRecord) -> PostHogLogRecord? {
+        PHObjCExceptionCatcher.invokeBlock(from: self, with: record) { exception in
+            let reason = exception.reason ?? "No reason provided"
+            hedgeLog("Objective-C log beforeSend callback raised \(exception.name.rawValue): \(reason). The log was dropped.")
+        } as? PostHogLogRecord
+    }
 }
--- a/PostHog/Utils/BoxedBeforeSend.swift
+++ b/PostHog/Utils/BoxedBeforeSend.swift
@@ -4,6 +4,11 @@
 //
 
 import Foundation
+#if compiler(>=6.0)
+    internal import PostHogObjCExceptionSupport
+#else
+    @_implementationOnly import PostHogObjCExceptionSupport
+#endif
 
 /// ObjC wrappers for the Swift function-typed `beforeSend` chains: Swift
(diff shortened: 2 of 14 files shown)


PostHogTests/PostHogSDKTest.swift
```diff
@@ -1330,6 +1333,35 @@ class PostHogSDKTest: QuickSpec {
                 expect(events[0].event).to(equal("other_test"))
             }
 
+            it("contains Objective-C exceptions from boxed callbacks") {
```
turnipdabeets, line 1353: [suggestion] This test passes whether the exception was caught *or* the `block` selector failed to resolve — `objc_msgSend` would raise `unrecognized selector`, the same `@try` would catch it, and we'd still see an empty queue and an uninvoked second callback.

Since every boxed callback now routes through `invokeSafely`, a broken selector means ObjC callers lose 100% of their events while this test stays green. There's currently no test in the target that a *working* boxed callback lets an event through, so nothing pins it.

A plain Swift-constructed box still exercises the `objc_msgSend` lookup, so this is enough:

```swift
it("runs boxed callbacks through the exception boundary") {
    sut = self.getSut(flushAt: 1)
    let boxes: [NSObject] = [
        BoxedBeforeSendBlock { event in
            event.event = "boxed_rename"
            return event
        },
    ]
    PHBeforeSendExceptionTestFixture.setBeforeSend(boxes, on: sut.config)

    sut.capture("objc_boxed_event")

    let events = getBatchedEvents(server)
    expect(events.count).to(equal(1))
    expect(events[0].event).to(equal("boxed_rename"))
}
```

PostHog/Utils/BoxedBeforeSend.swift
```diff
@@ -23,6 +28,13 @@ import Foundation
     public init(block: @escaping BeforeSendBlock) {
         self.block = block
     }
+
+    func invokeSafely(with event: PostHogEvent) -> PostHogEvent? {
```
turnipdabeets, line 32: [suggestion] `BoxedBeforeSendLogBlock` just below has the same hole — `PostHogLogsConfig.setBeforeSend(_ blocks: [BoxedBeforeSendLogBlock])` still does `blocks.map(\.block)`, so an ObjC log callback raising `NSException` crashes exactly the way the events path did before this PR. That's a live crash path today, not a latent one.

`PHObjCExceptionCatcher.invokeBlockFromBox:` is already generic over the box, so it's the same shape:

```swift
func invokeSafely(with record: PostHogLogRecord) -> PostHogLogRecord? {
    PHObjCExceptionCatcher.invokeBlock(from: self, with: record) { exception in
        let reason = exception.reason ?? "No reason provided"
        hedgeLog("Objective-C log beforeSend callback raised \(exception.name.rawValue): \(reason). The record was dropped.")
    } as? PostHogLogRecord
}
```

Fine as a follow-up if we'd rather keep this PR scoped to events — just want it tracked somewhere durable rather than in a merged PR thread.

PostHog/ObjCExceptionSupport/PHObjCExceptionCatcher.m
```diff
@@ -0,0 +1,21 @@
...
+
+typedef id _Nullable (^PHObjectTransformBlock)(id value);
+
+@implementation PHObjCExceptionCatcher
+
++ (nullable id)invokeBlockFromBox:(id)box
+                         withValue:(id)value
+                       onException:(void (^)(NSException *exception))onException {
+    @try {
+        PHObjectTransformBlock block = ((id (*)(id, SEL))objc_msgSend)(box, NSSelectorFromString(@"block"));
```
turnipdabeets, line 16: [nit] Worth a line saying *why* we reach for the selector instead of importing the box's type — this file lives inside the framework target and can't import the generated `PostHog-Swift.h`, and calling the block from an ObjC frame keeps Swift frames out of the unwind path. Both are good reasons, but right now it reads as incidental rather than deliberate.

turnipdabeets (approved):
Approving. Routing the boxed callbacks through a real ObjC exception boundary is the right fix, and keeping native Swift behavior untouched is the correct scope call.

Three comments inline: one on the regression test (worth doing here), one on the logs box, one nit.

PostHogTests/PostHogSDKTest.swift
marandaneto: Good catch! Fixed in 02f722c46 — added a positive boxed-callback test that verifies the callback executes and modifies the event sent on the wire.

PostHog/Utils/BoxedBeforeSend.swift
marandaneto: Fixed in 02f722c46 — boxed Objective-C log callbacks now use the same exception boundary, with coverage for dropping the log and stopping later callbacks.

PostHog/ObjCExceptionSupport/PHObjCExceptionCatcher.m
marandaneto: Fixed in 02f722c46 — documented why the Swift-generated box accessor is resolved dynamically and invoked from Objective-C.

PR #776, opened 2026-08-22
fix: wait for current dSYM before uploading symbols

### LINKED ISSUE #4574: `posthog-react-native/expo`: fix iOS dSYM build ordering and EAS versions
## Bug description

i think the `posthog-react-native/expo` plugin has two bugs around iOS symbol uploads

### the dSYM upload can run before the app dSYM is ready

i hit this during an EAS archive:

```text
ExampleApp.app.dSYM/Contents/Resources/DWARF/ExampleApp: The file was not recognized as a valid object file
```

the generated PostHog Xcode phase doesn't have an input dependency on the app's dSYM executable, so Xcode can run it before the dSYM has finished being generated

adding this as an input path fixed the build:

```text
$(DWARF_DSYM_FOLDER_PATH)/$(DWARF_DSYM_FILE_NAME)/Contents/Resources/DWARF/$(EXECUTABLE_NAME)
```

this is also the input path in PostHog's manual iOS setup docs, so i'm wondering if the Expo plugin has just missed adding it?

### the symbols can be uploaded against the wrong version

once the build worked, the symbols uploaded successfully, but they ended up under the wrong release in PostHog

the actual archive was version `2.10.0 (154)`, but the Hermes source map and native symbols were attached to:

```text
com.example.app@1.0+1
```

rather than:

```text
com.example.app@2.10.0+154
```

it looks like the upload scripts use `MARKETING_VERSION` and `CURRENT_PROJECT_VERSION`. With Expo prebuild these can still be the generated Xcode defaults, `1.0` and `1`, even though EAS remote versioning has written the real version and build number into the app's Info.plist

so the upload works, but the symbols are attached to a release that doesn't match the shipped app

## How to reproduce

1. set up an Expo app with `posthog-react-native` and enable iOS source map/dSYM uploads through the `posthog-react-native/expo` plugin
2. set `appVersionSource` to `remote` in `eas.json` and use `autoIncrement: true` for the iOS build profile
3. run an EAS iOS archive build with the PostHog upload credentials configured
4. the build can fail in the PostHog dSYM phase because the app dSYM isn't a valid object file yet
5. if it completes, check the release in PostHog. The symbols can be uploaded under the generated Xcode version, `1.0+1`, instead of the version in the archive

## Related sub-libraries

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-js-lite (web lite)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

## Additional context

i reproduced this with:

- Expo SDK 57.0.13
- `posthog-react-native` 4.63.2
- `@posthog/react-native-plugin` 2.4.0
- `posthog-ios` 3.69.6
- PostHog CLI 0.11.3
- EAS remote app versions

i've tested a `patch-package` workaround that:

1. adds the dSYM executable to the PostHog phase's `inputPaths`
2. reads `CFBundleShortVersionString` and `CFBundleVersion` from the source `INFOPLIST_FILE` for the Expo upload phases
3. leaves the existing `MARKETING_VERSION` and `CURRENT_PROJECT_VERSION` values alone if the plist values are missing

the input dependency fixed the EAS archive failure. the version change was verified against the generated Expo project, and there is now a PostHog release using the correct app version

one slightly dodgy bit i wanted to avoid: i don't think the upload phase should prefer `${TARGET_BUILD_DIR}/${INFOPLIST_PATH}` unless that processed plist is declared as an input. It could already exist from an earlier incremental build and contain a stale version. i didn't reproduce that, but using the source `INFOPLIST_FILE` first avoids the risk

relevant docs:

- [React Native source map setup](https://posthog.com/docs/error-tracking/upload-source-maps/react-native)
- [manual iOS symbol upload setup](https://posthog.com/docs/error-tracking/upload-source-maps/ios)
- [PostHog releases](https://posthog.com/docs/error-tracking/releases)
- [Expo app version management](https://docs.expo.dev/build-reference/app-versions/)

marandaneto:

Xcode can start the symbol upload phase while `dsymutil` is still producing the app dSYM. Uploading at that point can fail or send incomplete symbols. Declaring the dSYM as an Xcode input is not safe because it can create dependency cycles in apps with embedded extensions.

The upload script can also use `MARKETING_VERSION` and `CURRENT_PROJECT_VERSION` even when EAS remote versioning writes the shipped values directly to the source Info.plist.

This moves the native part of [PostHog/posthog-js#4602](https://github.com/PostHog/posthog-js/pull/4602) into the shared iOS upload script. It addresses [PostHog/posthog-js#4574](https://github.com/PostHog/posthog-js/issues/4574) for native iOS and React Native integrations.

The script now waits until the app executable and main dSYM have matching UUIDs and no process has the dSYM open for writing. It fails after 60 seconds by default so a release cannot succeed without its native symbols. `POSTHOG_DSYM_TIMEOUT` can change that timeout.

It also resolves literal, custom, and compound Xcode build settings from the source Info.plist. Existing Xcode version settings remain the fallback for missing, unresolved, or C-preprocessed plist values.

- Added shell tests for delayed and actively written dSYMs, timeout failure, CLI preflight ordering, Info.plist versions, custom and compound build settings, preprocessing fallback, and missing values.
- Built the `PostHogExampleWithSPM` sample for a Release iOS device with XcodeBuildMCP and posthog-cli 0.14.1. The real CLI uploaded the generated dSYM successfully.
- Rebuilt the sample with source plist values `$(APP_VERSION)` and `${BUILD_NUMBER}` while `MARKETING_VERSION` and `CURRENT_PROJECT_VERSION` remained `1.0` and `1`. The app, dSYM, and uploaded release all used `2.10.0+154`.
- Ran `make test`. All 763 SDK tests passed.
- Ran ShellCheck and Bash syntax validation for the upload scripts.

## :pencil: Checklist

## If releasing new changes

## 🤖 Agent context

Pi implemented and tested the change. The dSYM input-path approach was rejected because it can create Xcode dependency cycles. The upload script uses UUID, writer, and timeout checks instead. Shared Info.plist handling in posthog-cli is tracked in [PostHog/posthog#87775](https://github.com/PostHog/posthog/issues/87775).

### DIFF
--- a/.changeset/calm-dsyms-wait.md
+++ b/.changeset/calm-dsyms-wait.md
@@ -2,4 +2,4 @@
 "posthog-ios": patch
 ---
 
-Wait for the current app dSYM and use source Info.plist versions before uploading symbols.
+Upload symbols under the app version reported by Info.plist, including custom build settings. Wait for the current dSYM and fail after a configurable timeout instead of uploading invalid symbols.
--- a/.changeset/calm-dsyms-wait.md
+++ b/.changeset/calm-dsyms-wait.md
@@ -1,0 +1,5 @@
+---
+"posthog-ios": patch
+---
+
+Wait for the current app dSYM and use source Info.plist versions before uploading symbols.
--- a/build-tools/upload-symbols.sh
+++ b/build-tools/upload-symbols.sh
@@ -40,6 +40,38 @@ if [ -n "${CONFIGURATION}" ] && [ "${CONFIGURATION}" != "Release" ]; then
     exit 0
 fi
 
+# Xcode can start this phase before dsymutil finishes. Declaring the dSYM as an input can create
+# dependency cycles for apps with embedded extensions, so wait until the dSYM belongs to the current
+# executable instead. If it never becomes ready, skip the upload rather than sending stale symbols.
+if [ "${DEBUG_INFORMATION_FORMAT:-}" = "dwarf-with-dsym" ] && [ -n "${DWARF_DSYM_FOLDER_PATH:-}" ] && [ -n "${DWARF_DSYM_FILE_NAME:-}" ] && [ -n "${EXECUTABLE_NAME:-}" ] && [ -n "${TARGET_BUILD_DIR:-}" ] && [ -n "${EXECUTABLE_PATH:-}" ]; then
+    POSTHOG_MAIN_DWARF="${DWARF_DSYM_FOLDER_PATH}/${DWARF_DSYM_FILE_NAME}/Contents/Resources/DWARF/${EXECUTABLE_NAME}"
+    POSTHOG_APP_EXECUTABLE="${TARGET_BUILD_DIR}/${EXECUTABLE_PATH}"
+    POSTHOG_DSYM_ATTEMPT=1
+    POSTHOG_DSYM_MAX_ATTEMPTS=60
+    POSTHOG_DSYM_READY=0
+
+    while [ "$POSTHOG_DSYM_ATTEMPT" -le "$POSTHOG_DSYM_MAX_ATTEMPTS" ]; do
+        if [ -s "$POSTHOG_MAIN_DWARF" ] && [ -s "$POSTHOG_APP_EXECUTABLE" ]; then
+            POSTHOG_DSYM_UUIDS=$(xcrun dwarfdump --uuid "$POSTHOG_MAIN_DWARF" 2>/dev/null | awk '/^UUID: / {print $2}' | sort)
+            POSTHOG_APP_UUIDS=$(xcrun dwarfdump --uuid "$POSTHOG_APP_EXECUTABLE" 2>/dev/null | awk '/^UUID: / {print $2}' | sort)
+            if [ -n "$POSTHOG_DSYM_UUIDS" ] && [ "$POSTHOG_DSYM_UUIDS" = "$POSTHOG_APP_UUIDS" ]; then
+                POSTHOG_DSYM_READY=1
+                break
+            fi
+        fi
+
+        if [ "$POSTHOG_DSYM_ATTEMPT" -lt "$POSTHOG_DSYM_MAX_ATTEMPTS" ]; then
+            sleep 1
+        fi
+        POSTHOG_DSYM_ATTEMPT=$((POSTHOG_DSYM_ATTEMPT + 1))
+    done
+
+    if [ "$POSTHOG_DSYM_READY" -ne 1 ]; then
+        echo "warning: Main app dSYM was not ready after ${POSTHOG_DSYM_MAX_ATTEMPTS} attempts; skipping upload: $POSTHOG_MAIN_DWARF"
+        exit 0
+    fi
+fi
+
 # Validate environment
 if [ -z "${DWARF_DSYM_FOLDER_PATH}" ]; then
     echo "warning: DWARF_DSYM_FOLDER_PATH not set"
@@ -110,6 +142,34 @@ if [ "$LOWEST" != "$MIN_POSTHOG_CLI_VERSION" ]; then
     exit 1
 fi
 
+resolve_source_plist_value() {
+    local key="$1"
+    local plist_path="${INFOPLIST_FILE:-}"
+    local value
+
+    if [ -z "$plist_path" ]; then
+        return
+    fi
+    if [[ "$plist_path" != /* ]]; then
+        if [ -z "${SRCROOT:-}" ]; then
+            return
+        fi
+        plist_path="${SRCROOT}/${plist_path}"
+    fi
+    if [ ! -f "$plist_path" ]; then
+        return
+    fi
+
+    value=$(/usr/libexec/PlistBuddy -c "Print :${key}" "$plist_path" 2>/dev/null) || return
+    if [ -z "$value" ] || [[ "$value" == *"\$("* ]] || [[ "$value" == *"\${"* ]]; then
+        return
+    fi
+    printf '%s' "$value"
+}
+
+POSTHOG_RELEASE_VERSION=$(resolve_source_plist_value "CFBundleShortVersionString")
+POSTHOG_BUILD_VERSION=$(resolve_source_plist_value "CFBundleVersion")
+
 # Build CLI arguments as an array so paths with spaces are preserved.
 CLI_ARGS=(--directory "${DWARF_DSYM_FOLDER_PATH}")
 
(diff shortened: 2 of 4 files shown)


build-tools/upload-symbols.sh
```diff
@@ -110,6 +142,34 @@ if [ "$LOWEST" != "$MIN_POSTHOG_CLI_VERSION" ]; then
...
+    local value
+
+    if [ -z "$plist_path" ]; then
+        return
+    fi
+    if [[ "$plist_path" != /* ]]; then
+        if [ -z "${SRCROOT:-}" ]; then
+            return
+        fi
+        plist_path="${SRCROOT}/${plist_path}"
```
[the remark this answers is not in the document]
marandaneto: Thanks for flagging this. Xcode expands referenced build settings before exposing `INFOPLIST_FILE` to a run script. I verified the example target whose project setting is `$(TARGET_NAME)/Info.plist` with `show-build-settings`, which reports `PostHogExampleVisionOS/Info.plist`. The script therefore receives the resolved path, and no additional shell expansion is needed.

build-tools/upload-symbols.sh
```diff
@@ -40,6 +40,38 @@ if [ -n "${CONFIGURATION}" ] && [ "${CONFIGURATION}" != "Release" ]; then
     exit 0
 fi
 
+# Xcode can start this phase before dsymutil finishes. Declaring the dSYM as an input can create
```
turnipdabeets: [nit] This waits up to ~60s before we've even looked for posthog-cli, so a machine without the CLI installed errors out after a minute instead of instantly. Could the wait move below the CLI lookup? (It does need to stay above the "No dSYM bundles found" check further down — that one would exit early if dsymutil hasn't started yet.)

build-tools/upload-symbols.sh
```diff
@@ -40,6 +40,38 @@ if [ -n "${CONFIGURATION}" ] && [ "${CONFIGURATION}" != "Release" ]; then
...
+        fi
+
+        if [ "$POSTHOG_DSYM_ATTEMPT" -lt "$POSTHOG_DSYM_MAX_ATTEMPTS" ]; then
+            sleep 1
+        fi
+        POSTHOG_DSYM_ATTEMPT=$((POSTHOG_DSYM_ATTEMPT + 1))
+    done
+
+    if [ "$POSTHOG_DSYM_READY" -ne 1 ]; then
+        echo "warning: Main app dSYM was not ready after ${POSTHOG_DSYM_MAX_ATTEMPTS} attempts; skipping upload: $POSTHOG_MAIN_DWARF"
```
turnipdabeets: [question] Matching UUIDs is a much better signal than a fixed sleep or a declared input path. One consequence though: this used to fail the build loudly ("not recognized as a valid object file"), and now it's a warning plus `exit 0`. With 60 attempts at `sleep 1`, an app whose `dsymutil` runs longer than ~60s ships a release with no symbols at all, build green, one warning line in an archive log nobody reads. Should the cap be env-tunable (`POSTHOG_DSYM_TIMEOUT`), or is silently skipping the outcome we want here?

build-tools/upload-symbols.sh
```diff
@@ -110,6 +142,34 @@ if [ "$LOWEST" != "$MIN_POSTHOG_CLI_VERSION" ]; then
...
+            return
+        fi
+        plist_path="${SRCROOT}/${plist_path}"
+    fi
+    if [ ! -f "$plist_path" ]; then
+        return
+    fi
+
+    value=$(/usr/libexec/PlistBuddy -c "Print :${key}" "$plist_path" 2>/dev/null) || return
+    if [ -z "$value" ] || [[ "$value" == *"\$("* ]] || [[ "$value" == *"\${"* ]]; then
```
turnipdabeets, line 196: [suggestion] When the plist holds a reference we fall back specifically to `MARKETING_VERSION`, so a project whose plist says `$(APP_VERSION)` uploads under the wrong version — the same case greptile flagged on PostHog/posthog-js#4618. Xcode exports every build setting into this script's environment, so we can resolve it instead of guessing, which also folds the `$(MARKETING_VERSION)` case into the same rule:

```bash
case "$value" in
    '$('*')') name=${value#'$('}; value=$(printenv "${name%)}" 2>/dev/null) ;;
    '${'*'}')  name=${value#'${'}; value=$(printenv "${name%\}}" 2>/dev/null) ;;
esac
```

I checked the shape locally: `$(APP_VERSION)` -> `9.9.9`, `${MARKETING_VERSION}` -> `1.0`, and unset or compound values like `$(A)-$(B)` come back empty so the existing fallback still catches them.

.changeset/calm-dsyms-wait.md
```diff
@@ -0,0 +1,5 @@
+---
+"posthog-ios": patch
+---
+
+Wait for the current app dSYM and use source Info.plist versions before uploading symbols.
```
turnipdabeets: [suggestion] This describes the mechanism, but the part users need to know is that a build can now succeed *without* uploading symbols where it previously failed. Could we lead with that and the version fix? Something like: "Symbols now upload under the app's real version from Info.plist (fixes Expo/EAS builds landing on 1.0+1). A dSYM that isn't ready in time is now skipped with a warning instead of failing the build."

turnipdabeets (approved):

build-tools/upload-symbols.sh
marandaneto: makes sense

marandaneto: Good catch! Fixed in 8f5cb148c. The script now checks for posthog-cli before polling, while DWARF-only builds still exit before requiring the CLI.

build-tools/upload-symbols.sh
marandaneto: Good catch! Fixed in 8f5cb148c. The resolver now expands custom and compound Xcode build-setting references from the environment and retains the existing fallback when a value cannot be resolved.

.changeset/calm-dsyms-wait.md
marandaneto: Updated in 8f5cb148c. The changeset now leads with release attribution and describes the configurable timeout failure behavior.

build-tools/upload-symbols.sh
marandaneto: Fixed in 8f5cb148c. The timeout is configurable through `POSTHOG_DSYM_TIMEOUT`, defaults to 60 seconds, and now fails the build instead of allowing a release without symbols.

marandaneto: Addressed in c7b76a140. `POSTHOG_DSYM_TIMEOUT` is configurable with a 60-second default, and reaching the timeout now fails the build instead of silently shipping without symbols. fc534fcee also waits until no process has the dSYM open for writing.

build-tools/upload-symbols.sh
marandaneto: Good suggestion. Fixed in 8f5cb148c. Exact build-setting references are resolved from the Xcode environment, while compound substitutions use the processed Info.plist only when the source value cannot be resolved. The source plist remains preferred so a stale processed plist cannot override it.

.changeset/calm-dsyms-wait.md
marandaneto: Updated in c7b76a140. The changeset now leads with the release-attribution behavior and states that an unavailable dSYM fails after the configurable timeout rather than being uploaded.

build-tools/upload-symbols.sh
marandaneto: Good catch. Fixed in e11e6ac11. The script now checks whether the build emits dSYMs and locates posthog-cli before entering the readiness wait, while preserving the existing early skips for missing dSYM settings and DWARF-only builds.

marandaneto:
Tested this with two real Release iOS device builds of `PostHogExampleWithSPM` using posthog-cli 0.14.1.

1. Built the unmodified sample using its generated Info.plist. Xcode exported `DEBUG_INFORMATION_FORMAT=dwarf-with-dsym`, the readiness checks passed, and the CLI uploaded the dSYM successfully as `1.0+1`.
2. Rebuilt with source Info.plist values `$(APP_VERSION)` and `${BUILD_NUMBER}`, custom settings `2.10.0` and `154`, and unchanged Xcode defaults `1.0` and `1`. The built app, dSYM, CLI release attribution, and uploaded release all used `2.10.0+154`.

Both builds and real symbol uploads completed successfully. The temporary sample changes were removed afterward.

PR #777, opened 2026-08-23
fix(replay): mask React Native New Architecture (Fabric) text and image views

ivdezine:
### :bulb: Motivation and Context

Session replay masking matches view classes, and for React Native it only knows the old architecture classes `RCTTextView` and `RCTImageView`. React Native 0.82+ removed the old architecture, and on the New Architecture (Fabric) text renders as `RCTParagraphComponentView` and images as `RCTImageComponentView`. Neither matches, so on any current React Native app:

- Nothing rendered as text is masked. Replays record every label, balance, and merchant name in the clear even with `maskAllTextInputs` enabled; only real UIKit text inputs stay redacted.
- Images mostly get masked only through a side door: the Fabric image view embeds a `UIImageView` subclass that the generic `UIImageView` branch happens to catch. This PR makes the image view match by its component class directly.
- react-native-svg content (`RNSVGSvgView`) is not matched by anything.

We hit this in production hardening of a React Native 0.85 finance app: with both mask flags on, replays showed the full net worth screen legible. We could not find an existing issue or PR covering the Fabric classes.

The change mirrors the existing pattern exactly: three more `NSClassFromString` lookups, with `RCTParagraphComponentView` gated on `maskAllTextInputs` (like the legacy `RCTTextView`) and `RCTImageComponentView` plus `RNSVGSvgView` gated on `maskAllImages` (like the legacy `RCTImageView`). Apps without React Native are unaffected since the classes simply do not resolve.

If you would prefer to treat react-native-svg as out of scope for the core RN fix, happy to split `RNSVGSvgView` into a separate PR.

(posthog-android is not affected: it masks any `TextView`, and React Native's `ReactTextView` is a `TextView` subclass.)

### :green_heart: How did you test it?

- Reproduced the leak on device: React Native 0.85 (Expo SDK 56) app with session replay enabled, `maskAllTextInputs` and `maskAllImages` on; all rendered text was legible in recorded replays.
- Applied this change on top of 3.58.1, pinned the patched pod into the app, and validated on device (internal staging build 78, 2026-08-23) that recorded replays mask text, images, and SVG content: the same net worth flow that previously recorded fully legible now records fully redacted.
- Added `PostHogTests/PostHogReactNativeMaskingTest.swift`. The integration resolves the RN classes by name, so the suite registers plain `UIView` stubs under the real Objective-C runtime names and asserts `collectMaskableRects(in:)` returns each stub's rect under its matching flag only (and nothing with both flags off). No React Native dependency needed.
- Honest limitation: this machine cannot run `make lint` / `make test` / `make buildSdk` locally (macOS 12, Xcode 14.2, below the required 16.1), so we are relying on CI for the full suite. The patch and test mirror the surrounding code style closely.

### :pencil: Checklist

- [ ] I updated the docs if needed. (Happy to update the session replay privacy docs if you point me at the right page; the behavior change is "masking now also covers Fabric and react-native-svg views", which is what users already expect the flags to mean.)

### If releasing new changes

- [x] Changeset file added (`.changeset/fabric-replay-masking.md`, patch bump). Written by hand in the same format `pnpm changeset` generates, since this machine cannot run the toolchain.

### 🤖 Agent context

- Directed by @ivdezine (assign as PR assignee / DRI). Built with Claude Code (Claude Fable 5) during device validation and hardening of our app's session replay rollout.
- Decisions along the way: gated `RCTParagraphComponentView` on `maskAllTextInputs` rather than introducing a new flag, to match the legacy `RCTTextView` handling; kept `RNSVGSvgView` in scope because SVG frequently carries sensitive user content in RN apps (charts, QR codes), while offering to split it out; wrote the test with ObjC-named stub classes instead of adding a React Native dev dependency to the test bundle.
- The commit is authored by the human driver; agent assistance is disclosed in this section rather than in commit trailers.

### DIFF
--- a/PostHog/Replay/PostHogReplayIntegration.swift
+++ b/PostHog/Replay/PostHogReplayIntegration.swift
@@ -137,10 +137,13 @@
         private let reactNativeTextView: AnyClass? = NSClassFromString("RCTTextView")
         private let reactNativeImageView: AnyClass? = NSClassFromString("RCTImageView")
         // React Native New Architecture (Fabric) renders text and images with these component views
-        // instead of RCTTextView/RCTImageView; react-native-svg draws into RNSVGSvgView
+        // instead of RCTTextView/RCTImageView; react-native-svg draws its node hierarchy into RNSVGSvgView
         private let reactNativeParagraphView: AnyClass? = NSClassFromString("RCTParagraphComponentView")
         private let reactNativeImageComponentView: AnyClass? = NSClassFromString("RCTImageComponentView")
         private let reactNativeSvgView: AnyClass? = NSClassFromString("RNSVGSvgView")
+        private let reactNativeSvgRenderableView: AnyClass? = NSClassFromString("RNSVGRenderable")
+        private let reactNativeSvgGroupView: AnyClass? = NSClassFromString("RNSVGGroup")
+        private let reactNativeSvgTextView: AnyClass? = NSClassFromString("RNSVGText")
         // These are usually views that don't belong to the current process and are most likely sensitive
         private let systemSandboxedView: AnyClass? = NSClassFromString("_UIRemoteView")
 
@@ -834,6 +837,31 @@
             return wireframe
         }
 
+        private func reactNativeSvgContent(in view: UIView) -> (hasText: Bool, hasGraphic: Bool) {
+            var hasText = false
+            var hasGraphic = false
+
+            for subview in view.subviews {
+                if let reactNativeSvgTextView, subview.isKind(of: reactNativeSvgTextView) {
+                    hasText = true
+                    continue
+                }
+
+                if let reactNativeSvgRenderableView, subview.isKind(of: reactNativeSvgRenderableView) {
+                    let isGroup = reactNativeSvgGroupView.map { subview.isKind(of: $0) } ?? false
+                    if !isGroup {
+                        hasGraphic = true
+                    }
+                }
+
+                let nestedContent = reactNativeSvgContent(in: subview)
+                hasText = hasText || nestedContent.hasText
+                hasGraphic = hasGraphic || nestedContent.hasGraphic
+            }
+
+            return (hasText, hasGraphic)
+        }
+
         private func findMaskableWidgets(_ view: UIView, _ window: UIWindow, _ maskableWidgets: inout [CGRect], _ maskChildren: inout Bool) {
             // User explicitly marked this view (and its subviews) as non-maskable through `.postHogNoMask()` view modifier
             if view.postHogNoMask {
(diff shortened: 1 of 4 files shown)


PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -872,6 +884,20 @@
...
 
+            if let reactNativeImageComponentView = reactNativeImageComponentView {
+                if view.isKind(of: reactNativeImageComponentView), config?.sessionReplayConfig.maskAllImages == true {
+                    maskableWidgets.append(view.toAbsoluteRect(window))
+                    return
+                }
+            }
+
+            if let reactNativeSvgView = reactNativeSvgView {
+                if view.isKind(of: reactNativeSvgView), config?.sessionReplayConfig.maskAllImages == true {
```
marandaneto: blocking: SVG text remains visible under text masking

`RNSVGSvgView` is a mixed drawing surface: `react-native-svg` renders `<Text>` and `<TSpan>` directly rather than through `UILabel` or `RCTParagraphComponentView`. With `maskAllTextInputs = true` and `maskAllImages = false`, SVG text is therefore recorded unmasked, breaking the option’s all-text guarantee. Please mask this root when either text or image masking is enabled (or reliably detect SVG text nodes), and update the text-only test expectation.

PostHogTests/PostHogReactNativeMaskingTest.swift
```diff
@@ -0,0 +1,126 @@
...
+    @MainActor
+    struct PostHogReactNativeMaskingTest {
+        private typealias Sut = (sdk: PostHogSDK, integration: PostHogReplayIntegration)
+
+        /// An integration with a bound config, so the maskAllTextInputs/maskAllImages
+        /// heuristics are live (a bare `PostHogReplayIntegration()` has no config and
+        /// skips them). Resets the static install flag a prior replay suite may have
+        /// left set, so this install binds fresh instead of no-opping onto a stale one.
+        private func makeSut(maskText: Bool, maskImages: Bool) -> Sut {
+            let config = PostHogConfig(apiKey: "phc_reactNativeMaskingTest")
```
marandaneto: nit: Avoid the deprecated initializer

This initializer is deprecated and emits a warning, with removal planned for the next major version. Please use `PostHogConfig(projectToken: "phc_reactNativeMaskingTest")` instead.

marandaneto:
thanks @ivdezine

PostHog/Replay/PostHogReplayIntegration.swift
marandaneto: i will push some changes here, svg can be an image, a text, and both, so we need better heuristics here to avoid over masking, or leaking pii

ioannisj: Not sure if we should be masking vector graphics automatically at all tbh. People can manually mask any svg that may carry PII but in general I would consider svg as the icon engine for RN and leave it unmasked?

marandaneto: SVG itself is not PII; it is a rendering format. Its contents can display sensitive information, including:

 - Names, email addresses, account numbers, or other <Text> content
 - Embedded photos or signatures
 - QR codes and barcodes
 - User-specific charts, maps, or diagrams
 - Text converted into vector paths, which cannot be recognized as RNSVGText

https://github.com/PostHog/posthog-ios/pull/777/changes/78c2d35670e26dc6e84faa2c411f9b58d097254b helps a bit with that

marandaneto (approved):

ioannisj (approved):

PR #781, opened 2026-08-24
fix(replay): prevent fatal SIGPIPE when tearing down console log capture

### LINKED ISSUE #780: Fatal SIGPIPE in PostHogConsoleLogInterceptor.stopCapturing when backgrounding with sessionReplayConfig.captureLogs enabled
## Summary

With session replay console-log capture enabled, backgrounding the app can kill the process with a fatal `SIGPIPE` (signal 13) inside `PostHogConsoleLogInterceptor.stopCapturing`. It surfaces as a hard crash in our crash reporter, not a caught error.

## Version

- `posthog-ios` **3.48.2** (installed via SPM)
- Reproduced on iOS 26.3.1, iPhone15,3
- Reviewing the source, the relevant code appears unchanged through the current release, so we expect newer 3.x versions to be affected too.

## Stack trace

Symbolicated, most-recent frame last:

```
__CFNOTIFICATIONCENTER_IS_CALLING_OUT_TO_AN_OBSERVER__
ApplicationLifecyclePublisher.appDidEnterBackground   ApplicationLifecyclePublisher.swift:108
PostHogMulticastCallback.invoke                       PostHogMulticastCallback.swift:63
closure in PostHogReplayIntegration.start
PostHogReplayIntegration.pauseAllPlugins              PostHogReplayIntegration.swift:377
PostHogSessionReplayConsoleLogsPlugin.pause           PostHogSessionReplayConsoleLogsPlugin.swift:45
PostHogConsoleLogInterceptor.stopCapturing            PostHogConsoleLogInterceptor.swift:122
dup2
```

Crash type: `SIGPIPE: Signal 13, Code 0`, mechanism `signal`.

## What we think is happening

The interceptor swaps a pipe over the standard output descriptors to capture console output, and `stopCapturing` restores the saved originals with `dup2`. Tearing that down races with anything still writing to the pipe: once the read end is gone, the pending write gets `EPIPE`, and because `SIGPIPE` has its default disposition the process is terminated rather than the write simply failing.

The teardown runs from the `didEnterBackground` notification, which is exactly when the system is also flushing logging, so the window is easy to hit.

## Suggested fix

Any of these would prevent the process death:

1. Set `F_SETNOSIGPIPE` on the descriptors involved (via `fcntl`) so a broken pipe returns `EPIPE` instead of raising a signal. This is the most targeted fix and does not change global process state.
2. Ignore `SIGPIPE` around the swap and restore.
3. Drain and close the pipe in an order that guarantees no writer is left pointing at a closed read end before `dup2` runs.

Option 1 seems the most appropriate for a library, since installing a process-wide signal handler on a host app's behalf is a side effect an SDK ideally avoids.

## Impact and workaround

Low frequency but fatal when it happens, and it is entirely inside the SDK's own teardown path, so a host app cannot catch it.

Our workaround is to disable the feature:

```swift
config.sessionReplayConfig.captureLogs = false
```

That has stopped it for us, at the cost of losing console logs in replays. Happy to test a fix or provide more crash detail if useful.

turnipdabeets:

Fixes #780.

With `sessionReplayConfig.captureLogs` enabled, backgrounding the app could kill the process with a fatal `SIGPIPE`. The host app cannot catch it, so it lands in crash reporters as a hard crash inside the SDK's own teardown.

`stopCapturing()` closed `originalStdout` / `originalStderr` **before** detaching the `readabilityHandler`s that write to them, and those handlers captured the descriptor **by value**. So a handler still in flight on its dispatch queue could `write()` to a descriptor the main thread had just closed. A backgrounding app immediately recycles that descriptor number into a `URLSession` socket whose peer is gone — and that write raises `SIGPIPE`.

The reported top frame is `dup2` at `PostHogConsoleLogInterceptor.swift:122`, which is the main thread's *next* line; the signal comes from the reader queue.

**The fix suggested in the issue is not sufficient.** Measured on a standalone stress harness (verbatim copy of the interceptor's pipe machinery, driven by a logging thread plus socket churn):

| Variant | Runs that died with `SIGPIPE` |
| --- | --- |
| Unpatched | 10 / 10 |
| `F_SETNOSIGPIPE` only (the issue's preferred option) | 4 / 6 |
| Reorder teardown only | 6 / 6 |
| Shared lock (this PR) | 0 / 15 |

`F_SETNOSIGPIPE` can't help because the signal comes from the *recycled* descriptor, which isn't ours to flag. Reordering can't help because `readabilityHandler = nil` doesn't wait for a handler already running. The lock is the load-bearing part.

- **New regression tests** in `PostHogConsoleLogInterceptorTest.swift`. Against the unpatched interceptor, the concurrency test **crashes the test runner** and the `NOSIGPIPE` test fails; both pass with the fix. I re-ran that negative control after the tests were reworked, to confirm the repro still fires.
- **Full iOS simulator suite:** 836 tests / 130 suites pass.
- **`make test`:** 763 tests pass.
- **`make lint`:** clean, no new violations.

These tests are `#if os(iOS)`, so `make test` (swift test on macOS) compiles them away — they run only in the `test-ios-simulator` job. That makes the `project.pbxproj` registration load-bearing rather than cosmetic.

## :pencil: Checklist

## If releasing new changes

## Notes for reviewers

**Three invariants in this file are load-bearing** — a plausible-looking tidy-up breaks them, so they're commented in place:

1. `handle.availableData` must stay **inside** `fdLock` — it's what stops `closeFile()` racing an in-flight read (`EBADF` → uncatchable `NSFileHandleOperationException`).
2. `readabilityHandler = nil` must stay **outside** `fdLock` — folding it in deadlocks teardown against a running handler.
3. `unlock()` must stay **before** `processOutput` — the callback re-enters `postHog.capture`, and `fdLock` is a non-recursive `NSLock`.

**One accepted process-global side effect.** `F_SETNOSIGPIPE` is set on the file description shared with the app's real stdout/stderr and is never cleared, so after the first `startCapturing` any write to stdout/stderr that previously raised `SIGPIPE` returns `EPIPE` instead. Strictly safer (an uncatchable fatal signal becomes a checkable errno), no API change, `patch` bump still correct. Reverting it on teardown is deliberately **not** done: clearing the flag on a description an in-flight echo write may still be using is a path straight back to the crash class this PR closes.

**Known gaps, filed as follow-ups rather than widened into this PR:**

- `startCapturing` is not atomic — it releases the lock between `stopCapturing()` and `setupPipeRedirection()`. Two concurrent starts can permanently leak the process's only handle on the real console. Fixing it inside the interceptor alone would buy false safety, because `PostHogSessionReplayConsoleLogsPlugin.isActive` and `PostHogReplayIntegration.isEnabled` on the same path are unsynchronized too.
- `F_SETNOSIGPIPE` is never restored on teardown (see above).

## 🤖 Agent context

Claude Code (Opus 5). Reproduced the crash first with a standalone stress harness rather than reasoning from the stack trace alone, which is what surfaced that the issue's own preferred fix (`F_SETNOSIGPIPE`) doesn't actually work — the ablation table above is measured, not argued.

The change then went through an automated review loop (independent review → skeptical triage → implementation → verification). That loop caught a real regression in the first draft of this fix: the handler had started returning *before* `handle.availableData` when the descriptor sentinel was `-1`, so if `dup(STDOUT_FILENO)` ever failed, nothing would drain the pipe and every `print()` in the app would eventually block forever. Replacing the `-1` sentinel with a `fileHandleForReading === handle` identity check fixed that and closed a stale-handler hazard at the same time. The loop also added the positive-capture test, without which a wrong identity check would have silently disabled console capture with a fully green suite.

Deliberately rejected during triage: adding a `lifecycleLock` (only half-covers the problem, adds a new main-thread block during backgrounding) and reverting `F_SETNOSIGPIPE` on teardown (reintroduces the crash class). Both are written up as follow-ups above.

### DIFF
--- a/.changeset/console-logs-sigpipe.md
+++ b/.changeset/console-logs-sigpipe.md
@@ -1,0 +1,5 @@
+---
+"posthog-ios": patch
+---
+
+Fix a fatal SIGPIPE when session replay console log capture is torn down, for example when the app backgrounds with `sessionReplayConfig.captureLogs` enabled. Teardown closed the descriptors the pipe readers were still writing to, which could kill the process.
--- a/PostHog.xcodeproj/project.pbxproj
+++ b/PostHog.xcodeproj/project.pbxproj
@@ -299,6 +299,7 @@
 		DA52CACB2F24027500305F3D /* PostHogMulticastCallback.swift in Sources */ = {isa = PBXBuildFile; fileRef = DA52CACA2F24027500305F3D /* PostHogMulticastCallback.swift */; };
 		DA52CAD32F2402E000305F3D /* PostHogMulticastCallbackTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = DA52CAD22F2402E000305F3D /* PostHogMulticastCallbackTest.swift */; };
 		DA52CAEF2F24319800305F3D /* PostHogSessionReplayPluginTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = DA52CAEE2F24319800305F3D /* PostHogSessionReplayPluginTest.swift */; };
+		DA52CAEF2F2431980030AA01 /* PostHogConsoleLogInterceptorTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = DA52CAEE2F2431980030AA01 /* PostHogConsoleLogInterceptorTest.swift */; };
 		DA53DE762D3E29A900C38DCA /* fixture_remote_config.json in Resources */ = {isa = PBXBuildFile; fileRef = DA53DE702D3E299F00C38DCA /* fixture_remote_config.json */; };
 		DA53DE7E2D3E66AA00C38DCA /* PostHogRemoteConfigTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = DA53DE7D2D3E66A300C38DCA /* PostHogRemoteConfigTest.swift */; };
 		DA578E982D6768BC00B3A56C /* PostHogScreenViewIntegration.swift in Sources */ = {isa = PBXBuildFile; fileRef = DA578E972D6768B400B3A56C /* PostHogScreenViewIntegration.swift */; };
@@ -1078,6 +1079,7 @@
 		DA52CACA2F24027500305F3D /* PostHogMulticastCallback.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = PostHogMulticastCallback.swift; sourceTree = "<group>"; };
 		DA52CAD22F2402E000305F3D /* PostHogMulticastCallbackTest.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = PostHogMulticastCallbackTest.swift; sourceTree = "<group>"; };
 		DA52CAEE2F24319800305F3D /* PostHogSessionReplayPluginTest.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = PostHogSessionReplayPluginTest.swift; sourceTree = "<group>"; };
+		DA52CAEE2F2431980030AA01 /* PostHogConsoleLogInterceptorTest.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = PostHogConsoleLogInterceptorTest.swift; sourceTree = "<group>"; };
 		DA53DE702D3E299F00C38DCA /* fixture_remote_config.json */ = {isa = PBXFileReference; lastKnownFileType = text.json; path = fixture_remote_config.json; sourceTree = "<group>"; };
 		DA53DE7D2D3E66A300C38DCA /* PostHogRemoteConfigTest.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = PostHogRemoteConfigTest.swift; sourceTree = "<group>"; };
 		DA578E972D6768B400B3A56C /* PostHogScreenViewIntegration.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = PostHogScreenViewIntegration.swift; sourceTree = "<group>"; };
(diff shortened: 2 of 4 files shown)


dustinbyrne:
## Pull request risk assessment: PASS

I checked commit `a61b35853499c3e9d0acdfc10fadaf3a9246c0dd` against `main` at `668b69113f33deba0d4fd01de39990eccfe0638d`.

The merge base was `a437780deee17304deb4a0ee632c44ebc37aa04e`. The clean merge-base diff contains four files.

| Check | Result | Evidence |
| --- | --- | --- |
| The pull request fixes a real issue. | PASS | An identical iOS sandbox at the merge base exited with `SIGPIPE(13)` after 2.338 seconds. The head sandbox completed the same descriptor and lifecycle stress. The lock and duplicated echo descriptor protect the failing path in `PostHogConsoleLogInterceptor.swift:87-138,175-211`. |
| All non-bot authors are internal contributors. | PASS | `turnipdabeets` is the pull request author and the only commit author. There are no co-authors. GitHub returned `204` for PostHog organization membership. |
| The production change uses only private or internal logic. | PASS | The only production file is the internal `PostHogConsoleLogInterceptor`. The new enum, lock, and helper are private. |
| The public API does not change. | PASS | No public declaration, package entry point, or API snapshot changed. The exact-head Public API snapshot check passed. |
| Default data capture does not change. | PASS | Console capture still defaults to `false` in `PostHogSessionReplayConfig.swift:87-90`. Identical base and head default sandboxes emitted the same event sequence and the same normal-capture property schema. |
| The example application works without an observed regression. | PASS | I built and ran an isolated app with Xcode 26.6 on an iPhone 16 Pro iOS 18.6 simulator. The app survived 400 background and active notifications with concurrent logging and socket descriptor reuse. It remained active and emitted a normal capture event plus 162 snapshot events. The full iOS suite also passed 836 tests in 130 suites. |
| The changeset has the correct scope and release type. | PASS | `.changeset/console-logs-sigpipe.md` names only `posthog-ios`, uses `patch`, and describes this fix. |
| The change is not a major breaking change. | PASS | No supported API, persisted schema, package path, or default capture behavior changes. Consumers do not need a migration. Opt-in console capture now retains `NOSIGPIPE` after teardown, but this does not affect the default configuration. |

All checks passed. This pull request meets the low-risk rules.

### Required action

None.

dustinbyrne (approved):
Risk assessment: PASS.

Details: https://github.com/PostHog/posthog-ios/pull/781#issuecomment-5401281534

PR #782, opened 2026-08-25
fix: use posthog-cli Info.plist support for symbol uploads

marandaneto:

`upload-symbols.sh` resolves release metadata from the source Info.plist, including compound Xcode build-setting references that posthog-cli does not resolve itself. posthog-cli 0.15.1 added safe `--info-plist` fallback support, so it can fill metadata that the script cannot resolve.

Starting with 0.15.1, the script passes a valid source Info.plist alongside its explicit release arguments. Explicit values preserve compound build-setting resolution and `PRODUCT_BUNDLE_IDENTIFIER` behavior. Older CLI versions keep the existing arguments, and malformed or preprocessed plists fall back without failing the Release build.

- Ran `make testUploadSymbols`.
- Ran `make lint`.
- Ran Bash syntax checks and ShellCheck.
- Built the example app in Release mode and ran the upload script with real posthog-cli versions 0.14.1, 0.15.0, and 0.15.1 in dry-run mode.
- Verified compound build settings and malformed plist fallback with the real posthog-cli 0.15.1 binary.

## :pencil: Checklist

## If releasing new changes

## 🤖 Agent context

Implemented with the Pi coding agent. XcodeBuildMCP was used for the Release example builds. The version cutoff, explicit metadata precedence, and fallback behavior were verified with locally installed posthog-cli binaries.

### DIFF
--- a/build-tools/upload-symbols.sh
+++ b/build-tools/upload-symbols.sh
@@ -159,9 +159,29 @@ if [ "$LOWEST" != "$MIN_POSTHOG_CLI_VERSION" ]; then
     exit 1
 fi
 
+INFO_PLIST_MIN_POSTHOG_CLI_VERSION="0.15.1"
+LOWEST=$(printf '%s\n%s\n' "$INFO_PLIST_MIN_POSTHOG_CLI_VERSION" "$PH_CLI_VERSION" | sort -t. -k1,1n -k2,2n -k3,3n | head -n1)
+POSTHOG_CLI_SUPPORTS_INFO_PLIST=0
+if [ "$LOWEST" = "$INFO_PLIST_MIN_POSTHOG_CLI_VERSION" ]; then
+    POSTHOG_CLI_SUPPORTS_INFO_PLIST=1
+fi
+
+POSTHOG_INFO_PLIST_PATH=""
+# Bare C preprocessor macros cannot be expanded safely, and the product plist may belong to a
+# previous build. Preserve the Xcode-setting fallback for preprocessed plists.
+if [ "${INFOPLIST_PREPROCESS:-}" != "YES" ] && [ -n "${INFOPLIST_FILE:-}" ]; then
+    if [[ "$INFOPLIST_FILE" = /* ]]; then
+        POSTHOG_INFO_PLIST_PATH="$INFOPLIST_FILE"
+    elif [ -n "${SRCROOT:-}" ]; then
+        POSTHOG_INFO_PLIST_PATH="${SRCROOT}/${INFOPLIST_FILE}"
+    fi
+    if [ ! -f "$POSTHOG_INFO_PLIST_PATH" ]; then
+        POSTHOG_INFO_PLIST_PATH=""
+    fi
+fi
+
 resolve_source_plist_value() {
     local key="$1"
-    local plist_path="${INFOPLIST_FILE:-}"
     local value
     local token
     local name
@@ -169,22 +189,8 @@ resolve_source_plist_value() {
     local prefix
     local suffix
 
-    # Bare C preprocessor macros cannot be expanded safely here, and the product plist may belong to
-    # a previous build. Preserve the existing Xcode-setting fallback for preprocessed plists.
-    if [ "${INFOPLIST_PREPROCESS:-}" = "YES" ] || [ -z "$plist_path" ]; then
-        return
-    fi
-    if [[ "$plist_path" != /* ]]; then
-        if [ -z "${SRCROOT:-}" ]; then
-            return
-        fi
-        plist_path="${SRCROOT}/${plist_path}"
-    fi
-    if [ ! -f "$plist_path" ]; then
-        return
-    fi
-
-    value=$(/usr/libexec/PlistBuddy -c "Print :${key}" "$plist_path" 2>/dev/null) || return
+    [ -n "$POSTHOG_INFO_PLIST_PATH" ] || return
+    value=$(/usr/libexec/PlistBuddy -c "Print :${key}" "$POSTHOG_INFO_PLIST_PATH" 2>/dev/null) || return
     while [[ "$value" =~ (\$\(([A-Za-z_][A-Za-z0-9_]*)\)|\$\{([A-Za-z_][A-Za-z0-9_]*)\}) ]]; do
         token=${BASH_REMATCH[1]}
         name=${BASH_REMATCH[2]:-${BASH_REMATCH[3]}}
@@ -199,8 +205,12 @@ resolve_source_plist_value() {
     printf '%s' "$value"
 }
 
-POSTHOG_RELEASE_VERSION=$(resolve_source_plist_value "CFBundleShortVersionString")
-POSTHOG_BUILD_VERSION=$(resolve_source_plist_value "CFBundleVersion")
+POSTHOG_RELEASE_VERSION=""
+POSTHOG_BUILD_VERSION=""
+if [ "$POSTHOG_CLI_SUPPORTS_INFO_PLIST" -ne 1 ]; then
+    POSTHOG_RELEASE_VERSION=$(resolve_source_plist_value "CFBundleShortVersionString")
+    POSTHOG_BUILD_VERSION=$(resolve_source_plist_value "CFBundleVersion")
+fi
 
 # Build CLI arguments as an array so paths with spaces are preserved.
 CLI_ARGS=(--directory "${DWARF_DSYM_FOLDER_PATH}")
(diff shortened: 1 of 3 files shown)


marandaneto:
I verified both sides of the version cutoff with real posthog-cli binaries and a real Release build of `PostHogExample`.

For each CLI version, I:

1. Built `PostHogExample` in Release mode with `DEBUG_INFORMATION_FORMAT=dwarf-with-dsym`.
2. Confirmed the app binary and `PostHogExample.app.dSYM` were produced.
3. Ran `build-tools/upload-symbols.sh` against that dSYM with `POSTHOG_CLI_DRY_RUN=true`.
4. Captured the arguments passed by the script to the real CLI. Dry-run mode did not use credentials or contact PostHog.

### posthog-cli 0.14.1 and 0.15.0

The script used the existing explicit release arguments:

```text
--release-name com.posthog.PostHogExample
--release-version 4.0
--build 10
```

`--info-plist` was not passed, and both CLI dry-runs succeeded.

### posthog-cli 0.15.1

The script passes the valid source plist alongside explicit release values:

```text
--info-plist <absolute path to Info.plist>
--release-name com.posthog.PostHogExample
--release-version 4.0
--build 10
```

The explicit values preserve compound Xcode build-setting resolution and the configured bundle identifier. The CLI uses the plist only to fill missing values.

After review feedback, I also verified the updated script directly with the real 0.15.1 binary:

- A plist containing `$(VERSION_MAJOR).$(VERSION_MINOR)` and `$(BUILD_PREFIX)$(BUILD_NUMBER)` passed `--info-plist` plus resolved values `2.10.0` and `154`.
- A malformed plist omitted `--info-plist`, retained Xcode fallback values `1.0` and `1`, and completed successfully.

Automated coverage runs literal, custom, compound, preprocessed, unresolved, and malformed plist cases across the relevant legacy and 0.15.1 paths. `make testUploadSymbols` passes.

### Final sample rebuild after the review changes

I rebuilt `PostHogExample` from commit `2f5107848` in Release mode for the iOS Simulator with `DEBUG_INFORMATION_FORMAT=dwarf-with-dsym`. The resulting app and dSYM were universal Mach-O artifacts containing arm64 and x86_64 slices.

Running the updated upload script against those artifacts with the real posthog-cli 0.15.1 binary and `POSTHOG_CLI_DRY_RUN=true` succeeded. It passed:

```text
--info-plist <absolute source Info.plist path>
--release-name com.posthog.PostHogExample
--release-version 4.0
--build 10
```

The built app's processed Info.plist also contained `com.posthog.PostHogExample`, `4.0`, and `10`, matching the explicit upload metadata. No credentials were used and nothing was uploaded.

build-tools/upload-symbols.sh
```diff
@@ -199,8 +205,12 @@ resolve_source_plist_value() {
     printf '%s' "$value"
 }
 
-POSTHOG_RELEASE_VERSION=$(resolve_source_plist_value "CFBundleShortVersionString")
-POSTHOG_BUILD_VERSION=$(resolve_source_plist_value "CFBundleVersion")
+POSTHOG_RELEASE_VERSION=""
+POSTHOG_BUILD_VERSION=""
+if [ "$POSTHOG_CLI_SUPPORTS_INFO_PLIST" -ne 1 ]; then
```
turnipdabeets: With the CLI >= 0.15.1 we stop resolving here, but 0.15.1's resolver only handles a value that's entirely one `$(VAR)`/`${VAR}`, or a literal - `resolve_build_setting_value` in `cli/src/utils/xcode.rs` returns `None` for anything compound and falls through to `MARKETING_VERSION`/`CURRENT_PROJECT_VERSION`. Confirmed against the real 0.15.1 binary with the fixture from `test_resolves_compound_build_settings`: `--info-plist` gives `Release version: 1.0 / Build: 1`, today's args give `2.10.0 / 154`. Nothing fails - the symbols just bind to a release the SDK never reports. I think we need to keep resolving on the new path and pass `--info-plist` alongside the explicit values: `resolve_info_plist` uses `.or()`, so explicit wins and the plist only fills the gaps.

build-tools/upload-symbols.sh
```diff
@@ -210,20 +220,25 @@ if [ -n "${DWARF_DSYM_FILE_NAME}" ]; then
...
-fi
-if [ -n "$POSTHOG_BUILD_VERSION" ]; then
-    CLI_ARGS+=(--build "$POSTHOG_BUILD_VERSION")
-elif [ -n "${CURRENT_PROJECT_VERSION}" ]; then
-    CLI_ARGS+=(--build "${CURRENT_PROJECT_VERSION}")
+# Prefer values from the source Info.plist. EAS remote versioning can update these values without
+# changing MARKETING_VERSION or CURRENT_PROJECT_VERSION. Newer CLIs resolve the plist and its Xcode
+# build-setting references directly; older CLIs retain the script's legacy resolution.
+if [ "$POSTHOG_CLI_SUPPORTS_INFO_PLIST" -eq 1 ] && [ -n "$POSTHOG_INFO_PLIST_PATH" ]; then
+    CLI_ARGS+=(--info-plist "$POSTHOG_INFO_PLIST_PATH")
```
turnipdabeets, line 225: [question] Dropping `--release-name` hands the release name to the plist's `CFBundleIdentifier`, and `resolve_release_fields` only falls back to `PRODUCT_BUNDLE_IDENTIFIER` when the plist value is absent or unresolvable. Checked with the 0.15.1 binary: a plist literal `com.acme.app` now wins over build setting `com.acme.app.staging`, where the build setting always did before. No difference when the plist uses `$(PRODUCT_BUNDLE_IDENTIFIER)` - is the hardcoded-literal case one we care about?

build-tools/upload-symbols.sh
```diff
@@ -159,32 +159,38 @@ if [ "$LOWEST" != "$MIN_POSTHOG_CLI_VERSION" ]; then
...
+POSTHOG_INFO_PLIST_PATH=""
+# Bare C preprocessor macros cannot be expanded safely, and the product plist may belong to a
+# previous build. Preserve the Xcode-setting fallback for preprocessed plists.
+if [ "${INFOPLIST_PREPROCESS:-}" != "YES" ] && [ -n "${INFOPLIST_FILE:-}" ]; then
+    if [[ "$INFOPLIST_FILE" = /* ]]; then
+        POSTHOG_INFO_PLIST_PATH="$INFOPLIST_FILE"
+    elif [ -n "${SRCROOT:-}" ]; then
+        POSTHOG_INFO_PLIST_PATH="${SRCROOT}/${INFOPLIST_FILE}"
+    fi
+    if [ ! -f "$POSTHOG_INFO_PLIST_PATH" ]; then
```
turnipdabeets: [question] We check the plist exists but not that it parses, and the CLI's `PlistInfo::from_plist` is fatal - an unparseable plist exits 1 and fails the Release build (checked against 0.15.1: `Oops! Failed to parse Info.plist`). The old path degraded to the Xcode settings via `PlistBuddy ... 2>/dev/null || return`. Worth keeping that soft-fail, or is "a plist the build accepted but the CLI can't read" not a real case?

build-tools/upload-symbols.test.sh
```diff
@@ -45,6 +53,7 @@ create_fixture() {
     TEST_BUILD_PREFIX=""
     TEST_INFOPLIST_PREPROCESS="NO"
     TEST_INFOPLIST_PATH=""
+    TEST_CLI_VERSION="0.10.0"
```
turnipdabeets, line 56: [suggestion] With `0.10.0` as the default, all four resolution tests now cover only the legacy branch while the new one has a single happy-path test - that's why the suite stays green through the compound-reference change. Could we run `test_resolves_compound_build_settings` and friends over both `0.10.0` and `0.15.1`, asserting the resolved values rather than just flag presence?

build-tools/upload-symbols.sh
```diff
@@ -159,32 +159,38 @@ if [ "$LOWEST" != "$MIN_POSTHOG_CLI_VERSION" ]; then
     exit 1
 fi
 
+INFO_PLIST_MIN_POSTHOG_CLI_VERSION="0.15.1"
```
turnipdabeets, line 163: [nit] `--info-plist` actually landed in 0.15.0 - 0.15.1 is the right cutoff because 0.15.0 would `bail!` on an unresolved plist and fail the build (fixed in PostHog/posthog#87890). Worth a one-line comment so nobody "corrects" this down to 0.15.0 later.

marandaneto: Added in 2f5107848. The comment now records that the flag landed in 0.15.0 but unresolved plist fields remained fatal until 0.15.1.

build-tools/upload-symbols.sh
marandaneto: Agreed that we should preserve this behavior. Fixed in 2f5107848 by continuing to pass `PRODUCT_BUNDLE_IDENTIFIER` as `--release-name`; the explicit value takes precedence over a literal plist identifier.

build-tools/upload-symbols.sh
marandaneto: Good catch. Fixed in 2f5107848. The script now resolves plist values on both paths and passes those explicit values alongside `--info-plist`, so compound references remain `2.10.0` and `154` while the CLI fills only missing fields.

build-tools/upload-symbols.test.sh
marandaneto: Added in 2f5107848. Custom, preprocessed, compound, and unresolved plist cases now run against both 0.10.0 and 0.15.1, and the supported-CLI test asserts both `--info-plist` and the resolved explicit values. I also added a malformed plist regression test.

build-tools/upload-symbols.sh
marandaneto: Yes, keeping the soft failure is safer. Fixed in 2f5107848 by validating the source plist with PlistBuddy before passing `--info-plist`. A malformed plist now omits that flag and uses the explicit Xcode fallback values instead of failing the Release build.

ioannisj (approved):

PR #786, opened 2026-08-28
fix(replay): keep session replay masks aligned with the captured frame

ioannisj:

Session replay masks drift off the content they cover while a screen scrolls or animates. The mask rects get measured at one instant and the pixels get rendered at another, so on a fast scroll the redaction lands behind the text it's meant to hide.

Two things cause it:

- Mask geometry came from the model layer (`toAbsoluteRect`). During a Core Animation animation the model layer sits at its destination while the screenshot renders the presentation layer's in-flight position.
- `drawHierarchy` lags the render server, so even correct rects can sit ahead of the pixels that actually ship.

The base fix is `toPresentationRect`: mask geometry now reads the presentation tree, the same source the screenshot renders from. Wireframe geometry stays on `toAbsoluteRect`, it has no pixels to agree with.

On top of that there are two mechanisms, both driven by the same measurement. Before each capture `scheduleSettledCapture` samples mask geometry, waits 25ms (roughly one display-pipeline depth), samples again, and bands the result by the largest per-mask displacement. All of it lives in `PostHog/Replay/PostHogMaskSettleAnalysis.swift`.

### 1. Drop to the presentation-tree renderer on fast scrolling

The band picks the renderer:

- Still, at or under 80pt/s (2pt across the window): rects used as collected, full `drawHierarchy`.
- Drift, at or under 2000pt/s (50pt): still `drawHierarchy`, with the masks inflated as described below.
- Motion, anything faster: `render(in:)` against the presentation tree, whose masks can't disagree with its own pixels.

Unreadable geometry fails closed into motion, so an owner that appeared or disappeared between the samples, a duplicate owner, a count mismatch, or a nil sample all take the aligned renderer rather than guessing. Pairing is by owner identity (`MaskedRegion`), not by index, because index pairing would compare unrelated rects and report a false settle whenever the set changed shape.

The cost of the motion band is real and it's in the changeset: `render(in:)` draws blur, video and Metal flat, and it ignores `layer.mask`, `maskView` and Core Animation `filters`. Anything that must never appear in a recording needs `postHogMask()`.

### 2. Inflate masks across the two samples

In the drift band the frame keeps full fidelity and the masks absorb the movement instead. Each mask becomes the union of four rects: where it was, where it is, a lead of 1/16 of the measured displacement past the newer sample, and the older sample pushed back by the full displacement. That last one covers a position that was never sampled, since the display pipeline is deeper than a single settle window and the pixels on screen trail the older sample.

`screenshotModeBackgroundCapture` uses the same idea without a threshold. `performBracketedBackgroundCapture` puts the off-main render between two samples and masks the per-owner union, so the mask provably covers wherever the content sat while the render ran. An owner present in only one of those samples emits its rect as collected rather than dropping the frame, because a scroll that recycles cells changes the owner set constantly and dropping those frames would freeze the recording for the whole interaction.

### Defaults are deliberately conservative

The thresholds are set so that in practice only a fling changes anything. Measured on device across a 25ms window: slow drift moves about 1.5pt, a Core Animation animation about 4pt, a scroll fling 185pt or more. So ordinary movement stays on the renderer session replay has always used, and the flattened frames are limited to the part of a fling nobody is reading anyway. Every constant is tunable in one place with the reasoning next to it, so these are easy to move once feedback lands.

33 tests in `PostHogTests/PostHogMaskScenarioTest.swift`, in two groups:

- Scenario coverage: per-row and whole-row masks tracking new positions after scrolling, leaf masks hugging their labels, fail-closed nesting of `postHogMask()` and `postHogNoMask()`.
- Settle bands: both boundaries exactly and just past, a nil or unpairable sample falling to motion, a duplicate owner not trapping, permuted sample order still settling, and the inflated rects covering the swept region.
- Manual tests on simulator and physical device on local mock server inspecting the snapshots as they come through 

https://github.com/user-attachments/assets/70aa0f79-f9f3-49d3-8ff7-7157f5a70ba1

If you have a repro or a sample project that showed masks sliding off during a scroll, please run this branch against it and check the playback, particularly whether the flattened frames during a fling are an acceptable trade for aligned masks. That judgement is the part I'd most like a second opinion on.

## :pencil: Checklist

## If releasing new changes

## 🤖 Agent context

**DRI:** @ioannisj

### DIFF
--- a/.changeset/gentle-geese-argue.md
+++ b/.changeset/gentle-geese-argue.md
@@ -2,4 +2,4 @@
 "posthog-ios": patch
 ---
 
-Fix session replay masks drifting away from the content they cover while a screen scrolls or animates. During fast scrolling, frames are captured with a renderer that keeps masks aligned but draws blur, video and Metal content flat — use `postHogMask()` for anything that must never appear in a recording.
+Fix session replay masks drifting away from the content they cover while a screen scrolls or animates. During fast scrolling, frames are captured with a renderer that keeps masks aligned but draws blur, video and Metal content flat, and that also ignores `layer.mask`, `maskView` and CoreAnimation `filters` — content hidden by those alone is drawn in full in those frames. Use `postHogMask()` for anything that must never appear in a recording.
--- a/.changeset/gentle-geese-argue.md
+++ b/.changeset/gentle-geese-argue.md
@@ -1,0 +1,5 @@
+---
+"posthog-ios": patch
+---
+
+Fix session replay masks drifting away from the content they cover while a screen scrolls or animates. During fast scrolling, frames are captured with a renderer that keeps masks aligned but draws blur, video and Metal content flat — use `postHogMask()` for anything that must never appear in a recording.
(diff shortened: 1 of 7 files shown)


turnipdabeets (approved):
Report: https://claude.ai/code/artifact/f95dc420-13bb-4657-b912-8c9cd651cc2c

turnipdabeets (approved):

PR #787, opened 2026-08-31
feat: honor ph-no-mask accessibility token in session replay masking

olalekantakinola:
## Problem

On React Native, session replay masks every static `Text` when `maskAllTextInputs` is enabled: both `RCTTextView` (Paper) and `RCTParagraphComponentView` (Fabric) are swept under that flag in `findMaskableWidgets`. Disabling the flag is not a real alternative (it unmasks genuine inputs, and produces fully black recordings per #40006), so RN recordings end up as unreadable wireframes with no practical opt-out: every heading, tab label, and button is a black bar.

The SDK already has an unmask escape hatch, `postHogNoMask`, but it is only reachable from SwiftUI/UIKit via the `.postHogNoMask()` modifier. React Native apps cannot set it.

## Change

Honor a `ph-no-mask` token on `accessibilityIdentifier` / `accessibilityLabel` in the `findMaskableWidgets` early-exit, mirroring the existing `ph-no-capture` token and reusing `containsAccessibilityToken`. React Native apps can then keep the conservative `maskAllTextInputs: true` default and selectively unmask known-safe chrome via `testID` (which maps to `accessibilityIdentifier` on iOS):

```jsx
<Text testID="screen-title-ph-no-mask">Quick Actions</Text>
```

The token carries the same precedence as the `.postHogNoMask()` modifier it mirrors (checked in the same early-exit, applies to the view and its subviews).

Includes a changeset (patch).

## Testing

Validated end-to-end in a production React Native 0.81 app (new architecture, `posthog-react-native` 4.66.2, `@posthog/react-native-plugin` 2.5.1, this SDK at 3.70.1 with the patch applied): a dashboard heading tagged with the token renders readable in the processed replay while every other text element stays masked as before. Happy to attach the before/after replay frames from that validation run.

Happy to add a unit test if you can point me at the preferred harness for `findMaskableWidgets` coverage.

## Related

- #40006 describes the React Native all-text-masked behavior this provides an escape hatch for.

### DIFF
--- a/.changeset/great-panthers-smile.md
+++ b/.changeset/great-panthers-smile.md
@@ -1,0 +1,5 @@
+---
+"PostHog": patch
+---
+
+Session replay: honor a `ph-no-mask` token on `accessibilityIdentifier` or `accessibilityLabel` to exclude a view (and its subviews) from masking, mirroring the existing `ph-no-capture` token. This gives platforms that cannot reach the `.postHogNoMask()` modifier, such as React Native (where `testID` maps to `accessibilityIdentifier`), a way to selectively unmask known-safe views while keeping `maskAllTextInputs` on. On React Native every `Text` currently masks under `maskAllTextInputs`, so recordings are unreadable wireframes with no opt-out; with this token, apps can keep the conservative default and unmask navigation chrome explicitly.
--- a/.changeset/great-panthers-smile.md
+++ b/.changeset/great-panthers-smile.md
@@ -1,5 +1,5 @@
 ---
-"PostHog": patch
+"posthog-ios": minor
 ---
 
-Session replay: honor a `ph-no-mask` token on `accessibilityIdentifier` or `accessibilityLabel` to exclude a view (and its subviews) from masking, mirroring the existing `ph-no-capture` token. This gives platforms that cannot reach the `.postHogNoMask()` modifier, such as React Native (where `testID` maps to `accessibilityIdentifier`), a way to selectively unmask known-safe views while keeping `maskAllTextInputs` on. On React Native every `Text` currently masks under `maskAllTextInputs`, so recordings are unreadable wireframes with no opt-out; with this token, apps can keep the conservative default and unmask navigation chrome explicitly.
+Session replay: honor a `ph-no-mask` token on `accessibilityIdentifier` to exclude a view (and its subviews) from masking. This gives platforms that cannot reach the `.postHogNoMask()` modifier, such as React Native (where `testID` maps to `accessibilityIdentifier`), a way to selectively unmask known-safe views while keeping `maskAllTextInputs` on. The token is deliberately not honored on `accessibilityLabel`, which can carry localized, server-provided, or user-derived content: unmasking is only reachable through the developer-controlled identifier channel.
--- a/.changeset/great-panthers-smile.md
+++ b/.changeset/great-panthers-smile.md
@@ -2,4 +2,4 @@
 "posthog-ios": minor
 ---
 
-Session replay: honor a `ph-no-mask` token on `accessibilityIdentifier` to exclude a view (and its subviews) from masking. This gives platforms that cannot reach the `.postHogNoMask()` modifier, such as React Native (where `testID` maps to `accessibilityIdentifier`), a way to selectively unmask known-safe views while keeping `maskAllTextInputs` on. The token is deliberately not honored on `accessibilityLabel`, which can carry localized, server-provided, or user-derived content: unmasking is only reachable through the developer-controlled identifier channel.
+Session replay: honor a `ph-no-mask` token on `accessibilityIdentifier` or `accessibilityLabel` to exclude a view (and its subviews) from masking. This gives platforms that cannot reach the `.postHogNoMask()` modifier, such as React Native (where `testID` maps to `accessibilityIdentifier`), a way to selectively unmask known-safe views while keeping `maskAllTextInputs` on.
--- a/PostHog/Replay/PostHogReplayIntegration.swift
+++ b/PostHog/Replay/PostHogReplayIntegration.swift
@@ -863,10 +863,8 @@
         }
 
         private func findMaskableWidgets(_ view: UIView, _ window: UIWindow, _ maskableWidgets: inout [MaskedRegion], _ maskChildren: inout Bool) {
-            // User explicitly marked this view (and its subviews) as non-maskable through
-            // the `.postHogNoMask()` view modifier or a `ph-no-mask` accessibilityIdentifier
-            // token. Checked first so an explicit unmask wins over the sensitive-type
-            // early-returns below, matching the modifier's precedence.
+            // Checked first so an explicit unmask wins over the sensitive-type early-returns
+            // below, matching the modifier's precedence.
             if view.isNoMask() {
                 return
             }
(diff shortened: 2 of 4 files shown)


PostHog/Replay/UIView+Util.swift
```diff
@@ -21,6 +21,15 @@
             containsAccessibilityToken("ph-no-capture")
         }
 
+        /// Whether this view (and its subviews) is explicitly marked as non-maskable,
+        /// via the `ph-no-mask` token on its `accessibilityIdentifier` or `accessibilityLabel`.
+        /// Mirrors the `ph-no-capture` token so platforms that cannot reach the
+        /// `.postHogNoMask()` modifier (e.g. React Native, via `testID`) can unmask
+        /// known-safe views while keeping conservative masking defaults.
+        func isNoMask() -> Bool {
+            containsAccessibilityToken("ph-no-mask")
```
marandaneto: blocking: Do not let accessibility content disable masking — `containsAccessibilityToken` also searches `accessibilityLabel`, which commonly contains localized, server-provided, or user-derived content. If that content includes `ph-no-mask`, the early return in `findMaskableWidgets` skips this view and its descendants, potentially exposing sensitive data in screenshots. Unlike the existing no-capture/rage-click tokens, this marker reduces privacy protection, so it should only be accepted through an explicit non-content channel such as `accessibilityIdentifier`.

marandaneto (commented):
Automated advisory code review.

PostHog/Replay/UIView+Util.swift
```diff
@@ -21,6 +21,15 @@
...
+        }
```
ioannisj, line 30: > Happy to add a unit test if you can point me at the preferred harness for findMaskableWidgets coverage.   

`PostHogRageClickIntegrationTest.swift:240-256` is the closest template. It covers the `ph-no-rageclick` token on both carriers, and line 244 uses a compound identifier (`checkout-quantity ph-no-rageclick`) to pin the substring match, which matters here since your own example is `testID="screen-title-ph-no-mask"`.

I think the important case to be tested is the subtree one: a sensitive `UITextField` nested under a `ph-no-mask` ancestor doesn't get collected. Worth adding that?

.changeset/great-panthers-smile.md
```diff
@@ -0,0 +1,5 @@
+---
+"PostHog": patch
```
ioannisj: package name is wrong here. Also, I'd consider this a minor since it's really adding a new functionality here, even though `patch` is also defensible

```suggestion
"posthog-ios": minor
```

PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -863,8 +863,9 @@
         }
 
         private func findMaskableWidgets(_ view: UIView, _ window: UIWindow, _ maskableWidgets: inout [MaskedRegion], _ maskChildren: inout Bool) {
-            // User explicitly marked this view (and its subviews) as non-maskable through `.postHogNoMask()` view modifier
-            if view.postHogNoMask {
+            // User explicitly marked this view (and its subviews) as non-maskable through the
+            // `.postHogNoMask()` view modifier or the `ph-no-mask` accessibility token
+            if view.postHogNoMask || view.isNoMask() {
```
ioannisj: @[nit] Small one: this is the first statement in `findMaskableWidgets` now, so every view in the hierarchy pays a `containsAccessibilityToken` search on every snapshot frame. Before, only the views falling through all the type checks did, via `isNoCapture()` at line 1025.

It's cheap, and it reuses the safe `super.accessibilityLabel` path rather than React Native's recursive one, so this isn't a real cost. But moving it below the `as?` checks, or bailing out early when both accessibility fields are nil, gets it back for nothing?

PostHog/Replay/UIView+Util.swift
```diff
@@ -21,6 +21,15 @@
             containsAccessibilityToken("ph-no-capture")
         }
 
+        /// Whether this view (and its subviews) is explicitly marked as non-maskable,
+        /// via the `ph-no-mask` token on its `accessibilityIdentifier` or `accessibilityLabel`.
+        /// Mirrors the `ph-no-capture` token so platforms that cannot reach the
+        /// `.postHogNoMask()` modifier (e.g. React Native, via `testID`) can unmask
+        /// known-safe views while keeping conservative masking defaults.
+        func isNoMask() -> Bool {
```
ioannisj, line 28: [nit] `isNoRageClick()` just below folds both sources into the helper (`postHogNoRageClick || containsAccessibilityToken(...)`). Folding `postHogNoMask` in here the same way would let line 868 reduce to `if view.isNoMask() { return }`: a line shorter, and no chance of a later call site checking one source and missing the other.

Nothing's inconsistent today, line 868 is the only reader of either. Make sense here?

PostHog/Replay/UIView+Util.swift
olalekantakinola: Fixed in 918ad0d: the token is now honored only on accessibilityIdentifier, and the doc comment states why the label channel is excluded (content-derived, and this marker reduces protection rather than adding it). React Native reaches the identifier channel via testID, so the motivating use case is unchanged. Added a test pinning that a ph-no-mask accessibilityLabel does NOT unmask.

PostHog/Replay/UIView+Util.swift
olalekantakinola: Added in 918ad0d, modeled on the rage-click template: a compound-identifier unmask ('screen-title ph-no-mask'), the subtree case you suggested (a sensitive UITextField nested under a tokened ancestor is not collected), and the label-channel negative from the blocking thread. All in PostHogReactNativeMaskingTest since the motivation is React Native.

.changeset/great-panthers-smile.md
olalekantakinola: Changed to posthog-ios: minor in 918ad0d.

PostHog/Replay/PostHogReplayIntegration.swift
olalekantakinola: Kept it as the first statement: it mirrors the .postHogNoMask() modifier's position, and moving it below the as? checks would invert precedence (a tokened UITextView would stay masked). The cost side is addressed by the identifier-only change from the blocking thread: the common path is now a single nil property check with no label walk, and the fold means this is the lone call site.

PostHog/Replay/UIView+Util.swift
olalekantakinola: Folded in 918ad0d: isNoMask() now includes postHogNoMask, and the call site reduces to `if view.isNoMask() { return }`.

PostHog/Replay/UIView+Util.swift
ioannisj: I actually went the other way here to honor the token on `accessibilityLabel` too, so it lines up with `ph-no-capture` and `ph-no-rageclick`, as well as posthog-android implementation.

The content worry mostly doesn't reach this path. `containsAccessibilityToken` reads `super.accessibilityLabel`, which skips subclass getters, so it only ever sees a label someone explicitly assigned. React Native's recursive label and `UILabel`'s derive-from-`text` never show up there, which puts it at the same trust level as `testID`.

What the split costs us is that `ph-no-mask` becomes the one token with a narrow carrier, and it fails silently: someone sets `accessibilityLabel="ph-no-mask"`, nothing happens, no warning. posthog-android honors it on both `tag` and `contentDescription`, and that's what our docs already tell people to do.

ioannisj (approved):

PR #791, opened 2026-09-01
chore: deprecate POSTHOG_NO_RELEASE_BIND in the dsym upload

ablaszkiewicz:
## Related PRs

Event mode is live today, and it is opt-in. These PRs settle where it stays and where it goes.

**Deprecating the mobile knobs.** The symbol id on these paths is already a content hash, so two releases collide only when they ship a byte-identical artifact. An ordinary release that changes code gets its own symbol set and never collides. The dSYM path also lost release attribution for embedded targets, because one upload covers every target while it creates one release. Review asked for deprecation instead of removal, so every knob stays accepted as a warned no-op.

- PostHog/posthog#92401
- PostHog/posthog-android#747
- PostHog/posthog-ios#791 ← this PR

**Making it the default.**

- PostHog/posthog#91823 (stacked on PostHog/posthog#92401)
- PostHog/posthog-js#4705

**React Native.** Scopes the mode to the Hermes upload and defaults it to event. That is the one path where two releases really do ship the same artifact.

- PostHog/posthog-js#4717

## Problem

- Event mode helps only when two releases ship a byte-identical binary. The symbol id is the Mach-O `LC_UUID`.
- The unbound path also lost release attribution for embedded targets.
- The upload covers every extension dSYM, but it creates one release, from `PRODUCT_BUNDLE_IDENTIFIER`.
- An extension reports its own bundle identifier as `$app_namespace`, so its crashes matched no release.
- A silent removal would leave an opted-in user with no signal that their dSYMs bind again.

## Changes

- A build that sets `POSTHOG_NO_RELEASE_BIND` sees a build-log warning naming the deprecation and the fix. The script ignores the variable and uploads symbol sets bound to the release it creates.
- `dsym upload` no longer receives `--no-release-bind`.
- The script header documents the variable as deprecated instead of dropping it.
- The posthog-cli floor returns to 0.7.7, or 0.7.12 with `POSTHOG_SKIP_ON_CONFLICT`.

- `bash build-tools/upload-symbols.test.sh` passes.
- A new case sets the variable and asserts the warning prints, the upload succeeds, and `--no-release-bind` is not forwarded. It catches the flag forwarding coming back, which would fail the Xcode phase against a CLI that removes the flag for real.

## 🤖 Agent context

Written with Claude Code (Opus 5); the warning added with Claude Code (Fable 5) per the review comment. Skills invoked: `/writing-pr-descriptions`, `/writing-tests`.

This replaces PostHog/posthog-ios#789. That PR also carried a fix for a zero-padded `CFBundleVersion`, which follows in a separate PR. PostHog/posthog#92401 keeps `dsym upload --no-release-bind` parseable as a hidden no-op, which covers a build that upgrades the CLI before this pod.

### DIFF
--- a/.changeset/remove-dsym-no-release-bind.md
+++ b/.changeset/remove-dsym-no-release-bind.md
@@ -1,0 +1,7 @@
+---
+'posthog-ios': minor
+---
+
+Remove `POSTHOG_NO_RELEASE_BIND` from the dSYM upload build phase. The script uploads symbol sets bound to the release it creates, which is what it did before the variable existed.
+
+Event mode only helps when two releases ship a byte-identical binary, because the symbol id is the Mach-O `LC_UUID`. It also cost release attribution for embedded targets: the upload covers every extension dSYM but creates one release, so an extension crash resolved no release once the binding was gone. The variable was experimental and undocumented, so it goes out directly.
--- a/.changeset/remove-dsym-no-release-bind.md
+++ b/.changeset/remove-dsym-no-release-bind.md
@@ -1,7 +1,0 @@
----
-'posthog-ios': minor
----
-
-Remove `POSTHOG_NO_RELEASE_BIND` from the dSYM upload build phase. The script uploads symbol sets bound to the release it creates, which is what it did before the variable existed.
-
-Event mode only helps when two releases ship a byte-identical binary, because the symbol id is the Mach-O `LC_UUID`. It also cost release attribution for embedded targets: the upload covers every extension dSYM but creates one release, so an extension crash resolved no release once the binding was gone. The variable was experimental and undocumented, so it goes out directly.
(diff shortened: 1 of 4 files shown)


marandaneto:
print a deprecate/no-op warning if POSTHOG_NO_RELEASE_BIND is set so people know about this after upgrading

dustinbyrne (approved):

.changeset/remove-dsym-no-release-bind.md
```diff
@@ -0,0 +1,7 @@
+---
+'posthog-ios': minor
```
dustinbyrne: > The variable was experimental and undocumented, so it goes out directly

double checking - is this behavior changing? if not could this be even be a patch?

ablaszkiewicz: changed to patch

ablaszkiewicz:
@marandaneto deprecated it instead of getting rid of it as you suggested

Issue #664, opened 2026-06-16
File-backed queue folders (events, replay, logs) are not excluded from iCloud backup


ioannisj:
## Summary

On iOS/macOS the SDK's file-backed queue folders — events, session replay, replay buffer, and logs — are **not excluded from iCloud/iTunes backup**. They live under `Application Support` (which is backed up by default), so a user's device backup includes the SDK's queued telemetry.

## Details

A grep for `isExcludedFromBackup` across the SDK finds only **one** producer that sets it: `PostHogStorage.setData(forKey:contents:)` (`PostHog/PostHogStorage.swift:299-301`), which covers the key-value `StorageKey` files only.

The file-backed queues go through `PostHogFileBackedQueue` and never set the attribute:
- `setup(_:)` creates the directory via `FileManager.createDirectory` — no resource values set.
- `add(_:)` writes each item via `contents.write(to:)` — no resource values set.

Affected storage keys / folders:
- `.queue` (events)
- `.replayQeueue` and the replay buffer
- `.logsQueue`

Base directory (`applicationSupportDirectoryURL()`):
- **iOS/macOS:** `.applicationSupportDirectory` → **backed up to iCloud/iTunes by default**.
- **tvOS:** `.cachesDirectory` → not backed up, so this is moot there.

## Impact

Recreatable/transient telemetry (potentially large session-replay snapshots and event backlogs) ends up in users' device backups. Apple's guidance is that recreatable data should not be backed up; large backlogs also bloat backups.

## Recommended fix

Exclude the **base per-bundle folder** once (e.g. `<AppSupport>/<bundleId>` and/or the per-token folder in `getAppFolderUrl`) by setting `isExcludedFromBackup = true` on that directory. Excluding the directory covers its entire subtree, which makes the existing per-file exclusion in `setData` and any per-folder exclusions redundant — one place, all queues + storage covered.

## Notes

Surfaced during review of the exception-steps PR (#656). A directory-level exclusion was initially added for the exception-steps store but reverted, since fixing only that one folder is inconsistent with the rest of the SDK — the correct fix is the SDK-wide base-folder exclusion tracked here.

PR #803, opened 2026-09-05
fix: exclude telemetry queues from iCloud backups

marandaneto:

File-backed telemetry queues in Application Support were eligible for iCloud backup because only individual key-value files were excluded. This addresses #664.

Storage initialization now excludes the per-project directory from backups, including existing installations. This covers event, replay, replay-buffer, and log queues without changing their contents or excluding unrelated data in the parent bundle or app-group directory. Attribute-write failures are logged without preventing storage initialization.

- Reproduced the bug before the fix: all four new/existing and standard/app-group regression cases failed on missing exclusion metadata. All four passed after the fix.
- Reran the related storage and queue suites through `make test`: 63 Swift Testing tests across seven suites passed. Standard Application Support was isolated and shared app-group fixture use was coordinated. The filtered run executed no legacy Quick/XCTest cases.
- `make format`, `make lint`, and `git diff --check` passed. Lint reports 27 existing warnings and no serious violations.
- Exact-HEAD branch autoreview against `origin/main` passed at `7dae52418c132f3fe0cf641fd99f12d15223c7a1` with no findings. Independent source review also passed.
- macOS SDK and test targets compiled through `make test`. Cross-platform `make buildSdk` validation, using lane-local build outputs, stopped before compilation because the Xcode project references a missing relative `../posthog-ios` package in this worktree layout. Other platforms, examples, and the full suite remain unvalidated. No project-path workaround was committed.
- Tests check filesystem exclusion metadata, not an actual device backup or restore. Existing project directories are updated when their storage initializes.

## :pencil: Checklist

- [x] I updated the docs if needed. No public API or documentation changes were needed.

## If releasing new changes

- [ ] Ran `pnpm changeset` to generate a changeset file. A patch changeset was added directly in `.changeset/quiet-queues-backup.md`.

## 🤖 Agent context

Implemented and prepared with Pi using repository inspection, make targets, Git, GitHub CLI, and the isolated autoreview helper. Local publication session: `510ebebc-a750-45a5-88ce-92821ccd3256` (no public session link).

The fix uses the existing per-project storage boundary rather than changing each queue writer. It preserves the parent directory and existing per-file exclusions. Human review is required before merge.

### DIFF
--- a/.changeset/quiet-queues-backup.md
+++ b/.changeset/quiet-queues-backup.md
@@ -1,0 +1,5 @@
+---
+"posthog-ios": patch
+---
+
+Exclude SDK storage folders from device backups, including existing event, session replay, replay buffer, and log queues.
--- a/PostHog/PostHogStorage.swift
+++ b/PostHog/PostHogStorage.swift
@@ -386,10 +386,20 @@ class PostHogStorage {
     }
 
     private static func getAppFolderUrl(from configuration: PostHogConfig) -> URL {
-        let apiDir = getBaseAppFolderUrl(from: configuration)
+        var apiDir = getBaseAppFolderUrl(from: configuration)
             .appendingPathComponent(configuration.projectToken)
 
         createDirectoryAtURLIfNeeded(url: apiDir)
+
+        // Exclude the entire SDK subtree, including queues and existing installations.
+        // The parent bundle/app-group folder may also contain unrelated application data.
+        do {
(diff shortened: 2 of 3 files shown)


dustinbyrne (approved):
The backup attribute is scoped to the existing project directory and does not remove local queue data or alter restart persistence. No conflict found with the applicable storage contracts; device-backup policy is not directly specified.

Issue #768, opened 2026-08-19
custom push notifications impl using @UIApplicationDelegateAdaptor broken after 3.69.0


wesleynw:
### Version

3.69.6

### Steps to Reproduce

`UIApplicationDelegate.didRegisterForRemoteNotificationsWithDeviceToken` never called

It looks like you recently merged a fix for UNUserNotificationCenterDelegate method swizzling never reaching the original implementation (#748), but it looks like this is still happening for applications using [`UIApplicationDelegateAdaptor`](https://developer.apple.com/documentation/swiftui/uiapplicationdelegateadaptor).

My objc and swizzling knowledge is very limited, but it looks like this other project implemented some kind of "swizzle forwarding": https://github.com/OneSignal/OneSignal-iOS-SDK/pull/1091

### Expected Result

`UIApplicationDelegate.didRegisterForRemoteNotificationsWithDeviceToken` is called

### Actual Result

`UIApplicationDelegate.didRegisterForRemoteNotificationsWithDeviceToken` is not called

marandaneto:
thanks @wesleynw 
Checking this out, sorry!

wesleynw:
thanks for the quick fix @marandaneto! I can confirm that pinning the library to the main branch w/ your latest commit fixes my issue.

PR #807, opened 2026-09-08
feat(surveys): collect partial responses and resume unfinished surveys

### LINKED ISSUE #447: Surveys: Support partial response collection
## 🚨 IMPORTANT
This issue is likely user-facing in the main PostHog app, see [`surveyVersionRequirements.ts`](https://github.com/PostHog/posthog/blob/master/frontend/src/scenes/surveys/surveyVersionRequirements.ts). If you delete or close this issue, be sure to update the version requirements list there.

## Summary
The posthog-ios SDK does not support partial response collection for surveys.

## Current State
- The `enable_partial_responses` field is not present in the Survey model
- There is a `// TODO: Partial responses` comment in `PostHogSurveyIntegration.swift:437` but no implementation
- Currently, responses are only sent when the survey is fully completed

## Expected Behavior
When `enable_partial_responses` is enabled on a survey:
1. The SDK should track responses incrementally as the user answers each question
2. When a user dismisses or abandons a survey without completing it, the partial responses should be captured
3. Events should include `$survey_partially_completed: true` property when responses exist but survey isn't complete
4. Question shuffling should be disabled when partial responses are enabled (responses are tracked by question ID)

## Reference Implementation
See posthog-js browser: `packages/browser/src/extensions/surveys/surveys-extension-utils.tsx`
- `_surveyHasResponses()` function checks if any responses exist
- `_buildSurveyEventProperties()` includes `$survey_partially_completed` property
- `getDisplayOrderQuestions()` disables shuffling when `enable_partial_responses` is true

## Tracking
This is tracked in the survey SDK feature parity issue: https://github.com/PostHog/posthog/issues/45658

_This issue was generated by Claude during a Surveys SDK audit._

lucasheriques:

Add partial responses and persistent resume to iOS surveys, moving toward **survey feature parity across all PostHog SDKs**, with [posthog-js](https://github.com/PostHog/posthog-js/blob/f1395b6ce91af909ae0fdf095886f251f569a891/packages/browser/src/extensions/surveys.tsx) as the reference. Closes #447. Auto-submit is separate in #808.

After submitting an answer, users can restart and resume at the next branching destination with saved answers and the same `$survey_submission_id`. With `enable_partial_responses=true`, each answer emits a cumulative `survey sent` event; false/absent emits only on completion. Resume works in both modes.

Key behavior:

- Completion, dismissal, reset, removed/ended surveys and incompatible questions clear progress. Unavailable configuration preserves it; a confirmed empty list clears it.
- Reset invalidates saved answers and stale callbacks across app-group processes, including when the anonymous ID is reused. Old attempts cannot overwrite newer ones.
- Historical question wording is retained. Completion uses the current answer language; dismissal uses the last saved answer language.
- Resuming bypasses seen/internal targeting checks; other targeting still applies. Custom renderers opt in with `supportsSurveyResume = true` and must then honor `initialQuestionIndex`; others start fresh without restored answers or submission IDs.

### Review path

1. [Resume test](https://github.com/PostHog/posthog-ios/blob/49ecdc180/PostHogTests/PostHogSurveyResumeTest.swift): fresh SDK, restored branch, cumulative answers and language attribution.
2. [Progress store](https://github.com/PostHog/posthog-ios/blob/49ecdc180/PostHog/Surveys/SurveyProgressStore.swift) and [integration](https://github.com/PostHog/posthog-ios/blob/49ecdc180/PostHog/Surveys/PostHogSurveyIntegration.swift): persistence, eligibility and event lifecycle.
3. [Storage](https://github.com/PostHog/posthog-ios/blob/49ecdc180/PostHog/PostHogStorage.swift) and [SDK](https://github.com/PostHog/posthog-ios/blob/49ecdc180/PostHog/PostHogSDK.swift): cross-process reset coordination and configured-delegate cleanup on close.

`make test` passes: 873 Swift Testing tests and 172 XCTest cases. Formatting, lint and the CodeScene check pass. A regression that failed before the opt-in fix now covers legacy, opted-out and opted-in delegates, including eligibility, fresh IDs and response payloads.

The public API snapshot and SwiftLint checks pass in CI. Local iOS platform builds are blocked by mismatched Xcode/CoreDevice components; platform build verification runs in CI.

Regression tests cover restart, branching, payloads, reset and stale callbacks. Restart tests recreate SDKs through the real delegate/controller; OS process death was not exercised.

## :pencil: Checklist

- [x] Regression tests and independent review completed.
- [x] Delegate documentation and changeset included.

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted). Codex implemented the changes with independent QA agents; human review is required.

### DIFF
--- a/PostHog/Surveys/PostHogSurveyIntegration.swift
+++ b/PostHog/Surveys/PostHogSurveyIntegration.swift
@@ -56,6 +56,8 @@
 
         private var activeSurveyLock = NSLock()
         private var activeSurvey: PostHogSurvey?
+        private var progressStore: SurveyProgressStore?
+        private var activeProgressWasPersisted = false
         private var activeSurveySubmissionId: String?
         private var activeSurveyLanguage: String?
         /// Language the survey was rendered with, frozen at show time and reused as `$survey_language` on
@@ -66,6 +68,7 @@
         /// keeping them in `$survey_language` rather than a later live re-translation.
         private var activeSurveyRenderedQuestionTranslations: [PostHogSurveyQuestionTranslation?]?
         private var activeSurveyResponses: [String: PostHogSurveyResponse] = [:] // keyed by question identifier
+        private var activeSurveyResponseLanguage: String?
         /// Question text as shown when each question was answered, keyed like `activeSurveyResponses`, so
         /// a later re-translation can't rewrite the language a question was answered in.
         private var activeSurveyResponseQuestionText: [String: String] = [:]
@@ -75,6 +78,7 @@
         func install(_ postHog: PostHogSDK) -> PostHogIntegrationInstallResult {
             installIfNeeded(using: Self.integrationInstallState) {
                 self.postHog = postHog
+                if let storage = postHog.storage { progressStore = SurveyProgressStore(storage: storage) }
                 start()
             }
         }
@@ -113,6 +117,7 @@
             didBecomeActiveToken = nil
             didLayoutViewToken = nil
             personPropertiesChangedToken = nil
+            clearActiveSurvey()
             unsubscribeFromRemoteConfigUpdates()
             #if os(iOS)
                 if #available(iOS 15.0, *) {
@@ -132,7 +137,7 @@
                 let matchingSurveys = surveys
                     .lazy
                     .filter { // 1. unseen surveys,
-                        !self.getSurveySeen(survey: $0)
+                        self.hasProgress($0) || !self.getSurveySeen(survey: $0)
                     }
                     .filter(\.isActive) // 2. that are active,
                     .filter { survey in // 3. and match display conditions,
@@ -146,12 +151,11 @@
                         self.hasWaitPeriodPassed(survey: survey)
                     }
                     .filter { survey in // 4. and match linked flags
-                        guard !survey.requiresFeatureFlagEvaluation || self.canEvaluateSurveyFeatureFlags else { return false }
                         let allKeys: [String?] = [
                             [survey.linkedFlagKey],
                             [survey.targetingFlagKey],
                             // we check internal targeting flags only if this survey cannot be activated repeatedly
-                            [survey.canActivateRepeatedly ? nil : survey.internalTargetingFlagKey],
+                            [survey.canActivateRepeatedly || self.hasProgress(survey) ? nil : survey.internalTargetingFlagKey],
                             survey.featureFlagKeys?.compactMap { kvp in
                                 kvp.key.isEmpty ? nil : kvp.value
                             } ?? [],
@@ -160,6 +164,7 @@
                         .compactMap { $0 }
                         .filter { !$0.isEmpty }
 
+                        guard allKeys.isEmpty || self.canEvaluateSurveyFeatureFlags else { return false }
                         // all keys must be enabled
                         return Set(allKeys)
                             .allSatisfy(self.isSurveyFeatureFlagEnabled)
@@ -168,7 +173,7 @@
                         survey.hasEvents ? self.isSurveyEventActivated(survey: survey) : true
                     }
 
-                callback(Array(matchingSurveys))
+                callback(Array(matchingSurveys).sorted { self.hasProgress($0) && !self.hasProgress($1) })
             }
         }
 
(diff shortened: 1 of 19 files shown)


PostHog/Surveys/PostHogSurveyIntegration.swift
```diff
@@ -546,114 +572,118 @@
...
+                guard !activeProgressWasPersisted || progressStore?.load(activeSurvey)?.submissionId == submissionId else {
+                    clearActiveSurveyLocked()
+                    return nil
+                }
 
-            let nextSurveyQuestion = PostHogNextSurveyQuestion(
-                questionIndex: nextIndex,
-                isSurveyCompleted: isCompleted
-            )
+                guard !activeSurveyCompleted, index >= 0, index == activeSurveyQuestionIndex, activeSurvey.questions.indices.contains(index) else { return nil }
```
turnipdabeets, line 593: [question] What happens here with a custom delegate that doesn't read `initialQuestionIndex`? If I'm reading it right, a resumed survey restores `activeSurveyQuestionIndex` to, say, 2, the delegate still shows question 0, and every answer fails this check and returns nil. posthog-flutter is one of those delegates. It accepts any posthog-ios `< 4.0.0`, so Flutter apps would pick this up and get stuck on the first question. Its Dart code also does `result as Map` on the nil, which throws.

Could we resume only for the built-in sheet, plus any delegate that opts in (e.g. an optional `supportsSurveyResume` on the protocol)? Everyone else would just start fresh, like today. That lets us keep this check, and Flutter can opt in once it actually starts at `initialQuestionIndex`.

lucasheriques: Yes, that's a compatibility bug: the SDK must not restore answers and the submission ID unless the renderer can also restore the question position.

I propose an optional `@objc var supportsSurveyResume: Bool { get }` on `PostHogSurveysDelegate`. Missing/false means unsupported; the built-in delegate returns true. Only an opted-in renderer gets restored progress and the corresponding resume eligibility. Other delegates start at question 0 with empty answers and a new submission ID, while partial-response events continue to work within that attempt.

Persistence and submission ownership stay in the SDK. Flutter can opt in once its bridge and UI honor the restored position. On Android I'd use a separate resume-capability interface so existing delegate implementations acquire no new requirements, and keep this independent of reset handling.

Does this API shape work for you? I'll keep the public API change pending confirmation, as required by CONTRIBUTING.md.

turnipdabeets: Yes, the optional `@objc var supportsSurveyResume: Bool { get }` works. `PostHogSurveysDelegate` is already `@objc`, so existing implementers don't change. A couple of things to go with it:

- The changeset line "Custom survey delegates should honor `initialQuestionIndex`" and the doc on `renderSurvey` should say this only applies once a delegate opts in.
- Could the new property's doc comment cover the default (`false`), what opting in means (start at `initialQuestionIndex`, answers and submission ID restored), and a short example?

I left a note on #768 suggesting Android handle reset internally the way iOS does, so the public API matches on both platforms.

lucasheriques: Implemented in e9a97be. The optional property defaults to unsupported when omitted; the built-in delegate opts in. Only opted-in renderers restore position, answers and submission ID or bypass seen/internal targeting for resume. The renderSurvey docs and changeset now make that conditional, and the new property's docs include the default, contract and example.

The regression covers legacy, false and true delegates: eligibility, fresh submission IDs, valid callbacks from question 0, and no old answers in a fresh completion. It failed before the fix; all 873 Swift Testing tests and 172 XCTest cases now pass, plus lint and CodeScene. Local iOS API generation is blocked by an Xcode/CoreDevice mismatch; the new snapshot entry was extracted from the compiled macOS symbol graph, with the iOS CI API check still required.

turnipdabeets (approved):

lucasheriques:
Pre-merge validation: all 14 required checks pass, including SDK/platform builds, tests and API validation; the current head is approved.

The non-blocking compliance report still has one failure: `retry_behavior.max_retries_respected` expected 4 requests and observed 6 on this PR. I checked the latest main run, which also fails that test (expected 4, observed 5). This remains an unresolved retry/compliance issue, so the green workflow should not be read as 45/45 compliance tests passing.

- PR run: https://github.com/PostHog/posthog-ios/actions/runs/35382371974
- Main baseline: https://github.com/PostHog/posthog-ios/actions/runs/35277788629

Issue #813, opened 2026-09-10
Investigate main-run-loop scheduling for session replay instead of the global layout swizzle


marandaneto:
## Summary

Investigate replacing session replay’s global `UIView.layoutSublayers(of:)` swizzle with a main-run-loop-based capture scheduler. This is an architectural proposal to reduce interference with UIKit layout and improve capture scheduling—not a confirmed fix for the production crash reported in #806.

## Current architecture and motivation

`ApplicationViewLayoutPublisher` exchanges the implementations of `UIView.layoutSublayers(of:)` and `ph_swizzled_layoutSublayers(of:)` when it gains subscribers. The wrapper forwards synchronously to UIKit, then notifies PostHog on main. Native replay subscribes when recording starts, independently of screenshot mode; Surveys also uses this shared publisher.

This puts SDK code on a very frequently used UIKit method and introduces global runtime-hook lifecycle and interoperability concerns.

Related investigation: #806. We reproduced the background thread-cleanup path `CA::Transaction::release_thread → layoutSublayers → NSISEngine optimize`, both with and without the PostHog swizzle. The original UIKit call still precedes #533’s notification thread guard. Removing the hook would remove this replay observation path from the stack, but would not itself repair an invalid background UIKit transaction or establish the customer’s production root cause.

A separate lifecycle investigation also confirmed unsynchronized access to `hasSwizzled` with Thread Sanitizer and demonstrated out-of-order subscriber-count notifications. Those findings should be tracked/fixed independently rather than requiring this larger redesign.

## Architecture to evaluate

- Register a `CFRunLoopObserver` on the **main run loop**, in common modes so tracking/scrolling remains covered.
- Observe late-cycle opportunities such as `.beforeWaiting` and `.exit`, with a high observer order. Treat these as opportunities to consider capture, not proof that a new frame is visible or that all Core Animation work has completed. Ordering relative to framework internals is not a public stability guarantee.
- Track processed run-loop activity to avoid duplicate consideration within a cycle. Run-loop activity alone is not evidence of a visual change.
- Preserve the existing rendering, masking, deduplication, throttling, and in-flight capture protections initially; change scheduling separately from rendering behavior.
- Keep live-view inspection/rendering on main. Background processing should operate on captured image data, not move UIKit layout off-main.
- Serialize observer lifecycle changes on main, using ownership tokens/generation checks so stale start/stop operations cannot affect a newer recording session. Avoid holding locks across main-queue dispatches.
[...]

Issue #806, opened 2026-09-08
Session Replay's layoutSublayers swizzle still crashes NSISEngine on background threads — #533's guard doesn't cover the forwarded original call


NavaneethP10:
### Summary

**What's causing it:** `ph_swizzled_layoutSublayers(of:)` unconditionally forwards to the original `UIView.layoutSublayers(of:)` implementation before any thread check runs. When CoreAnimation invokes `layoutSublayers(of:)` on a background thread during thread-local cleanup (`CA::Transaction::release_thread` — a documented, pre-existing OS behavior, not something the app triggers), that forwarded call reaches Apple's `NSISEngine` (Auto Layout) off the main thread, which throws `NSInternalInconsistencyException`. This happens whenever Session Replay is active, because it swizzles `layoutSublayers(of:)` on every `UIView` app-wide.

**Was it fixed? No.** [PR #533](https://github.com/PostHog/posthog-ios/pull/533) (3.48.1) added a `Thread.isMainThread` guard, but only around a *different* line — PostHog's own notification call — not around the forwarded original call where the exception actually originates (detailed below, with PostHog's own stack trace as evidence). We checked the exact file at the current `main` HEAD (`113b33e7a634919cd45e630ffa29660f57e0775a`, identical to the latest release, 3.71.6, published 2026-09-07) and the unconditional call is still there, unchanged since #533 merged in March 2026. No PR or commit has touched this file since.

### Version

3.69.2 (checked: this code is unchanged on `main` @ `113b33e7a634919cd45e630ffa29660f57e0775a` — i.e. current latest release 3.71.6 — the file is byte-identical to 3.69.2 for this code path)

### Reproducibility

We can't reproduce this on demand, on simulator or device, and don't have a minimal repro project to offer — please read this as a code-inspection finding backed by production telemetry, not a report of a crash we've reliably triggered ourselves.

What we do have:

[...]

PR #814, opened 2026-09-10
fix: make layout observation lifecycle thread-safe

marandaneto:

Converts this reproduction-only PR into a production lifecycle fix. The reproductions are now passing regression tests; no intentionally failing tests remain.

Related to #806, but these confirmed lifecycle defects do **not** establish the cause of the customer's production `NSISEngine` crash. The diagnostic warning from merged #815 is preserved in the new hook. The run-loop proposal in #813 remains separate work. This preserves synchronous UIKit forwarding on its original calling thread; it does not suppress background layout.

### Changes

- Eagerly initialize the callback publisher instead of racing Swift's lazy initialization.
- Serialize and coalesce subscriber-count observer delivery, reconciling the latest count without holding callback-state locks during observer execution. Concurrent and reentrant observers cannot overlap or apply an older count after the final current count.
- Bound subscriber-count delivery to 32 observer calls per batch. Queue pending reconciliation on a private serial queue while retaining notification ownership, so sustained churn cannot keep the initiating thread in an unbounded drain. This bounds call count, not the duration of arbitrary observer code.
- Replace exchange-based forwarding with hooks that capture immutable original IMPs. A layout dispatch that retained a hook before unsubscribe can still forward safely after uninstall, without recursive alias lookup.
- Cache hooks per original IMP and retain their block implementations for process lifetime, since in-flight dispatches and other swizzlers may retain those function pointers. Ordinary restarts reuse the same hook.
- Preserve an observed newer swizzler during teardown, but retire our ownership when the last subscriber leaves. On positive subscriber-count changes, check the actual current IMP and recover if another library removed or bypassed our hook. This also works when another subscriber, such as Surveys, remains active.
- Reactivate a restored cached PostHog IMP rather than wrapping ourselves. An opaque external hook may retain our old hook, so recovery can add one cached outer wrapper. The active token prevents duplicate notifications, and repeated restarts reuse a stable chain.
- Use dedicated test view classes so runtime regression fixtures do not interfere with the shared publisher or other suites.

An older library restoring UIKit can remove PostHog from the chain. Recovery now handles that sequence on the next subscriber-count notification rather than trusting stale installation state. It does not continuously monitor runtime replacement or coordinate arbitrary concurrent third-party writes. Forwarding implementations retained by other libraries must remain valid.

Reproduction-first validation confirmed stale count delivery, a `hasSwizzled` TSan race, multiple callback objects from concurrent first access, stack-overflowing forwarding after unsubscribe, and clobbering a newer swizzler before the fixes. A retained-old-hook regression also caught duplicate notifications during implementation.

| Validation | Result |
| --- | --- |
| Focused lifecycle suite with Thread Sanitizer | Pass, three repetitions/relaunches; includes 5,000 rounds of eight concurrent retained subscriptions |
| Reentrant and bounded-delivery regressions | Pass: 32 synchronous calls, deferred reconciliation to zero/nonzero counts, 257 calls across multiple batches, no overlapping observers |
| `make test` | Pass on `6f31b314c`: 799 Swift Testing tests, plus XCTest |
| `make testOniOSSimulator` | Pass on `6f31b314c`: 193 XCTest tests and 926 Swift Testing tests |
| Detached-hook recovery | Pass: older uninstall and newer bypass, each with/without another retained subscriber; 100 restart cycles verify exactly-once forwarding/notifications and stable cached IMP reuse |
| SDK targets via `make build` | Pass: iOS, macOS, Mac Catalyst, tvOS, watchOS, visionOS |
| `make format`, `make lint`, `git diff --check` | Pass |
| `make build` | SDK and platform examples passed; external-client example blocked by checkout/package identity mismatch described below |

Full build stops at `ExternalSDK-iOS`: the example expects package identity `posthog-ios`, while the isolated checkout is named `posthog-ios-issue-806-lifecycle`. Subsequent CocoaPods examples were not reached. No unrelated build configuration was changed.

Autoreview command: `$HOME/.pi/agent/skills/autoreview/scripts/autoreview --mode branch --base origin/main --prompt-file autoreview-scope.md`. One test-ordering finding was rejected after checking the actual code: this test uses only zero-interval **trailing** subscriptions, whose `invokeTrailing` directly enqueues on main before the continuation barrier. It does not traverse the background leading-throttle queue. The bundle-only reviewer could not see that unchanged implementation. No accepted review blockers remain.

The bounded-delivery follow-up (`5826de1b0`) passed isolated autoreview with no findings: `$HOME/.pi/agent/skills/autoreview/scripts/autoreview --mode commit --commit HEAD --prompt-file autoreview-bounded-scope.md`. Its tests first failed against the unbounded loop, then passed along with the existing lifecycle tests under Thread Sanitizer across three repetitions/relaunches.

### Latest main merge validation (`a4be3e52c`)

All 47 CI checks passed for this exact commit, including SDK/example builds, tests, replay masking snapshots, and compliance. The combined warning/lifecycle tests passed locally under TSan across three repetitions, and the full iOS suite passed (190 XCTest tests and 921 Swift Testing tests). Format and lint passed. Local macOS full-suite attempts had varying failures in unchanged feature-flag/event/queue timing tests; the queue test passed in isolation, and the full macOS CI job passed. No tests were weakened or disabled. The existing local external-client package-identity build limitation remains; its CI build passed.

### Current head (`6f31b314c`)

Recovery fix `7e043c190` addresses [the clarified review sequence](https://github.com/PostHog/posthog-ios/pull/814#discussion_r3981315753). Both detach sequences first failed with zero notifications. The retained-subscriber variants also failed on the partial zero-only fix before runtime reconciliation was added. No intentionally failing tests remain.

Merged latest `main` (`05bbc8d6e`, including v3.73.0) without conflicts in `6f31b314c`. The table above reflects this head: full SPM/iOS suites passed, combined layout and bounded/reentrant tests passed under TSan across three repetitions/relaunches, and format/lint passed. The local external-client build limitation is unchanged. CI is running for this exact head.

Final autoreview: `$HOME/.pi/agent/skills/autoreview/scripts/autoreview --mode branch --base origin/main --prompt-file autoreview-final-lifecycle-scope.md`. Clean, no actionable findings.

## :pencil: Checklist

- [x] I updated the docs if needed. Internal lifecycle behavior is documented above and in code.

## If releasing new changes

- [x] Added a patch changeset: `.changeset/stable-layout-lifecycle.md`.

## 🤖 Agent context

Manoel requested implementing the lifecycle fixes in this existing PR worktree and pushing to this PR. Pi used reproduction-first tests, file edits, make-based validation, Thread Sanitizer, GitHub CLI, and isolated Pi autoreview. Changes remain within layout-hook lifecycle and throttled subscriber-count delivery; no run-loop architecture or public API changes are included. Human review is required.

### DIFF
--- a/PostHog/ApplicationViewLayoutPublisher.swift
+++ b/PostHog/ApplicationViewLayoutPublisher.swift
@@ -17,64 +17,94 @@
     final class ApplicationViewLayoutPublisher: ViewLayoutPublishing {
         static let shared = ApplicationViewLayoutPublisher()
 
-        private(set) lazy var onViewLayout = PostHogThrottledMulticastCallback<Void> { [weak self] subscriberCount in
-            if subscriberCount > 0 {
-                self?.swizzleLayoutSubviews()
-            } else {
-                self?.unswizzleLayoutSubviews()
+        var onViewLayout: PostHogThrottledMulticastCallback<Void> { callbacks }
+        private var callbacks: PostHogThrottledMulticastCallback<Void>!
+        private let viewClass: UIView.Type
+
+        private struct LayoutHook {
+            let original: IMP
+            let replacement: IMP
+            let token: NSObject
+        }
+
+        // Only accessed by the multicast's serialized subscriber-count callbacks.
+        private var installedLayout: LayoutHook?
+        private var layoutHooks: [UInt: LayoutHook] = [:]
+        private let activeLayoutLock = NSLock()
+        private var activeLayoutToken: NSObject?
+        private typealias LayoutImplementation = @convention(c) (UIView, Selector, CALayer) -> Void
+
+        init(viewClass: UIView.Type = UIView.self) {
+            self.viewClass = viewClass
+            // Initialize before publication; Swift lazy properties are not safe on concurrent first access.
+            callbacks = PostHogThrottledMulticastCallback<Void> { [weak self] subscriberCount in
+                if subscriberCount > 0 {
+                    self?.swizzleLayoutSubviews()
+                } else {
+                    self?.unswizzleLayoutSubviews()
+                }
             }
         }
 
-        private var hasSwizzled: Bool = false
-
         private func swizzleLayoutSubviews() {
-            guard !hasSwizzled else { return }
-            hasSwizzled = true
+            guard installedLayout == nil,
+                  let method = class_getInstanceMethod(viewClass, #selector(UIView.layoutSublayers(of:)))
+            else { return }
 
-            swizzle(
-                forClass: UIView.self,
-                original: #selector(UIView.layoutSublayers(of:)),
-                new: #selector(UIView.ph_swizzled_layoutSublayers(of:))
-            )
+            let hook = layoutHook(for: method_getImplementation(method))
+            activeLayoutLock.withLock { activeLayoutToken = hook.token }
+            method_setImplementation(method, hook.replacement)
+            installedLayout = hook
         }
 
         private func unswizzleLayoutSubviews() {
-            guard hasSwizzled else { return }
-            hasSwizzled = false
+            guard let installedLayout,
+                  let method = class_getInstanceMethod(viewClass, #selector(UIView.layoutSublayers(of:)))
+            else { return }
 
-            // swizzling twice will exchange implementations back to original
-            swizzle(
-                forClass: UIView.self,
-                original: #selector(UIView.layoutSublayers(of:)),
-                new: #selector(UIView.ph_swizzled_layoutSublayers(of:))
-            )
+            // A newer swizzler may still forward through our IMP. Do not overwrite its hook or wrap it again.
+            guard method_getImplementation(method) == installedLayout.replacement else { return }
+            method_setImplementation(method, installedLayout.original)
+            activeLayoutLock.withLock { activeLayoutToken = nil }
+            self.installedLayout = nil
         }
 
-        // Called from swizzled `UIView.layoutSubviews`
-        fileprivate func layoutSubviews() {
+        private func layoutHook(for original: IMP) -> LayoutHook {
+            let key = unsafeBitCast(original, to: UInt.self)
+            if let hook = layoutHooks[key] {
+                return hook
+            }
+
+            let forward = unsafeBitCast(original, to: LayoutImplementation.self)
+            let selector = #selector(UIView.layoutSublayers(of:))
+            let token = NSObject()
+            let block: @convention(block) (UIView, CALayer) -> Void = { [weak self] view, layer in
+                forward(view, selector, layer)
+                if Thread.isMainThread {
+                    self?.layoutSubviews(token: token)
+                } else {
+                    DispatchQueue.main.async { [weak self] in
+                        self?.layoutSubviews(token: token)
+                    }
+                }
+            }
+            let hook = LayoutHook(original: original, replacement: imp_implementationWithBlock(block), token: token)
+            // In-flight dispatches and other swizzlers can retain the IMP after unsubscribe.
+            // Keep it valid and reuse it rather than allocating a new block on every restart.
+            layoutHooks[key] = hook
+            return hook
+        }
+
+        private func layoutSubviews(token: NSObject) {
+            // An older hook may still be in another swizzler's call chain.
+            guard activeLayoutLock.withLock({ activeLayoutToken === token }) else { return }
             onViewLayout.invoke(())
         }
 
         #if TESTING
             func simulateLayoutSubviews() {
-                layoutSubviews()
+                onViewLayout.invoke(())
             }
         #endif
-    }
-
-    extension UIView {
-        @objc func ph_swizzled_layoutSublayers(of layer: CALayer) {
-            ph_swizzled_layoutSublayers(of: layer) // call original, not altering execution logic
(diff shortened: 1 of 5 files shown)


marandaneto:
@ioannisj or @turnipdabeets can you help me testing this out? its very specific and we dont want to break anything, burnt quite some time figuring this out already

marandaneto:
Live session replay smoke test passed on an iPhone 17 Pro simulator (iOS 26.5), using commit `f07a39f32` against US Cloud.

- Exercised stop/resume, 10 rapid stop/resume cycles, new sessions, background/foreground transitions, navigation, sheets, and synthetic text input.
- Tested with Surveys enabled and disabled.
- Observed 49 successful snapshot upload requests and inspected actual encoded replay frames to confirm masking.
- No crashes or hangs observed. One upload timeout around background/foreground recovered; subsequent uploads succeeded.

Replay: https://us.posthog.com/shared/ivb2F5GoUwu1-l4WxOuE4a4E1jVsLg?t=11

This validates the tested simulator flows, not every replay scenario or the production cause of #806. Web-player playback was not independently verified during the smoke test. Temporary example configuration/instrumentation was removed, the baseline app restored, and the worktree left clean. No project token was committed.

ioannisj:
Did a pre-pr and PR repro as well  and could not find anything that this changes could break. Extra locking was marked as insignificant and nothing above simulator noise. I'll run through the code as well but lgtm

PostHog/ApplicationViewLayoutPublisher.swift
```diff
@@ -42,36 +62,68 @@
...
         private func swizzleLayoutSubviews() {
-            guard !hasSwizzled else { return }
-            hasSwizzled = true
-
-            swizzle(
-                forClass: UIView.self,
-                original: #selector(UIView.layoutSublayers(of:)),
-                new: #selector(UIView.ph_swizzled_layoutSublayers(of:))
-            )
+            guard installedLayout == nil,
```
ioannisj: `unswizzleLayoutSubviews()` early-returns at the `method_getImplementation(method) == installedLayout.replacement` guard without clearing `installedLayout`, and `swizzleLayoutSubviews()` bails on `installedLayout == nil`.

ioannisj (approved):

PostHog/ApplicationViewLayoutPublisher.swift
marandaneto: This is intentional for the newer-swizzler case: a different current IMP does not necessarily mean our hook was removed. The chain can still be `newer hook → PostHog hook → UIKit`.

We retain `installedLayout` and its active token so a later subscription can reuse that hook without wrapping the newer hook again. If the newer swizzler subsequently restores our IMP, a later unsubscribe can restore the original implementation normally.

I reran `ApplicationViewLayoutPublisherTest/preservesNewerSwizzler()` on `a4be3e52c` with Thread Sanitizer through the make-based iOS test target. It passed: teardown preserves the newer IMP, resubscribe leaves that IMP unchanged, forwarding calls both the newer hook and original exactly once, and cleanup restores the original after the newer hook is removed.

If you mean a replacement/uninstall sequence that removes or bypasses our hook entirely, this guard does not detect that case. Could you share the sequence you have in mind? I would reproduce that separately before changing the ownership logic. Leaving this thread open for clarification.

ioannisj: The sequence I mean is a swizzler that was installed before us and later uninstalls, restoring exactly what it replaced, which is UIKit's IMP, not ours:

- launch: another library X installs `LX` over `UK` and remembers `UK` (UK is the original UIKit IMP) 
- replay starts: we install `PH` over `LX` (this is fine, we are in the chain)
- `LX` stops and restores `UK` (the IMP it replaced), so `PH` is now out of the chain
- replay stops: `unswizzleLayoutSubviews()` sees `UK` != `PH`, returns early, but gate `installedLayout` stays set
- replay starts again: `swizzleLayoutSubviews()` sees `installedLayout != nil`, returns, nothing gets installed

Library X did nothing wrong here, it restored what it replaced. From that point no layout notification fires again and replay records nothing, with no log. This was the same pre-PR, but a stop -> start would recover

I think it's an edge case though, and if we're planning to revisit this whole flow with https://github.com/PostHog/posthog-ios/issues/813, we could live with this. I just wonder if a simpler gate in `swizzleLayoutSubviews()` like "IMP != PH means we swizzle" would solve this

Already approved, so up to you to merge

marandaneto: Thanks for clarifying. I reproduced that exact older-library uninstall sequence: UIKit still ran once, but no layout notification arrived after replay restarted.

Fixed directly in this PR in 7e043c190. The start gate now checks the actual current IMP, as you suggested, rather than trusting retained installation state. Teardown retires ownership without overwriting another library’s implementation. A restored cached PostHog IMP is reactivated instead of wrapped again.

The regression covers both the older-library restore and a newer hook bypassing us, with and without another retained subscriber (such as Surveys). The retained-forwarding control also checks 100 restarts for exactly-once forwarding/notifications and stable wrapper reuse. Recovery happens on subscription changes, not continuously during recording.

I also merged latest main in 6f31b314c. Full SPM/iOS tests, three focused TSan repetitions, lint, and final autoreview passed on that head. CI is still running; the known local external-example package-identity build limitation is unchanged.

PR #815, opened 2026-09-10
fix: warn about off-main view layout in debug mode

marandaneto:

Related to #806. The layout hook forwards to UIKit before notifying PostHog, so an exception in UIKit can occur before the existing notification thread check. The investigation reproduced the crash path with and without the hook. The customer's production trigger remains unproven.

This adds a debug-gated warning before forwarding an off-main layout call. A dedicated lock makes the warning once per process. Calls while debug logging is disabled do not consume it, and restarting the layout subscription does not reset it.

UIKit forwarding remains synchronous on the calling thread. PostHog notifications still run on main. There is no layout suppression, new public option, or scheduler redesign. The lifecycle reproductions in #814 and the architecture proposal in #813 remain separate work.

- Added regression tests for main/background forwarding, view and layer identity, main-thread callback delivery, debug gating, warning order, concurrent once-only logging, and subscription restarts. The tests replace only a test view's call-through, not UIKit layout for other views.
- Verified the warning test failed against the original SDK because no warning was emitted.
- The focused iOS suite passes under Thread Sanitizer with three repetitions on iOS 26.5. The new subscriptions use replay's trailing delivery. The polling helper now runs on main to avoid reading callback counters concurrently.
- `make test`, `make format`, and `make lint` pass after syncing with main.
- `make build` passes all SDK platforms and platform examples, then fails in the external-SDK client example because its package reference assumes the checkout directory is named `posthog-ios`. The worktree is named `posthog-ios-issue-806`. CocoaPods examples were not reached.
- Isolated autoreview against `origin/main` passed for `88a4f5d0cfd80df9bf3d1c2bf9333aea8338d58f` with no actionable findings.

## :pencil: Checklist

- [x] I updated the docs if needed. No public API change.

## If releasing new changes

- [ ] Ran `pnpm changeset` to generate a changeset file. A patch changeset was added directly.

## 🤖 Agent context

Manoel directed the scope. Pi used file read/edit tools, shell commands, GitHub CLI, make-based validation, and isolated Pi autoreview. Investigation details are in #806.

The chosen change is diagnostic only. We explicitly rejected changing UIKit layout timing, adding a suppression flag, or claiming that the SDK's presence in the crash stack proves causation. A separate draft contains the lifecycle reproductions. Human review is required.

### DIFF
--- a/.changeset/quiet-layout-warning.md
+++ b/.changeset/quiet-layout-warning.md
@@ -1,0 +1,5 @@
+---
+"posthog-ios": patch
+---
+
+Log a one-time debug warning when the layout observation hook receives an off-main UIKit layout call, without changing the original layout behavior.
--- a/PostHog/ApplicationViewLayoutPublisher.swift
+++ b/PostHog/ApplicationViewLayoutPublisher.swift
@@ -26,6 +26,20 @@
         }
 
         private var hasSwizzled: Bool = false
+        private let backgroundLayoutWarningLock = NSLock()
+        private var hasWarnedAboutBackgroundLayout = false
+
+        fileprivate func warnAboutBackgroundLayout() {
+            guard hedgeLogEnabled else { return }
+            let shouldWarn = backgroundLayoutWarningLock.withLock {
+                guard !hasWarnedAboutBackgroundLayout else { return false }
+                hasWarnedAboutBackgroundLayout = true
+                return true
+            }
+            if shouldWarn {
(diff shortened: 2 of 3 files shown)


ioannisj (approved):
Not sure if warning about misuse of main thread should be this SDK's responsibility tbh. Esp since Main Thread checker is on by default for debug builds? Harmless change though so up to you

marandaneto:
> Not sure if warning about misuse of main thread should be this SDK's responsibility tbh. Esp since Main Thread checker is on by default for debug builds? Harmless change though so up to you

only with `hedgeLogEnabled` enabled and only once, so no harm, helps diagnose #806

PR #825, opened 2026-09-11
feat: attach $recording_status and $sdk_debug_* replay properties to captured events

### LINKED ISSUE #726: Session replay: attach replay debug properties to events
Part of the replay parity tracker PostHog/posthog#30889.

**Problem**

posthog-js attaches replay debug properties (`$recording_status`, `$sdk_debug_replay_*`) to every captured analytics event when recording. The iOS SDK attaches only `$session_id` / `$window_id` / `distinct_id` for replay (`buildProperties` in `PostHogSDK.swift`); the debug keys are absent.

**Scope**

- Attach the replay debug properties to captured events while recording, matching the web SDK keys.

Reference: posthog-js `packages/browser/src/posthog-core.ts:1558`, `session-recording.ts`, `constants.ts`.

**Parity across SDKs**
- Android: PostHog/posthog-android#640
- iOS: PostHog/posthog-ios#726 (this issue)
- React Native: PostHog/posthog-js#4185

ioannisj:

Closes #726. An OOM investigation got stuck because `$exception` events don't say whether session replay was recording at the time, so there was no way to tell from the event alone if replay was even a suspect.

posthog-js already merges `sessionRecording.sdkDebugProperties` into every event in `calculateEventProperties` (`packages/browser/src/posthog-core.ts`). This ports that to iOS: every non-`$snapshot` event now carries `$recording_status` plus the `$sdk_debug_*` keys from the spec in `sdk-specs` (`add-session-replay-debug-properties`). Flutter gets them for free since its captures go through `PostHogSDK.capture`.

Adds `PostHogReplayIntegration.debugProperties()` and merges it in `buildProperties`. When the integration isn't installed (replay off, or a non-iOS platform) the status is `disabled` and only the config and session keys are emitted. The buffering fields are read in one `bufferingLock` acquisition so a concurrent `start()` / `stop()` can't tear them across keys on the same event.

A few things I hit along the way that are in here too:

- `buildProperties` now reads `config`, `remoteConfig`, `queue`, `replayQueue` and `replayIntegration` from arbitrary caller threads, so those references are guarded by `setupLock` to not race `close()`. That commit also lives on `fix/sdk-close-race`, I'll rebase this once it lands separately.
- The crash-context snapshot (`notifyContextDidChange`) used to refresh only on identify/screen/register, so a crash after the first `/config` resolved still reported the setup-time `buffering`. The replay integration now re-notifies on every recording transition, and the point-in-time counters (`$sdk_debug_current_session_duration`, `$sdk_debug_pending_queue_size`, `$sdk_debug_replay_internal_buffer_length`) are stripped from the snapshot since they'd be stale by definition on the next launch.
- `$session_id` is resolved before the debug snapshot, so the event that rotates the session describes the new session's start and duration rather than the previous one's.
- `throttleDelayMs` clamps to `Int32.max`: `throttleDelay` is a public unvalidated `TimeInterval`, and `Int(Double(Int.max))` traps too.

One deliberate divergence from js: the queue depth is `$sdk_debug_pending_queue_size`, not `$sdk_debug_retry_queue_size`. Mobile has one disk-backed queue for normal batching and retry backoff, so the js name would read ordinary buffering as a stuck retry loop. That's recorded in the spec and posthog-android uses the same key.

New and updated tests in `PostHogTests/PostHogSDKTest.swift`, `PostHogSessionManagerTest.swift`, `PostHogSessionReplayRemoteConfigBufferTest.swift` and `PostHogSessionReplayEventTriggersTest.swift` cover the default `disabled` status on every event type, the buffering to active transition, the hold reasons, trigger statuses, the session-rotation ordering, the crash-context re-snapshot, and a concurrent `stop()` / `close()` while capturing. The event shape snapshot in `PostHogTests/Resources/event-shapes-batch.json` is updated for the new keys.

Ran `swift test -Xswiftc -DTESTING` on macOS (all green), and on an iPhone 16 Pro simulator `PostHogSDKTest`, `PostHogSessionManagerTest`, `PostHogSessionReplayRemoteConfigBufferTests`, `PostHogSessionReplayEventTriggersTests` and `PostHogEventSnapshotTests` (160 tests, all green). `make testOniOSSimulator` didn't get that far for me locally, the `name=` destination matched a paired Apple TV, so I ran xcodebuild against the simulator UDID directly.

## :pencil: Checklist

## If releasing new changes

## 🤖 Agent context

**DRI:** @ioannisj


marandaneto (approved):
The code looks good; no qualifying findings. Verdict: correct.

marandaneto:
Those are the existing ones already, right? That web also captures them?

ioannisj:
> Those are the existing ones already, right? That web also captures them?

Almost all with the exception of the following. 

Renamed: 
- `$sdk_debug_pending_queue_size`  (web sends `$sdk_debug_retry_queue_size`, renamed since we don't implement retry queues)

New, mobile only (we'll likely drop in the future as discussed in the other RFC, until then I think these would be useful as well)
- `$sdk_debug_replay_capture_mode`
- `$sdk_debug_replay_throttle_delay_ms`

We'll need to update taxonomy next. 

Do you think we should drop the mobile only ones?

PostHog/PostHogSDK.swift
```diff
@@ -31,10 +31,17 @@ let maxRetryDelay = 30.0
 /// Use `PostHogSDK.shared` for the default singleton instance, or `PostHogSDK.with(_:)`
 /// to create an additional configured instance.
 @objc public class PostHogSDK: NSObject { // swiftlint:disable:this type_body_length
-    private(set) var config: PostHogConfig
+    private var _config: PostHogConfig
+    /// `config`/`remoteConfig`/`queue`/`replayQueue` are written under `setupLock` (setup/close) and
+    /// read from arbitrary caller threads (capture(), debugProperties()) — guard the references
+    /// themselves with the same (recursive) lock so a reader never races `close()`'s teardown.
+    private(set) var config: PostHogConfig {
+        get { setupLock.withLock { _config } }
```
turnipdabeets, line 39: Reading `config` now takes `setupLock`, and two callers read it while already holding another lock: `optIn`/`optOut` under `optOutLock` (2537, 2576) and `identify` under `identifyLock` (905). `setup()` takes the same pairs in the opposite order — it holds `setupLock` from 167 and then takes `optOutLock` at 243 and calls `identify()` at 294. So a `close()` + `setup()` racing an in-flight `optOut()` or `identify()` deadlocks both threads. I think we need to hoist the read above the inner lock in those three spots:

```swift
let config = self.config
optOutLock.withLock {
    config.optOut = false
    storage?.setBool(forKey: .optOut, contents: false)
}
```

PostHog/PostHogSDK.swift
```diff
@@ -558,6 +587,20 @@ let maxRetryDelay = 30.0
         return true
     }
 
+    /// `$sdk_debug_session_start` / `$sdk_debug_current_session_duration` / `$sdk_debug_pending_queue_size`.
+    private func sessionDebugProperties() -> [String: Any] {
+        var props: [String: Any] = [:]
+        if let sessionStart = sessionManager.sessionStartTimestampSnapshot {
+            let nowSeconds = now().timeIntervalSince1970
+            props["$sdk_debug_session_start"] = Int64(sessionStart * 1000)
+            props["$sdk_debug_current_session_duration"] = Int64((nowSeconds - sessionStart) * 1000)
```
turnipdabeets: [nit] The session start comes from `getSessionId(at: timestamp ?? now())`, but the duration here uses `now()`. A backdated `capture(timestamp:)` that rotates the session then reports a duration equal to the backdate instead of 0 — worth passing the same resolved timestamp in?

PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -593,7 +613,9 @@
                 let canResolve = didResolveRemoteConfig && awaitingFirstRemoteConfig
                 return (canResolve, canResolve && !wasResolved)
             }
-            guard canResolve else { return }
+            // A reload that doesn't resolve the buffer can still flip the linked flag, which
+            // `$sdk_debug_replay_linked_flag_trigger_status` reads live; keep the crash context current.
+            guard canResolve else { return notifyRecordingStatusChanged() }
```
turnipdabeets, line 639: [question] Once the first `/config` has landed, this makes every `reloadFeatureFlags`/`identify` rebuild the whole property snapshot and rewrite the crash context, even when the linked flag didn't move. Could we notify only when the snapshot actually changed?

PostHogTests/PostHogSessionReplayRemoteConfigBufferTest.swift
```diff
@@ -446,5 +449,232 @@
...
+                }
+            }
+
+            sut.stopSessionRecording()
+            sut.close()
+
+            group.wait()
+
+            #expect(statuses.allSatisfy { $0 == "active" || $0 == "buffering" || $0 == "disabled" })
+            #expect(badHoldReasons == 0)
```
turnipdabeets: [nit] Both values come from one `debugProperties()` call, which derives the status and the hold reason from the same local — so `badHoldReasons` can't be non-zero regardless of the locking. The test still proves the concurrent path doesn't crash; maybe assert something that spans two locks (status vs buffer length) so it can catch a torn read too.

PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -317,13 +317,21 @@
...
+        private func notifyRecordingStatusChanged() {
+            postHog?.notifyContextDidChange()
         }
 
         /// Stops session replay recording.
         /// Note: This does not clear remoteConfigLoadedToken or eventCapturedToken as those are managed by install/uninstall.
         func stop() {
             guard isEnabled else { return }
-            isEnabled = false
+            bufferingLock.withLock { isEnabled = false }
```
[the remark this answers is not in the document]
ioannisj: Those reads are unchanged from main, the only thing this PR adds is the lock around the `stop()` write and `isActive()`. The check-then-act in `start()`/`stop()` is pre-existing, and wrapping each read in `bufferingLock` wouldn't close it either since the guard and the start are still two steps. Keeping it out of this PR, worth a follow-up to put the whole transition under one lock.

PostHog/PostHogSDK.swift
```diff
@@ -599,6 +650,21 @@ let maxRetryDelay = 30.0
...
+                    props.merge(replayIntegration.debugProperties()) { _, new in new }
+                } else {
+                    props["$recording_status"] = "disabled"
+                    props["$sdk_debug_replay_capture_mode"] = PostHogReplayIntegration.captureMode(config: config)
+                    props["$sdk_debug_replay_throttle_delay_ms"] = PostHogReplayIntegration.throttleDelayMs(config: config)
+                }
+            #else
+                props["$recording_status"] = "disabled"
+            #endif
+            props.merge(sessionDebugProperties()) { _, new in new }
```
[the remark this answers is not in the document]
ioannisj: The only caller that supplies its own `$session_id` is `$snapshot`, and that path sets `appendSharedProps: false` so it never reaches `sessionDebugProperties()`. There's a test in `PostHogSDKTest` asserting no `$sdk_debug_*` keys on a snapshot. Nothing else in the SDK passes a session id, so leaving this as is.

PostHog/Replay/PostHogReplayIntegration.swift
```diff
@@ -543,6 +559,7 @@
             } else {
                 replayQueue?.clearBuffer()
             }
+            notifyRecordingStatusChanged()
```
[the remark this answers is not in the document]
ioannisj: Fixed in e977c875b, `applyRemoteConfig` now re-snapshots on every apply, and `notifyRecordingStatusChanged` skips it when nothing reported actually moved.

PostHog/PostHogSDK.swift
ioannisj: Fixed in 7fdc3aabd, hoisted the read in all three spots. One residual: `PostHogConfig` is a class, so if a `setup()` swaps the instance between the hoisted read and the locked write, the write lands on the old object. Same window as before this PR, just moved, so I think it's fine.

PostHog/PostHogSDK.swift
ioannisj: Fixed in 8de50f182, the duration is measured at the event's resolved timestamp now and clamped at zero. The clamp is for the case that doesn't rotate: a backdated event earlier than the live session's start would go negative otherwise. js measures against the latest snapshot timestamp too, so this is closer to the reference than `now()` was. Added a test for both cases in `PostHogSessionManagerTest`.

PostHog/Replay/PostHogReplayIntegration.swift
ioannisj: Yeah, worth it, but the cost wasn't the real problem. e977c875b gates `notifyRecordingStatusChanged` on the reported values (status, hold reason, trigger statuses, capture mode, throttle), so a reload that moves nothing skips the rebuild.

The bigger thing it fixes: the rebuild resolved the session non read-only, so a background `/flags` reload after 30 idle minutes rotated the session. The snapshot resolves it read-only now.

PostHogTests/PostHogSessionReplayRemoteConfigBufferTest.swift
ioannisj: Yeah, that counter couldn't fail by construction. e977c875b samples `$recording_status` and `$sdk_debug_replay_internal_buffer_length` from one call instead, those come from two different locks. Kept it to shape checks (allowed status, non-negative length) since a `stop()` landing between the two reads legitimately changes either, so asserting a relation between them would flake.

PR #826, opened 2026-09-11
fix(replay): preserve masks across siblings and collapse animations

marandaneto:

Follow-up to two review findings on the merged #822: [inherited masking across siblings](https://github.com/PostHog/posthog-ios/pull/822#discussion_r3989418863) and [presentation bounds during collapse animations](https://github.com/PostHog/posthog-ios/pull/822#discussion_r3989418871).

Inside a full-window `ph-no-capture` view, visiting one child could clear the shared masking flag for later siblings. This could leave custom-drawn content unmasked. The walk now passes the inherited flag by value so each sibling receives the same parent state.

A clipping view could also be skipped when its model frame reached zero, even though its presentation bounds were still visible during an animation. The existing zero-frame/clipping check now consults presentation bounds before skipping the subtree. Other views do not pay for an extra presentation-layer read.

The synchronous Flutter bridge API is unchanged. Includes a patch changeset.

Both concerns were reproduced with failing regression tests on `main` before changing production code. The sibling test failed with both zero-size and normal-size preceding children. The collapse test failed when the model frame was zero but presentation bounds were nonempty. These tests now pass, along with controls for empty presentation bounds, explicit `ph-no-mask`, and masking scope.

- All 101 tests across nine relevant iOS Simulator suites passed on iPhone 17 Pro, iOS 26.5. They passed again after runtime validation and removal of the temporary harness.
- A separate real UIKit simulator run compared the base SDK with this PR, including a real collapse animation. All 52 uploaded frames across both runs were decoded and checked. Both leaks reproduced on the baseline and were masked with the fix. Explicit unmasking, subtree scope, fade-out, zero-size containers, injected render failure, and recovery controls passed. [Runtime results and limitations](https://github.com/PostHog/posthog-ios/pull/826#issuecomment-5638744953).
- `make test` passed with 172 XCTest tests and 799 Swift Testing tests.
- `make format`, `make lint`, and `make apiCheck` passed.
- `make build` passed all SDK platforms and platform examples. XCFramework creation passed, but its client failed dependency resolution because it expects package identity `posthog-ios` rather than the worktree directory name. CocoaPods targets were not reached.
- Autoreview of `e76b246d1` against `origin/main` reported no actionable findings.

## :pencil: Checklist

- [x] I updated the docs if needed. No public API documentation changes are needed.

## If releasing new changes

- [x] Added a patch changeset directly in `.changeset/`.

## 🤖 Agent context

Pi used `read`, `edit`, `write`, and `bash` tools to implement and validate this follow-up in a fresh worktree from `main`. XcodeBuildMCP and AXe were used for the simulator runtime checks. An isolated Pi autoreview checked the committed diff. No public session link is available.

The fixes keep mask state local to each subtree and retain the existing fast path for visible model geometry. They do not change the deferred Flutter capture-result behavior. Human review is required.

### DIFF
--- a/.changeset/replay-inherited-mask-and-rendered-bounds.md
+++ b/.changeset/replay-inherited-mask-and-rendered-bounds.md
@@ -1,0 +1,6 @@
+---
+"posthog-ios": patch
+---
+
+- Preserve inherited session replay masking across siblings inside full-window `ph-no-capture` views.
+- Keep collecting masks inside clipping views whose model frame reaches zero while their presentation bounds remain visible during an animation.
--- a/PostHog/Replay/PostHogReplayIntegration.swift
+++ b/PostHog/Replay/PostHogReplayIntegration.swift
@@ -858,7 +858,7 @@
             return (hasText, hasGraphic)
         }
 
-        private func findMaskableWidgets(_ view: UIView, _ window: UIWindow, _ maskableWidgets: inout [MaskedRegion], _ maskChildren: inout Bool) {
+        private func findMaskableWidgets(_ view: UIView, _ window: UIWindow, _ maskableWidgets: inout [MaskedRegion], _ maskChildren: Bool) {
             // Checked first so an explicit unmask wins over the sensitive-type early-returns
             // below, matching the modifier's precedence.
             if view.isNoMask() {
@@ -1018,6 +1018,7 @@
 
             // on RN, lots get converted to RCTRootContentView, RCTRootView, RCTView and sometimes its just the whole screen, we dont want to mask
             // in such cases
+            var maskDescendants = maskChildren
             if view.isNoCapture() || maskChildren {
                 let viewRect = view.toAbsoluteRect(window)
                 let windowRect = window.frame
@@ -1026,7 +1027,7 @@
                 if !viewRect.equalTo(windowRect) {
                     maskableWidgets.append(.init(view, in: window))
                 } else {
-                    maskChildren = true
+                    maskDescendants = true
                 }
             }
 
@@ -1036,10 +1037,9 @@
                         continue
                     }
 
-                    findMaskableWidgets(child, window, &maskableWidgets, &maskChildren)
+                    findMaskableWidgets(child, window, &maskableWidgets, maskDescendants)
                 }
             }
-            maskChildren = false
         }
 
         /// Recursively iterate through layer hierarchy to find maskable layers (iOS 26+)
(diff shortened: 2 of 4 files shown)


marandaneto:
## Real iOS Simulator validation

Tested the actual foreground app and replay uploads, in addition to the unit tests. Compared SDK source at the base `22d48898e` (3.73.2) with this PR at `e76b246d1` using the same temporary UIKit harness on **iPhone 17 Pro Simulator, iOS 26.5**.

Both runs used optimized Debug builds, a fake token, and localhost-only ingestion. The 52-second sequence produced **26 uploaded screenshot frames per run**. All 52 WebP frames were decoded and pixel-checked, and representative replay screenshots were inspected visually.

| Scenario | Baseline | This PR |
|---|---|---|
| Full-window `ph-no-capture` parent with a zero-size first child and custom-drawn later sibling | Synthetic secret was readable in the uploaded image. | Secret was masked. |
| Same layout with a normal-size first child | Synthetic secret was readable. | Secret was masked. |
| **Real 10-second collapse animation** setting the clipping parent's `frame = .zero` | Text remained readable while model geometry was zero and presentation bounds were still visible. | All **10 uploaded animation frames** kept the secret masked. |
| Explicit `ph-no-mask` child, protected sibling, and unmarked content outside the parent | Explicitly unmasked and outside content stayed visible. Protected sibling was masked. | Same behavior. |
| Completed collapse and fully transparent controls | Content was absent. | Content was absent. |
| Zero-size non-clipping parent and real 8-second fade-out | Visible sensitive content was masked. | Same behavior. |
| Injected failure in the scale-1 masking renderer | Four attempts were logged as dropped. No screenshot uploaded during the failure phase. | Same behavior. |
| Renderer recovery | Correctly masked capture resumed. | Same behavior. |

The collapse run used a real `UIView.animate`, **not the unit-test presentation-layer double**. Recorded geometry confirmed a zero model frame alongside nonempty presentation bounds. A simulator screenshot showed the text visibly on screen while the corresponding fixed replay image was blacked out.

All temporary app and renderer-hook edits were removed. The worktree is clean at `e76b246d1`, and **all 101 tests across the nine relevant iOS suites passed again** afterward. No additional SDK change was needed. Screenshots, video, original payloads, geometry samples, and pixel-check results are retained locally.

Scope: this validates the tested UIKit layouts on one simulator runtime. It does not prove every animation boundary or replace physical-device/RN/Flutter testing. The renderer failure was injected, not caused by actual memory pressure. No real PostHog project or hosted replay viewer was used.

turnipdabeets (approved):

PR #832, opened 2026-09-17
feat: SwiftUI tap autocapture

marandaneto:

SwiftUI controls can share a hosting view, which makes existing tap capture miss useful element labels or assign a neighboring control's label. This draft adds best-effort SwiftUI tap capture through the SDK's existing application-event publisher.

- Add `captureSwiftUIElementInteractions`, default `false`, independently of `captureElementInteractions` (API decision: #841).
- With the new option off, preserve legacy gesture capture, element chains, and label propagation. With both options on, retain UIKit capture and use the new SwiftUI route without duplicates.
- SwiftUI-only capture works with `captureElementInteractions = false`. Opting in changes SwiftUI event counts and element chains; existing actions may need updating.
- Resolve explicit `.postHogLabel` markers by their bounds without adding a competing gesture recognizer.
- Preserve UIKit capture and avoid duplicate events. Filter drags and long presses, support pointer input, and respect masking and consent gates.
- Add regression tests; temporary simulator-validation scaffolding has been removed.

Unresolved flattened SwiftUI surfaces are skipped. Accessibility-only identification still depends on OS behavior.

- `make test` passed, including 844 Swift Testing tests.
- The final focused iOS Simulator run passed 28 SwiftUI test functions (including parameterized cases) and 13 existing autocapture XCTest cases. Coverage includes all four configuration combinations, legacy recognized single/multi-taps, legacy label/chain preservation, pre-setup markers, no duplicate capture, and opt-out/opt-in/close. Legacy gesture and label regressions failed before gating and passed after.
- `make apiUpdate` and `make apiCheck` passed using external Xcode compatibility overrides.
- AXe drove a temporary example harness on iOS 26.5 before that harness was removed. Verified labeled button/nested-button/gesture capture, no UIKit duplicates, sheet/navigation interactions, masking, long presses, scrolling, and opt-out/opt-in/close behavior. Synthetic uploads received successful US Cloud batch responses.
- `make format`, `make lint`, passed. An earlier isolated autoreview also passed.
- Xcode 27 built the SDK platform matrix and platform examples using temporary deployment-target, architecture, and signing overrides. No repository deployment targets changed. The aggregate `make build` stopped at an external-example package-identity mismatch caused by the worktree directory name. CocoaPods targets were not reached. These builds do not validate the original minimum OS versions.

## :pencil: Checklist

## If releasing new changes

- [x] Added a minor changeset.
- [ ] Ready for release. This PR remains a draft pending broader platform and privacy review.

## 🤖 Agent context

Implemented with Pi worker agents in an isolated worktree. Validation used `make`, AXe, XcodeBuildMCP, and an isolated Pi autoreview. The shared touch publisher was chosen to avoid competing gesture recognizers. Runtime-only credentials are not included in the change or this description. The session is not linked because it contains project credentials.

Human review is required before merging. The earlier autoreview covered commit `397eec80f` against `origin/main` and reported no actionable findings.


PostHog/Autocapture/SwiftUI/View+PostHogLabel.swift
```diff
@@ -93,22 +98,17 @@
...
             //
-            if let view = findCousinView(of: PostHogSwiftUITaggable.self) {
+            if let view = findCousinView(of: PostHogSwiftUITaggable.self),
+               let window,
+               !bounds.isEmpty,
+               window.convert(bounds, from: self).insetBy(dx: -1, dy: -1).contains(window.convert(view.bounds, from: view)),
+               window.convert(view.bounds, from: view).insetBy(dx: -1, dy: -1).contains(window.convert(bounds, from: self))
+            {
                 taggedView = view
                 view.postHogLabel = label
```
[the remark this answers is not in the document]
marandaneto: Fixed in d442c6eb0. Reproduced with labelMarkerFindsMatchingCousinAndClearsSupersededTarget: the old label remained and the second cousin was not labeled. The search now applies geometry while traversing cousins and clears the superseded target, including when no match remains. The regression failed before and passes after the fix. Focused simulator validation passed 16 Swift Testing tests and 13 existing autocapture XCTest cases.

PostHog/Autocapture/SwiftUITapAutocapture.swift
```diff
@@ -0,0 +1,259 @@
...
+                classifier.move(to: point)
+            case .ended:
+                guard touchID == ObjectIdentifier(touch), let candidate,
+                      classifier.end(at: point, timestamp: touch.timestamp)
+                else {
+                    cancel()
+                    return
+                }
+                cancel()
+                processor?.process(source: .gestureRecognizer(description: EventType.kTouch), event: candidate)
```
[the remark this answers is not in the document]
marandaneto: Fixed in d442c6eb0. Reproduced with edgeReleaseOutsideMarkerDoesNotCapture: moving four points across the marker edge still emitted an event. Completion now hit-tests and resolves the ending element chain and requires it to match the starting chain. The regression failed before and passes after the fix. Focused simulator validation passed 16 Swift Testing tests and 13 existing autocapture XCTest cases.

marandaneto:
Investigated the compliance report: the same `retry_behavior.max_retries_respected` failure (expected 4 requests, got 6) is present on main in https://github.com/PostHog/posthog-ios/actions/runs/35034414181. The current maxRetries documentation explicitly excludes event/replay/log ingestion, and this PR does not change queue retry policy or the compliance adapter. Leaving that existing contract mismatch unchanged in this prototype.

Also addressed the Xcode 15.4 compilation failure in d442c6eb0 by compile-time gating the iOS 18 accessibilityHitTest optimization; indexed traversal remains available. Older-Xcode verification is pending CI. The separate macOS test-process crash (unknown Objective-C class during property serialization) did not reproduce: make test passed locally with all 844 Swift Testing tests plus XCTest. No speculative serialization changes made.

PostHogExample/AppDelegate.swift
```diff
@@ -14,6 +14,13 @@ class AppDelegate: NSObject, UIApplicationDelegate {
         // Hosted tests manage their own SDK instances.
         if NSClassFromString("XCTestCase") != nil { return true }
 
+        #if DEBUG && os(iOS)
+            if ProcessInfo.processInfo.environment["POSTHOG_PROTOTYPE"] == "1" {
```
marandaneto: remove all changes related to temp tests with POSTHOG_PROTOTYPE later

marandaneto:
Screenshot 2026-09-17 at 11 11 49

PostHog/Autocapture/SwiftUITapAutocapture.swift
```diff
@@ -0,0 +1,294 @@
...
+                }
+                remaining -= 1
+                let frame = object.accessibilityFrame
+                if frame.contains(point) {
+                    if object.isAccessibilityElement, frame.width * frame.height < traitArea {
+                        traits = object.accessibilityTraits
+                        traitArea = frame.width * frame.height
+                    }
+                    let id = accessibilityIdentifier(of: object)
+                    if id?.localizedCaseInsensitiveContains("ph-no-capture") == true || object.accessibilityTraits.contains(.notEnabled) {
```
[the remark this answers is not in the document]
marandaneto: Fixed in 232814b67. Added accessibilityLabelExclusionPreventsStructuralFallback with exact and mixed-case embedded ph-no-capture labels. Both cases failed on the previous head because the resolver returned an event, then passed after checking accessibility labels for exclusion tokens. Label contents are not added to the event payload. Focused simulator validation passed 21 Swift Testing tests plus 13 existing autocapture XCTest cases; format and lint passed. Full macOS test runs still had varying timeout/identity failures outside this iOS-only change, and aggregate builds retain the known external-example worktree package-identity blocker. Temporary prototype scaffolding is unchanged.

PostHog/Autocapture/PostHogAutocaptureEventTracker.swift
```diff
@@ -187,6 +191,9 @@
             let gestureDescription: String?
             switch self {
             case is UITapGestureRecognizer:
+                if SwiftUITapElementResolver.isSwiftUI(view) {
```
dustinbyrne: **Important — ordinary UIKit views match the SwiftUI heuristic**

`isSwiftUI` also matches app-defined classes such as `ImageHostingView: UIView` because their names contain `Hosting`. This early return suppresses their existing UIKit gesture tap events, and the unlabeled fallback does not recover them.

Consider narrowing ownership detection and covering both an unlabeled hosting-named UIKit view and genuine SwiftUI deduplication.

dustinbyrne (commented):
Reviewed the pinned head using source inspection and existing CI. The inline comments distinguish important findings, minor feedback, and decisions. No new tests or runtime reproductions were executed.

PostHog/Autocapture/PostHogAutocaptureEventTracker.swift
marandaneto: Fixed in d8148e6e3. Reproduced with hostingNamedUIKitViewKeepsGestureCapture: an unlabeled ImageHostingView with a completed UITapGestureRecognizer emitted zero events before the change, and now emits one UIKit event. Ownership detection now checks the reflected type’s SwiftUI/SwiftUICore module prefix rather than generic name substrings. Added a real UIHostingController test that exercises both the legacy gesture path and touch observer and verifies exactly one SwiftUI event. Existing resolver fixtures now use genuine SwiftUI hosts/content instead of a hosting-named UIKit stand-in. The make-wrapped simulator run passed 23 Swift Testing tests plus 13 existing autocapture XCTest cases; make test passed all 844 Swift Testing tests plus XCTest, and format/lint passed. Platform builds completed up to the existing external-example worktree package-identity blocker. Prototype-scaffolding cleanup is unchanged.

PostHogExample/AppDelegate.swift
marandaneto: Removed in 3af6356c7: the temporary SwiftUITapPrototypeView/model, POSTHOG_PROTOTYPE launch switches, and Xcode demo references. The normal example app entry points now match main. SDK implementation and regression tests are retained. No POSTHOG_PROTOTYPE or SwiftUITapPrototype references remain in this branch. The example simulator build, 23 SwiftUI tests plus 13 existing autocapture XCTest cases, make test (844 Swift Testing tests plus XCTest), format and lint passed. Aggregate builds retain the previously reported external-example worktree package-identity blocker.

PostHog/Autocapture/PostHogAutocaptureEventTracker.swift
turnipdabeets: [question] Apps already using `captureElementInteractions` would see different events after upgrading. A tap recognizer on a SwiftUI hosting view emits one event on `main` and none here. The replacement events rename the target to `button` / `SwiftUIElement` instead of the hosting class chain, and taps with no label are dropped. Existing actions that match on `$elements_chain` would stop matching. Could this go behind its own opt-in flag while it's a prototype?

PostHog/Autocapture/SwiftUI/View+PostHogLabel.swift
```diff
@@ -93,22 +98,20 @@
...
-                //           L PostHogLabelTaggerView (UIView) <- we are here
-                //
-                taggedView = superview?.superview
-                superview?.superview?.postHogLabel = label
+            let view: PostHogSwiftUITaggable? = findCousinView(of: PostHogSwiftUITaggable.self) { view in
+                guard let window, !bounds.isEmpty else { return false }
+                let markerBounds = window.convert(bounds, from: self)
+                let viewBounds = window.convert(view.bounds, from: view)
+                return markerBounds.insetBy(dx: -1, dy: -1).contains(viewBounds)
+                    && viewBounds.insetBy(dx: -1, dy: -1).contains(markerBounds)
```
turnipdabeets: [question] The ±1pt bounds match changes behavior for padded controls. `TextField("", text: $t).padding(20).postHogLabel("x")` labels the underlying `UITextField` on `main` but leaves it `nil` here, so its change events lose the label. Without the padding it still works. Could the match use containment instead of near-equal bounds, or keep the old fallback for UIKit-backed controls?

turnipdabeets (commented):
[question] The description calls this a prototype that isn't release-ready, but it targets `main` with a minor changeset, so the next release would ship it. Is that intended?

marandaneto:
> [question] The description calls this a prototype that isn't release-ready, but it targets `main` with a minor changeset, so the next release would ship it. Is that intended?

just outdated desc

PostHog/Autocapture/SwiftUI/View+PostHogLabel.swift
marandaneto: Fixed in a56754476. Added a real UIHostingController/TextField regression with 0pt and 20pt padding: the 20pt case failed before the fix while the unpadded case passed. Matching now allows nonempty UIKit control bounds contained within the marker bounds, rather than requiring near-equal bounds. Both cases pass after the fix, along with neighboring-control and stale-label regressions (24 SwiftUI tests plus 13 existing autocapture XCTest cases). make test, format, and lint passed. Aggregate make build reaches the existing external-example SwiftPM worktree package-name blocker.

PostHog/Autocapture/PostHogAutocaptureEventTracker.swift
marandaneto: its not a prototype, but I should check if theres incompatibilities when upgrading the SDK, this should be additive

marandaneto: Addressed in c18426242, following the maintainer API decision recorded in #841. Added public/Objective-C captureSwiftUIElementInteractions, default false and independent of captureElementInteractions. The new observer, hosting-tap suppression, and changed label propagation are gated by this opt-in. Leaving it off preserves legacy recognized gestures and element chains; SwiftUI-only capture works when the legacy flag is false. Both enabled captures once, not twice. Docs and the minor changeset explicitly describe the opt-in event semantics.

Retained regression tests for legacy hosting gestures and ancestor labels failed before gating and pass after. Final simulator validation passed 28 SwiftUI test functions (including parameterized single/multi-tap and four-configuration cases) plus 13 existing autocapture XCTest cases. Tests include pre-setup markers and opt-out/opt-in/close. make test (844 Swift Testing tests plus XCTest), format, lint, and API snapshot checks passed. Aggregate make build still encounters the external-example worktree package-name blocker; original deployment targets remain unchanged.

turnipdabeets (approved):

dustinbyrne (approved):
The opt-in SwiftUI tap path preserves default behavior, and the earlier hosting-name concern is addressed.

AI-assisted review: source inspection and existing CI; no new tests or device reproduction executed.

PR #836, opened 2026-09-17
fix: remember app versions when lifecycle capture is disabled

marandaneto:

The same bug as https://github.com/PostHog/posthog-android/pull/797 affects iOS. When lifecycle capture is disabled, the SDK does not save the app version or build. Enabling capture on a later launch can emit a false `Application Installed` event. Updates while capture is disabled can also leave stale previous version/build values.

Always perform version bookkeeping when consent permits integration installation, and save the current version/build before deciding whether to emit events. Keep the launch comparison in memory for the first capture-enabled client. Disabled clients do not subscribe to lifecycle notifications or claim event-capture ownership, so they cannot block a later enabled client. Opt-out still prevents integration installation. This cannot recover history that older SDK versions never stored.

- Confirmed the new regressions fail against the original code and pass with the fix on an iOS 26.5 simulator.
- Reproduced the review's disabled-first-client ownership regression with `make test filter=PostHogAppLifeCycleIntegrationTest`: the enabled client emitted no lifecycle events before the fix. The retained regression now passes for installs and upgrades, verifies previous version/build values, and checks that closing the disabled client does not release the enabled client's ownership.
- Added a release/recreate regression with explicit close as the control. `make test filter=PostHogIntegrationInstallationTest` failed only for deinitialization before the fix and passes in both cases now. Ownership is tracked independently of the weak SDK reference so deinitialization can release it safely.
- `make test`: all 847 tests passed on the final revision.
- Focused iOS lifecycle and integration-installation suites, run through a temporary make target: all 33 tests passed.
- `make format` and `make lint` passed. Lint reports existing warnings.
- `make build` with Xcode 26.6 passed SDK builds for iOS, macOS, Mac Catalyst, tvOS, watchOS and visionOS, plus the platform examples. It stopped at the external SDK client example because SwiftPM derives the local package identity from the worktree folder name (`posthog-ios-lifecycle-version-accounting` rather than `posthog-ios`). CocoaPods examples were not reached.
- Automated review of the committed branch found no actionable issues.

## :pencil: Checklist

## If releasing new changes

- [x] Added a patch changeset file manually, equivalent to `pnpm changeset`.

## 🤖 Agent context

Implemented with Pi using repository tools in a dedicated worktree. Used make-based tests and builds, GitHub CLI, and an isolated Pi autoreview. The fix keeps version storage inside the existing lifecycle integration so opt-out continues to prevent integration installation. Human review is required.


PostHog/PostHogConfig.swift
```diff
@@ -564,9 +564,7 @@ public typealias BeforeSendBlock = (PostHogEvent) -> PostHogEvent?
             integrations.append(PostHogScreenViewIntegration())
         }
 
-        if captureApplicationLifecycleEvents {
-            integrations.append(PostHogAppLifeCycleIntegration())
-        }
+        integrations.append(PostHogAppLifeCycleIntegration())
```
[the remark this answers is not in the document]
marandaneto: Fixed in b23a80527. Reproduced with `make test filter=PostHogAppLifeCycleIntegrationTest`: when a disabled client initialized first, the enabled client emitted only the marker event and no lifecycle events. The retained test now passes for both fresh installs and upgrades, including previous version/build properties. It also verifies that closing the disabled client cannot release the enabled client’s ownership.

Version bookkeeping now runs once per process without claiming lifecycle-capture ownership. The initial launch comparison stays available for the first enabled client, which alone subscribes to lifecycle notifications. Validation: 846 macOS tests passed on rerun, 32 focused iOS simulator tests passed, lint and all-platform SDK builds passed. Final branch autoreview found no actionable issues.

PostHog/AppLifeCycle/PostHogAppLifeCycleIntegration.swift
```diff
@@ -43,6 +52,8 @@ final class PostHogAppLifeCycleIntegration: PostHogIntegration {
     }
 
     func uninstall(_ postHog: PostHogSDK) {
+        guard self.postHog === postHog else { return }
```
dustinbyrne: **Important — lifecycle ownership survives SDK deinitialization**

When `PostHogSDK.deinit` calls uninstall, this integration's weak `postHog` reference reads as nil. The identity guard then skips the existing nil-tolerant cleanup, leaving process-wide lifecycle ownership claimed.

For an additional SDK instance that actually deallocates—for example, with swizzling, screen capture, and error autocapture disabled—releasing it without `close()` can prevent its replacement from capturing lifecycle events while manual capture still works. Explicit close is unaffected.

Consider tracking ownership independently of the weak reference, preserving the disabled-client protection, and adding a release/recreate regression with explicit close as the control.

dustinbyrne (commented):
Reviewed the pinned head using source inspection and existing CI. The inline comments distinguish important findings, minor feedback, and decisions. No new tests or runtime reproductions were executed.

PostHog/AppLifeCycle/PostHogAppLifeCycleIntegration.swift
marandaneto: Fixed in 363eccddf. Added a release/recreate regression with explicit close as the control. Before the fix, `make test filter=PostHogIntegrationInstallationTest` confirmed the first SDK actually deallocated, then failed because its replacement could not install lifecycle capture. The explicit-close control passed. Both cases now pass.

Ownership is tracked independently of the weak SDK reference. Only an integration that acquired ownership can release it, including during SDK deinitialization; disabled clients still cannot clear another client’s ownership. Validation: 847 macOS tests, 33 focused iOS simulator tests, formatting, lint, and all-platform SDK builds passed. Final branch autoreview found no actionable issues. The full example build still encounters the previously documented worktree package-name mismatch.

PostHog/AppLifeCycle/PostHogAppLifeCycleIntegration.swift
```diff
@@ -34,7 +37,15 @@ final class PostHogAppLifeCycleIntegration: PostHogIntegration {
     private var didFinishLaunchingToken: RegistrationToken?
 
     func install(_ postHog: PostHogSDK) -> PostHogIntegrationInstallResult {
-        installIfNeeded(using: Self.integrationInstallState) {
+        Self.versionLock.withLock {
+            Self.recordAppVersion()
+        }
+        guard postHog.config.captureApplicationLifecycleEvents else {
+            return .installed
```
turnipdabeets, line 44: nit/suggestion, non-blocking: returning `.installed` for disabled clients means the integration counts as installed even though it does nothing. That's why the `ownsLifecycleCapture` guard in `uninstall` is needed: without it, a disabled client's uninstall would release the enabled client's ownership. Did you consider keeping the `captureApplicationLifecycleEvents` check in `getIntegrations()` and calling `recordAppVersion()` from `installIntegrations()` instead? That already runs only when not opted out, and the ownership flag would go away. Fine to keep as is.

.changeset/quiet-foxes-remember.md
```diff
@@ -0,0 +1,5 @@
+---
+"posthog-ios": patch
+---
+
+Remember the app version and build when lifecycle capture is disabled, so enabling it on a later launch does not report a false Application Installed event or stale previous version.
```
turnipdabeets: nit: this also changes what happens when the build number stays the same: the version name is now saved, so the next `previous_version` will be different. Maybe add half a sentence here?

PostHogTests/PostHogAppLifeCycleIntegrationTest.swift
```diff
@@ -93,7 +93,96 @@ final class PostHogAppLifeCycleIntegrationTest {
...
+
+    @Test("a disabled client does not block lifecycle capture by an enabled client", arguments: [false, true])
+    func disabledClientDoesNotBlockLifecycleCapture(previousInstall: Bool) async throws {
+        setVersionDefaults(version: previousInstall ? "0.0.1" : nil, build: previousInstall ? "1" : nil)
+        let disabled = getSut(captureApplicationLifecycleEvents: false)
+        defer { disabled.close() }
+        let enabled = getSut(flushAt: 4)
+        defer { enabled.close() }
+
+        disabled.close()
```
turnipdabeets, line 110: nit: `disabled.close()` is also in the `defer` on line 107. It's harmless, but it looks like a mistake. Drop one?

turnipdabeets (approved):
LGTM. I checked opt-out, opt-in, releasing the SDK without `close()` and multiple SDK instances, and they all work. A few non-blocking comments inline.

Not a blocker, and not something this PR changed: opted-out launches don't install integrations, so the version isn't saved. If someone opts out, updates the app, and opts in on a later launch, they still get a stale `previous_version`. Android (#797) keeps saving the version while opted out. We could do the same here in a follow-up.

.changeset/quiet-foxes-remember.md
marandaneto: Addressed in 6e205a3af: the changeset now explicitly mentions saving version-name changes when the build number stays the same, and the effect on the next previous_version.

PostHogTests/PostHogAppLifeCycleIntegrationTest.swift
marandaneto: Addressed in 6e205a3af: removed the deferred duplicate close, retaining the explicit close before lifecycle notifications because it verifies that a disabled client cannot release the enabled client’s ownership. The focused iOS lifecycle/integration suites passed (33 tests).

PostHog/AppLifeCycle/PostHogAppLifeCycleIntegration.swift
marandaneto: Considered this; keeping the current design for this PR as suggested. Moving bookkeeping into installIntegrations would avoid installing the disabled integration, but also couples SDK installation to lifecycle bookkeeping. The ownership flag currently makes the acquired/not-acquired distinction explicit and survives the weak SDK reference becoming nil during deinit. Both release/recreate and disabled-client-close regressions pass on iOS. No change to opted-out bookkeeping in this PR.

marandaneto:
Follow-up in 6e205a3af addresses the changeset wording and duplicate close.

Validation: `make format` and `make lint` passed (existing lint warnings); `make test` passed all 847 tests on rerun after an unrelated identity deduplication test failed on the first run. A temporary make target ran the iOS lifecycle/integration and queue suites: 33 Swift Testing tests and 18 Quick tests passed. With Xcode 26.6, `make build` passed all SDK platforms and platform examples before the previously documented external-client local-package identity mismatch; CocoaPods examples were not reached.

CI investigation: existing GitHub checks are green, but the advisory compliance report has a real contract mismatch: `retry_behavior.max_retries_respected` expects 4 requests and observes 6. Ingestion intentionally retains retryable records beyond maxRetries (introduced in #788); PostHogConfig documents that maxRetries now limits push-subscription registration only. The existing iOS test `retains_retryable_HTTP_failures_past_maxRetries_and_drains_after_recovery` passed, as did its transport-error counterpart. I did not change this intended behavior to satisfy the stale harness expectation. The harness contract needs a separate correction for durable mobile ingestion; that report failure remains outstanding.

PR #838, opened 2026-09-17
fix(surveys): omit unconfigured iOS text placeholder

### LINKED ISSUE #837: fix(surveys): leave open-text placeholder empty when unset
## Problem

A customer left **Placeholder text** empty for an open-text survey question, but the native iOS survey still shows “Start typing...” in English. This was reported in support ticket #70469 and confirmed for iOS in the [customer Slack thread](https://posthog.slack.com/archives/C09J17L1K09/p1788271056990419). Web and React Native were fixed in [posthog-js#4725](https://github.com/PostHog/posthog-js/pull/4725).

## Cause

`OpenTextQuestionView` in `PostHog/Surveys/QuestionTypes.swift` renders `appearance.placeholder ?? "Start typing..."` when the input is empty. A missing placeholder therefore becomes an English SDK default.

## Expected behavior

- A missing or empty configured placeholder leaves the text input empty.
- A configured placeholder still appears while the response is empty and disappears once the user types.
- Add focused coverage for those cases.

lucasheriques:
## Problem

An open-text survey with no configured placeholder still shows the English SDK fallback “Start typing...” on iOS. This differs from the survey editor and the web/React Native behavior fixed in PostHog/posthog-js#4725.

## Change

Render placeholder text only when the survey supplies a nonempty value. Add a patch changeset.

Fixes #837.

## Verification

- `make test filter=PostHogSurveysTest` passed.
- `make lint` passed.
- `make buildIOS` could not complete on this Mac: `xcpretty` is missing and CoreSimulator is incompatible with the installed Xcode. The Swift package tests do not compile the iOS-only view, so CI's iOS build remains the compile check.


PostHog/Surveys/QuestionTypes.swift
```diff
@@ -30,8 +30,8 @@
                     .foregroundColor(inputTextColor)
                     .overlay(
                         Group {
-                            if text.isEmpty {
-                                Text(appearance.placeholder ?? "Start typing...")
+                            if text.isEmpty, let placeholder = appearance.placeholder, !placeholder.isEmpty {
```
turnipdabeets, line 33: [nit] Android trims here (`isNotBlank`, `SurveyAppearance.kt:107`), so a whitespace-only placeholder shows nothing there but would draw a blank hint on iOS. `!placeholder.trimmingCharacters(in: .whitespaces).isEmpty` would keep the two in step.

turnipdabeets (approved):

PostHog/Surveys/QuestionTypes.swift
lucasheriques: Addressed in #839, since this PR has already merged. The existing visibility condition now checks `.whitespacesAndNewlines`, covering spaces, tabs, and line breaks; the displayed string itself remains unchanged. The follow-up includes a patch changeset and records the local validation results.

